PR #2, opened 2014-05-17
Makes Tailer use ScheduledExecutorService from Java 5

triceo:
This has multiple benefits:
- it provides periodic runs of the read operations without needing to
  invoke Thread.sleep().
- it removes any looping from the run() method, making for cleaner code.
- generally brings Tailer closer to what modern Java looks like.

No public API changes have been made. All tests are passing without
modifications.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/Tailer.java
+++ b/src/main/java/org/apache/commons/io/input/Tailer.java
@@ -16,17 +16,14 @@
  */
 package org.apache.commons.io.input;
 
-import static org.apache.commons.io.IOUtils.EOF;
 
-import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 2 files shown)


garydgregory:
@triceo 
May you rebase on master?

garydgregory:
@triceo
May you rebase on master? The implementation has changed to use the builder pattern with an internal Duration instead of long to track the sleep delay. It looks like the code needs to be adapted to configure the builder with an executor.

triceo:
@garydgregory I may eventually get to it, but considering that the original PR is from 2014, I have moved on.

garydgregory:
Either it's useful or not, regardless of your moving about 😉

PR #9, opened 2015-12-16
Added a true iterator for traversing the file trees

thammegowda:
## What is the issue with existing iterator ?
1. The existing `iterateFiles()`  is not really an iterator. It uses an in memory list which is problematic when dealing with millions of child files.
2. The existing one blocks till  it traverses the whole tree to build the list. 
## 
## What is good about the new iterator?
1. This one performs depth first traversal on file tree with the help of a stack as and when the files are consumed,. So no in memory list.
2. It doesnt block till it read all the file descriptors. You are good to consume the moment the first file is read

### DIFF
--- a/src/main/java/org/apache/commons/io/DepthFirstFileIterator.java
+++ b/src/main/java/org/apache/commons/io/DepthFirstFileIterator.java
@@ -28,9 +28,9 @@ public class DepthFirstFileIterator implements Iterator<File> {
      */
     public DepthFirstFileIterator(File root, FileFilter fileFilter){
         this.stack = new Stack<File>();
+        this.fileFilter = fileFilter;
         this.stack.add(root);
         this.next = getNext();
-        this.fileFilter = fileFilter;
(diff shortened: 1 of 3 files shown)


krosenvold:
How does this actually compare to FileTreeWalker ?

thammegowda:
Thanks for checking this and replying :+1: 

I assume you are referring to `java.nio.file.FileTreeWalker`.

As you might aware, `Iterator` and `TreeWalker` are two different design patterns.
Although both can achieve the same functionality, IMHO `Iterator` is more easier to use in Java code. Its easier because more APIs are designed to work with Iterators than Callbacks/Delegates.

For FileTreeWalker is Javascript-ish design, this takes a ` FileVisitor` delegate, the call to this delegate may then become asynchronous.

garydgregory:
`FileTreeWalker' is package private, `java.nio.file.Files.walk*` is what matters here IMO.

PR #14, opened 2016-07-16
[IO-480] Removed the deprectaed method closeQuietly from the Java doc…

rajivpjs:
As the closeQuietly method has been deprecated, I thought I would use this ticket IO-480 to remove the Java documentation.


britter:
Hello @rajivpjs, we should also modify the JavaDoc of `closeQuietly` methods to reflect that they only catch IOExceptions but will propagate any RuntimeExceptions thrown from the `close()` method.

rajivpjs:
I will have that done shortly.

Thanks.

rajivpjs:
@britter - I have made changes. Please review. 

Just out of question, where would the propagation of the RuntimeExceptions come from?

Thanks.

PR #18, opened 2016-09-24
Added CopyToDirectory convenience method

BVanderWilp:
In addressing issue "Add convenience methods for copyToDirectory", the copyToDirectory function was added. It simply calls the pre-existing functions copyDirectoryToDirectory or copyFileToDirectory based on whether or not whether or not the input file is a directory. Additionally, tests were added that replicated the tests for these two functions, but instead used copyToDirectory in order to ensure the behavior was the same.


garydgregory:
The grammar fixes "a collection" -> "an collection" are incorrect; "a collection" is the correct usage.

PascalSchumacher:
@BVanderWilp I would be nice if you would incorporate garys feedback and rebase this pull request onto current master. Thanks!

PR #19, opened 2016-09-24
[IO-483] FilenameUtils.getPrefixLength fix for unix files/folders starting with colon

marko-vasic:
FilenameUtils.getPrefixLength now works correctly for unix files/folder that are in the root folder and start with colons

### DIFF
--- a/src/main/java/org/apache/commons/io/FilenameUtils.java
+++ b/src/main/java/org/apache/commons/io/FilenameUtils.java
@@ -667,6 +667,8 @@ public class FilenameUtils {
                         return 2;
                     }
                     return 3;
+                } else if (ch0 == UNIX_SEPARATOR) {
+                    return 1;
                 }
                 return NOT_FOUND;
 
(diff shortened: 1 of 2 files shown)


kinow:
I think the proposed solutions works with no regressions in the code. Checked out the pull request locally, played with the tests adding a few other entries, but found no issues.

Will leave it open for others to review. Might be interesting to add the new test case examples to the method Javadoc too, IMO.

marko-vasic:
Thank you for the quick follow up.
Let us see what will others say.

kinow:
Pull request merged. Mentioned you in the issue entry in changes.xml @marko-vasic. Let me know if your name is not Marko Vasic, or if you'd like me to use a different in that file.

https://github.com/apache/commons-io/blob/593de775be9e5cb9c9a6cbcebe9a40ce80cc4bfd/src/changes/changes.xml#L50

And thank you for your contribution
Bruno

PR #20, opened 2016-09-24
[IO-499] FilenameUtils.directoryContains false positive issue 

cagdasyelen:
IO-499 bug has been fixed. 

The old version was looking at if the child canonical path string starts with the parent's. 
However, it fails in the case of:

.../top/foo
.../top/foo2/b.txt

since the directory path of b.txt starts with the same directory path with the one above even though the second one is a different directory(foo2). This issue is resolved by comparing the path strings of foo and foo2.

### DIFF
--- a/src/main/java/org/apache/commons/io/FilenameUtils.java
+++ b/src/main/java/org/apache/commons/io/FilenameUtils.java
@@ -529,24 +529,48 @@ public class FilenameUtils {
      * @since 2.2
      * @see FileUtils#directoryContains(File, File)
      */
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
quoting Federico Bonellis comment from the jira issue (https://issues.apache.org/jira/browse/IO-499):

>@cagdasyelen I'm afraid this patch doesn't consider the case when the file path is composed using '\' separators (aka Windows case). We should adapt it to that case before pulling the patch.

PR #21, opened 2016-09-25
fixed IO-447

MengshiZhang:


src/main/java/org/apache/commons/io/FileSystemUtils.java
```diff
@@ -292,10 +292,15 @@ long freeSpaceOS(final String path, final int os, final boolean kb, final long t
      * @param timeout The timeout amount in milliseconds or no timeout if the value
      *  is zero or less
      * @return the amount of free drive space on the drive
-     * @throws IOException if an error occurs
+     * @throws IOException if an error occurs, NullPointerException if path cannot be parsed by FilenameUtils.normalize()
      */
-    long freeSpaceWindows(String path, final long timeout) throws IOException {
+    long freeSpaceWindows(String path, final long timeout) throws IOException, NullPointerException {
```
PascalSchumacher: `NullPointerException` is a `RuntimeException` so it should be part of the method signature.

src/main/java/org/apache/commons/io/FileSystemUtils.java
```diff
@@ -292,10 +292,15 @@ long freeSpaceOS(final String path, final int os, final boolean kb, final long t
      * @param timeout The timeout amount in milliseconds or no timeout if the value
      *  is zero or less
      * @return the amount of free drive space on the drive
-     * @throws IOException if an error occurs
+     * @throws IOException if an error occurs, NullPointerException if path cannot be parsed by FilenameUtils.normalize()
```
PascalSchumacher: this should be a separate `throws` declaration

PR #22, opened 2016-09-27
[IO-515] Add constructors that allow the initial size of the in memory buffer

brettlounsbury:
Add constructors that allow the initial size of the in memory buffer to be set to a value other than the default value in ByteArrayOutputStream.  The current default is 32B which can have large performance impact if the threshold is large enough due to multiple array doubling iterations.  As an example, to buffer 1MB the array would double 15 times and would create approximately 1MB of garbage arrays in addition to the 1MB of buffer at the end.

https://issues.apache.org/jira/browse/IO-515


brettlounsbury:
Verified the Redo commit in master is feature identical.  This closes #22.

PR #24, opened 2016-10-09
[IO-516] fixed line endings for source files and ignored line endings…

pdinc-oss:
CLA on file


PascalSchumacher:
Thank! 👍 

Sorry that I "stole" the authorship of the commit from you. I did not do it intentionally, I made a mistake while cherry-picking. :(

PR #27, opened 2016-11-08
[IO-510] Adding ReadAHeadInputStream

yairzaslavsky:
Adding ReadAHeadInputStream which acts
similar to BufferedInputStream only it uses
a thread to fill in the buffer with reads
from the internal stream

### DIFF
--- a/src/main/java/org/apache/commons/io/input/ReadAheadInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/ReadAheadInputStream.java
@@ -1,0 +1,193 @@
+package org.apache.commons.io.input;
+
+import static java.lang.System.arraycopy;
+import static org.apache.commons.io.IOUtils.EOF;
+
+import java.io.FilterInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+
+/**
+ * InputStream proxy with a buffer that uses a thread to read
+ * the underlying stream up to the available space
+ * of the defined buffer.
+ * The proxied input stream is closed when the {@link #close()} method is
+ * called on this proxy. It is configurable whether the internal input stream
+ * stream will also closed.
+ *
+ * @version $Id$
+ * @see java.io.BufferedInputStream
+ */
+public class ReadAheadInputStream extends FilterInputStream {
+
+    public static final int DEFAULT_BUFFER_SIZE = 8192;
+
+    private final byte[] buffer;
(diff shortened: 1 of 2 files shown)


garydgregory:
How about allowing to pass in an ExecutorService to the ctor so that the thread can be created out of a pool.

Any thoughts about subclassing FilterInputStream?

yairzaslavsky:
@garydgregory 
Thanks for your response.

A. Regarding to ExecutorService -  sounds like a great idea to me.
I assume you suggest using the submit method, am I correct?

B. I will try to do that -
I assume I will only need to override the read methods, am I correct?

Thanks for your review

garydgregory:
Hi,

A) Yes, I was thinking about submit(Runnable)
B) You'll have to see what is need, I've not tried it.

Gary

garydgregory:
Since you mentioned BufferedInputStream, note that BufferedInputStream subclasses FilteredInputStream.

yairzaslavsky:
@garydgregory  uploaded a new PR, based on your suggestions.

yairzaslavsky:
@garydgregory 
After I synced my fork with master, and rebased, not sure why coveralls fails.
I'm less familiar with this tool, and could not see a clear indication where my code breaks the coverage.
Any assistance on this will be helpful

Cheers,
Yair

PascalSchumacher:
I guess there are some random elements in the test code of commons-io. The code coverage seems to vary a bit even if there are not actual changes.

You can check https://coveralls.io/builds/8959373 verify if the coverage changes are really caused by your pull request.

yairzaslavsky:
@PascalSchumacher  - this is not what I see when I view the list of files a the  URL you provided (and I already checked that before posting the question).
At first , I did had an issue as my fork was not synchronized with the mirror of commons-io on github,

Any other tips on the issue are more than welcome :)

Thanks,
Yair

yairzaslavsky:
@PascalSchumacher  I meant the tests for  class, or the class itself do not appear in the list of files that coveralls mentions.

PascalSchumacher:
@yairzaslavsky to quote @bodewig from: https://github.com/apache/commons-compress/pull/12

>I think this is related to our Travis CI setup using different JDKs and Coveralls picks whichever build finishes first. For different JDKs you get slightly different coverage values.

PascalSchumacher:
@garydgregory What do you think, is this now good to merge? Thanks!

garydgregory:
I can look tomorrow. Note that the spelling it "ahead" as one word.

garydgregory:
Thank you for your updated patch. May you please add Javadocs to the public methods?

garydgregory:
Closing since `ReadAheadInputStream` is in Commons IO as of 2.9.0.

PR #31, opened 2017-02-18
DirectoryUtils - isEqual to compare directories

jfiala:
First implementation of DirectoryUtils.isEqual() including optional file content compare.


PascalSchumacher:
Jira issue for discussion: https://issues.apache.org/jira/browse/IO-532

jochenw:
What is the use case? I can hardly imagine a case, where I would want to compare directories. If I need equality, then I'd rather create a copy in a temporary directory.

jochenw:
Besides, rather than simply returning a boolean value, I'd hope for an event API, like

        public interface DirCompareListener {
            /**
             * Called, if a file has been detected in the first directory, which is missing in the second.
             * In other words, the difference can be remedied by copying the file from the first
             * directory to the second.
             * @param pFirstDir First directory being compared (root directory); the directory, where
             *                               a file has been found, which is not present in the second directory.
             * @param pSecondDir Second directory being compared (root directory); the directory, 
             *                               where a file is missing.
             * @param pFile Path pf the additional file, relative to the root directory.
             */
             public void fileAdded(Path pFirstDir, Path pSecondDir, String pFileName);
        }
        /** Method, which compares two directories by iterating over the files in the first directory,
          * checking, whether they are present (and identical) in the second directory, and vice versa.
          * Upon detection of differences, a suitable method of the listener is being invoked.
          */
        public void compareDirectories(Path pFirstDir, Path pSecondDir, DirCompareListener pListener);
        /** Method, which compares two directories by iterating over the files in the first directory,
          * checking, whether they are present (and identical) in the second directory, and vice versa.
          * Upon detection of differences, the value false is being returned, otherwise returns true.
          */
        public boolean compareDirectories(Path pFirstDir, Path pSecondDir) {
            final DirCompareListener listener = new DirCompareListener(){
                 public void fileAdded(Path pFirstDir, Path pSecondDir, pFileName) {
                     throw new IllegalPathStateException();
                 }
            };
            try {
                  compare(pFirstDir, pSecondDir, listener);
                  return true;
            } catch (IllegalPathStateException e) {
                  return false;
            }
       }

jfiala:
You're right it is probably not a daily use case, but I thought it would be handy to have a routine for this.
For small automated directory comparison the boolean should be OK, of course it can be improved to an API-like routine. The intention was to setup a quick comparison routine as such was not already present...

TarCV:
This would be great to have to do integration testing of libraries generating some file trees.
Is there any plans on merging this PR?

src/main/java/org/apache/commons/io/DirectoryUtils.java
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * General directory manipulation utilities.
+ * <p>
+ * Facilities are provided in the following areas:
+ * <ul>
+ * <li>comparing directories
+ * </ul>
+ * <p>
+ *
```
garydgregory: Closing of `p` tag is not correct.

src/main/java/org/apache/commons/io/DirectoryUtils.java
```diff
@@ -0,0 +1,149 @@
...
+ * </ul>
+ * <p>
+ * 
+ * Original code: http://stackoverflow.com/questions/14522239/test-two-directory-trees-for-equality
+ *
+ */
+public class DirectoryUtils {
+
+    /**
+     * checks if the directory file lists are equal.
```
garydgregory: Start a sentence with a capital letter.

src/main/java/org/apache/commons/io/DirectoryUtils.java
```diff
@@ -0,0 +1,149 @@
...
+     * 
+     * If checkFileContent is true, then also if the file content is equal
```
garydgregory: End a sentence with a period.

src/main/java/org/apache/commons/io/DirectoryUtils.java
```diff
@@ -0,0 +1,149 @@
...
+     * @param directory
+     *            the directory
+     * @param compareDirectory
+     *            the directory to compare with
+     * @param checkFileContent
+     *            also compare file content
+     * @return true if directory and compareDirectory are equal
+     * @throws IOException
+     */
+    public static boolean isEqual(Path directory, Path compareDirectory, boolean checkFileContent)
```
garydgregory: TL;DR: Use `final` where possible; IMO https://garygregory.wordpress.com/2013/01/26/the-final-kiss-in-java/

src/test/java/org/apache/commons/io/DirectoryUtilsTestCase_Nested.java
```diff
@@ -0,0 +1,66 @@
+package org.apache.commons.io;
+
+import static org.junit.Assert.*;
```
garydgregory: Do not use `*` imports.

garydgregory: Also based your PR on the HEAD of master. This does not compile since we now use JUnit 5.

src/test/java/org/apache/commons/io/DirectoryUtilsTestCase_Simple.java
```diff
@@ -0,0 +1,66 @@
...
+
+    String basePath = "src/test/resources/DirectoryUtils_equal/simple/";
+
+    @Test
+    public void compare_self() throws IOException {
+        boolean checkFileContent = false;
+        Path directory = FileSystems.getDefault().getPath(basePath + "directory");
+        Path compareDirectory = FileSystems.getDefault().getPath(basePath + "directory");
+
+        assertTrue(DirectoryUtils.isEqual(directory, compareDirectory, checkFileContent));
```
garydgregory: All tests method are incomplete and should be something like:
```
        assertTrue(DirectoryUtils.isEqual(directory, compareDirectory, checkFileContent));
        assertTrue(DirectoryUtils.isEqual(directory, directory, checkFileContent));
        assertTrue(DirectoryUtils.isEqual(compareDirectory, directory, checkFileContent));
        assertTrue(DirectoryUtils.isEqual(compareDirectory, compareDirectory, checkFileContent));
```

src/main/java/org/apache/commons/io/DirectoryUtils.java
```diff
@@ -0,0 +1,149 @@
...
+
+        if (directory != null && compareDirectory != null) {
+            // LOGGER.info("checking directory " + directory);
+            // LOGGER.info("checking compareDirectory " + compareDirectory);
+
+            File directoryFile = directory.toFile();
+            File compareFile = compareDirectory.toFile();
+
+            // check, if there is the same number of files/subdirectories
+            File[] directoryFiles = directoryFile.listFiles();
```
garydgregory: Don't use `listFiles()`, this will kill performance on large file trees. Use Java's streaming APIs.

src/main/java/org/apache/commons/io/DirectoryUtils.java
```diff
@@ -0,0 +1,149 @@
...
+
+        // LOGGER.info("compareDirectoryContents: " + directory);
+        try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(directory)) {
+
+            for (Path directoryFilePath : directoryStream) {
+                // LOGGER.info("processing " + directoryFilePath);
+                // search for directoryFile in the compareDirectory
+                Path compareFilePath = compareDirectory.resolve(directoryFilePath.getFileName());
+
+                if (compareFilePath != null) {
```
garydgregory: When does `Path#resolve()` ever return `null`?

src/main/java/org/apache/commons/io/DirectoryUtils.java
```diff
@@ -0,0 +1,149 @@
...
+        } else {
+            return false;
+        }
+    }
+
+    private static boolean compareDirectoryContents(Path directory, Path compareDirectory, boolean checkFileContent)
+            throws IOException {
+
+        // LOGGER.info("compareDirectoryContents: " + directory);
+        try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(directory)) {
```
garydgregory: How does this implementation account for file and directories present in `compareDirectory` but not in `directory`? Seems quite broken. You'd be better off with names like `directory1` and `directory2` IMO.

garydgregory:
Did you copy from https://stackoverflow.com/questions/14522239/test-two-directory-trees-for-equality ? Are you the author? If not, you should at least link or credit.

jfiala:
Yes I was the author of that SO snippet back in 2017 :). Of course we can link to the original SO reply.
Thx for the review - I'll see when I can clean it up.

Best regards,
Johannes

garydgregory:
Please see git master for new APIs `PathUtils`. 
Jira: https://issues.apache.org/jira/browse/IO-648
This can be closed IMO.
Feedback welcome.

jfiala:
Thx alot for merging & improving.
PathUtils at master looks good :).

PR #32, opened 2017-02-23
Add Tailable interface to tail files accessed using alternative libraries such as jCIFS or commons-vfs

jonfreedman:
I have a derivation of org.apache.commons.io.input.Tailer in our codebase which allows tailing of remote files using different user credentials via [jCIFS](https://jcifs.samba.org/src/docs/api/).  It would be very nice if we could remove this duplicated code with the introduction of this Tailable interface.  It shouldalso be possible to tail files using commons-vfs e.g. [HdfsRandomAccessContent](https://commons.apache.org/proper/commons-vfs/apidocs/org/apache/commons/vfs2/provider/hdfs/HdfsRandomAccessContent.html),


jonfreedman:
It appears the test cases are fragile, the travis build has failed on one test with openjdk7 which passes on oraclejdk7 and oraclejdk8
```
FileFilterTestCase.testAgeFilter:856->assertFiltering:64 Filter(File) org.apache.commons.io.filefilter.AgeFileFilter not false for /home/travis/build/apache/commons-io/test/io/new.txt expected:<false> but was:<true>
```

PascalSchumacher:
jira issue: https://issues.apache.org/jira/browse/IO-533

garydgregory:
We cannot depend on jcifs IIRC, which is why it is still in the Commons VFS sandbox.

PascalSchumacher:
@garydgregory The title of the pull request is a bit misleading. It does not add a jcifs dependency. It adds an interfaces which the user can implement, e.g. to add tail jcifs files.

jonfreedman:
I've renamed the issue to hopefully clarify there is no jcifs or commons-vfs dependency.  For reference the pull request is functionally complete (as we've been running comparable code for a few months now).

garydgregory:
May you please rebase on master? New public and protected elements should have a Javadoc @since tag.

jonfreedman:
Is this now a candidate to be merged? It's over 4 years since I wrote this code and I've moved jobs twice since then!

garydgregory:
We are all volunteers here... I'm taking the time to review what's outstanding this weekend. If you are no longer interested, feel free to close this PR ;-)

jonfreedman:
I still think it's a valid change, I'll refresh my memory, rebase and update the java doc, I assume you want it to say @since 2.12.0?

garydgregory:
> I still think it's a valid change, I'll refresh my memory, rebase and update the java doc, I assume you want it to say @SInCE 2.12.0?

Correct and thank you.

jonfreedman:
Not sure what's going on here, I haven't changed 466 files...
See https://github.com/apache/commons-io/compare/master...jonfreedman:master for the diff

garydgregory:
I can't review 466 files... you might consider redoing the PR.

jonfreedman:
I will re-create my fork and submit another PR, not sure what that will do to this one.  ---- On Sun, 05 Sep 2021 22:55:11 +0100  ***@***.***  wrote ----
I can't review 466 files... you might consider redoing the PR.

jonfreedman:
OK, now this looks better

garydgregory:
Hi @jonfreedman,
Thank you for the reworked PR. You will need a new unit test for each of the new methods.

jonfreedman:
I've written a load of tests for the constructor/create overloads, my test is super simple and I am observing it occasionally fail with the `Tailer` reading the one line from the file twice. I've commited as I figured you may want to take a look at that...

jonfreedman:
I see the cause of the unstable tests, missed that `#create` starts a thread for you

src/main/java/org/apache/commons/io/input/Tailer.java
```diff
@@ -556,4 +784,181 @@ private long readLines(final RandomAccessFile reader) throws IOException {
...
+         * Tests if this tailable is newer than the specified time reference.
+         *
+         * @param timeMillis the time reference measured in milliseconds since the
+         *                   epoch (00:00:00 GMT, January 1, 1970).
+         * @return true if this tailable has been modified after the given time reference.
+         */
+        boolean isFileNewer(final long timeMillis);
+
+        /**
+         * @param mode the access mode {@link RandomAccessFile}
```
garydgregory, line 842: The 1st sentence is missing for this method.

src/main/java/org/apache/commons/io/input/Tailer.java
```diff
@@ -556,4 +784,181 @@ private long readLines(final RandomAccessFile reader) throws IOException {
...
     }
+
+    /**
+     * Abstraction on {@link java.io.File} which allows substitution of remote files for example using jCIFS
+     *
+     * @since 2.12.0
+     */
+    public interface Tailable {
+        /**
+         * @return  The name of the file denoted by this tailable
```
garydgregory, line 795: The 1st sentence is missing for this method.

src/main/java/org/apache/commons/io/input/Tailer.java
```diff
@@ -556,4 +784,181 @@ private long readLines(final RandomAccessFile reader) throws IOException {
...
+    }
+
+    /**
+     * Abstraction on {@link java.io.RandomAccessFile} which allows substitution of remote files for example using jCIFS
+     *
+     * @since 2.12.0
+     */
+    public interface RandomAccessTailable extends Closeable {
+        /**
+         * Returns the current offset in this tailable.
```
garydgregory, line 856: "Returns" -> "Gets"

src/main/java/org/apache/commons/io/input/Tailer.java
```diff
@@ -556,4 +784,181 @@ private long readLines(final RandomAccessFile reader) throws IOException {
...
+         * tailable, at which the next read or write occurs.  The offset may be
+         * set beyond the end of the tailable. Setting the offset beyond the end
+         * of the tailable does not change the tailable length.  The tailable
+         * length will change only by writing after the offset has been set beyond
+         * the end of the tailable.
+         *
+         * @param      pos   the offset position, measured in bytes from the
+         *                   beginning of the tailable, at which to set the
+         *                   tailable pointer.
+         * @exception  IOException  if {@code pos} is less than
```
garydgregory, line 875: "@exception" -> "@throws"

src/main/java/org/apache/commons/io/input/Tailer.java
```diff
@@ -556,4 +784,181 @@ private long readLines(final RandomAccessFile reader) throws IOException {
...
+
+        /**
+         * @return  The full path of this tailable
+         */
+        String getPath();
+
+        /**
+         * Returns the length of this tailable
+         *
+         * @return  The length, in bytes, of this tailable, or <code>0L</code>
```
garydgregory, line 807: <code>foo</code> -> {@code foo}

src/main/java/org/apache/commons/io/input/Tailer.java
```diff
@@ -556,4 +784,181 @@ private long readLines(final RandomAccessFile reader) throws IOException {
...
+        /**
+         * @param mode the access mode {@link RandomAccessFile}
+         * @return a random access file stream to read from
+         * @throws FileNotFoundException if the tailable object does not exist
+         */
+        RandomAccessTailable getRandomAccess(final String mode) throws FileNotFoundException;
+    }
+
+    /**
+     * Abstraction on {@link java.io.RandomAccessFile} which allows substitution of remote files for example using jCIFS
```
garydgregory, line 850: Sentences should end in a period.

src/main/java/org/apache/commons/io/input/Tailer.java
```diff
@@ -556,4 +784,181 @@ private long readLines(final RandomAccessFile reader) throws IOException {
...
+    public interface Tailable {
+        /**
+         * @return  The name of the file denoted by this tailable
+         */
+        String getName();
+
+        /**
+         * @return  The full path of this tailable
+         */
+        String getPath();
```
garydgregory, line 802: Confusing API name because it does not return a `Path`; either rename the API or make it return a `Path`.

src/main/java/org/apache/commons/io/input/Tailer.java
```diff
@@ -556,4 +784,181 @@ private long readLines(final RandomAccessFile reader) throws IOException {
...
+    /**
+     * Abstraction on {@link java.io.File} which allows substitution of remote files for example using jCIFS
+     *
+     * @since 2.12.0
+     */
+    public interface Tailable {
+        /**
+         * @return  The name of the file denoted by this tailable
+         */
+        String getName();
```
garydgregory, line 797: The name of... what? the base file name? The full file name?

src/main/java/org/apache/commons/io/input/Tailer.java
```diff
@@ -191,6 +203,18 @@ public Tailer(final File file, final TailerListener listener, final long delayMi
         this(file, listener, delayMillis, false);
     }
 
+    /**
```
garydgregory, line 206: There are way too many new factory APIs here IMO. For the first cut, let's add one and see what other use-cases require in the future. If a delay is provided, please make that a `Duration`, not a `long`. I'll redo the internals after this PR comes in to use a Duration instead of a `long`. If you add a constructor, make it private or package private.

garydgregory (changes requested):
@jonfreedman 
Thank you for the update.
Hi, I did a review and scattered my comments throughout.

jonfreedman:
I have a feeling I just lifted all the java doc from the underlying methods, they may have changed in the last 5 years and I'll also incorporate your suggestions. Regarding the constructors/factory methods, personally I would prefer to replace the whole lot with a builder but that's obviously quite a departure from the existing code, I had designed this PR to be as close as possible to the existing but I'm happy to make it 'better' if it would be accepted?  ---- On Mon, 06 Sep 2021 19:40:47 +0100  ***@***.***  wrote ----
@garydgregory requested changes on this pull request.

@jonfreedman
Thank you for the update.
Hi, I did a review and scattered my comments throughout.

In src/main/java/org/apache/commons/io/input/Tailer.java:
> +         * @return  <code>true</code> if and only if the tailable exists;
+         *          <code>false</code> otherwise
+         */
+        boolean exists();
+
+        /**
+         * Tests if this tailable is newer than the specified time reference.
+         *
+         * @param timeMillis the time reference measured in milliseconds since the
+         *                   epoch (00:00:00 GMT, January 1, 1970).
+         * @return true if this tailable has been modified after the given time reference.
+         */
+        boolean isFileNewer(final long timeMillis);
+
+        /**
+         * @param mode the access mode ***@***.*** RandomAccessFile}

The 1st sentence is missing for this method.



In src/main/java/org/apache/commons/io/input/Tailer.java:
> @@ -556,4 +784,181 @@ private long readLines(final RandomAccessFile reader) throws IOException {
             return rePos;
         }
     }
+
+    /**
+     * Abstraction on ***@***.*** java.io.File} which allows substitution of remote files for example using jCIFS
+     *
+     * @since 2.12.0
+     */
+    public interface Tailable {
+        /**
+         * @return  The name of the file denoted by this tailable

The 1st sentence is missing for this method.



In src/main/java/org/apache/commons/io/input/Tailer.java:
> +        /**
+         * @param mode the access mode ***@***.*** RandomAccessFile}
+         * @return a random access file stream to read from
+         * @throws FileNotFoundException if the tailable object does not exist
+         */
+        RandomAccessTailable getRandomAccess(final String mode) throws FileNotFoundException;
+    }
+
+    /**
+     * Abstraction on ***@***.*** java.io.RandomAccessFile} which allows substitution of remote files for example using jCIFS
+     *
+     * @since 2.12.0
+     */
+    public interface RandomAccessTailable extends Closeable {
+        /**
+         * Returns the current offset in this tailable.

"Returns" -> "Gets"



In src/main/java/org/apache/commons/io/input/Tailer.java:
> +         * @exception  IOException  if an I/O error occurs.
+         */
+        long getFilePointer() throws IOException;
+
+        /**
+         * Sets the file-pointer offset, measured from the beginning of this
+         * tailable, at which the next read or write occurs.  The offset may be
+         * set beyond the end of the tailable. Setting the offset beyond the end
+         * of the tailable does not change the tailable length.  The tailable
+         * length will change only by writing after the offset has been set beyond
+         * the end of the tailable.
+         *
+         * @param      pos   the offset position, measured in bytes from the
+         *                   beginning of the tailable, at which to set the
+         *                   tailable pointer.
+         * @exception  IOException  if ***@***.*** pos} is less than

***@***.***" -> ***@***.***"



In src/main/java/org/apache/commons/io/input/Tailer.java:
> +     */
+    public interface Tailable {
+        /**
+         * @return  The name of the file denoted by this tailable
+         */
+        String getName();
+
+        /**
+         * @return  The full path of this tailable
+         */
+        String getPath();
+
+        /**
+         * Returns the length of this tailable
+         *
+         * @return  The length, in bytes, of this tailable, or <code>0L</code>

foo -> ***@***.*** foo}



In src/main/java/org/apache/commons/io/input/Tailer.java:
> +         * @param timeMillis the time reference measured in milliseconds since the
+         *                   epoch (00:00:00 GMT, January 1, 1970).
+         * @return true if this tailable has been modified after the given time reference.
+         */
+        boolean isFileNewer(final long timeMillis);
+
+        /**
+         * @param mode the access mode ***@***.*** RandomAccessFile}
+         * @return a random access file stream to read from
+         * @throws FileNotFoundException if the tailable object does not exist
+         */
+        RandomAccessTailable getRandomAccess(final String mode) throws FileNotFoundException;
+    }
+
+    /**
+     * Abstraction on ***@***.*** java.io.RandomAccessFile} which allows substitution of remote files for example using jCIFS

Sentences should end in a period.



In src/main/java/org/apache/commons/io/input/Tailer.java:
> +
+    /**
+     * Abstraction on ***@***.*** java.io.File} which allows substitution of remote files for example using jCIFS
+     *
+     * @since 2.12.0
+     */
+    public interface Tailable {
+        /**
+         * @return  The name of the file denoted by this tailable
+         */
+        String getName();
+
+        /**
+         * @return  The full path of this tailable
+         */
+        String getPath();

Confusing API name because it does not return a Path; either rename the API or make it return a Path.



In src/main/java/org/apache/commons/io/input/Tailer.java:
> @@ -556,4 +784,181 @@ private long readLines(final RandomAccessFile reader) throws IOException {
             return rePos;
         }
     }
+
+    /**
+     * Abstraction on ***@***.*** java.io.File} which allows substitution of remote files for example using jCIFS
+     *
+     * @since 2.12.0
+     */
+    public interface Tailable {
+        /**
+         * @return  The name of the file denoted by this tailable
+         */
+        String getName();

The name of... what? the base file name? The full file name?



In src/main/java/org/apache/commons/io/input/Tailer.java:
> @@ -191,6 +203,18 @@ public Tailer(final File file, final TailerListener listener, final long delayMi
         this(file, listener, delayMillis, false);
     }
 
+    /**

There are way too many new factory APIs here IMO. For the first cut, let's add one and see what other use-cases require in the future. If a delay is provided, please make that a Duration, not a long. I'll redo the internals after this PR comes in to use a Duration instead of a long. If you add a constructor, make it private or package private.

garydgregory:
If you want to provide a builder, that's fine with me. Just be aware that we CANNOT break binary compatibility (mvn clean package japicmp:cmp)

garydgregory:
@jonfreedman 
I redid the guts to use a Duration instead of a long.
May you rebase on master please with your builder impl?

src/main/java/org/apache/commons/io/input/Tailer.java
jonfreedman: I've renamed this to `getFileName` more descriptive but now inconsistent with `java.io.File#getName`

src/main/java/org/apache/commons/io/input/Tailer.java
jonfreedman: I've renamed this to `getPathName` more descriptive but now inconsistent with `java.io.File#getPath`

garydgregory:
Hi @jonfreedman,

Thank you for your patience.
Please see git master for the new implementation. I committed the changes from this PR combined with:
- Use `Objects.requireNonNull()` instead manual checks.
- Normalize Javadoc.
- Add missing Javadoc.
- Add missing Javadoc tags.
- Sort members.
- Don't initialize ivars to defaults.
- Renamed some types, ivars, and params.
- Fix some Javadoc.
- No need to use FQCNs in Javadoc @link.
- Use `{@code}` instead of `<code></code>`.
- YAGNI: Remove some methods that are not tested and never called.
- Re-implement some internals using NIO.
- Add some constructor parameter validation.
- Add a toString() method.

May you please make sure your use case is satisfied?
If you want the code to do more, please rebase on master or provide a new PR.
TY.

jonfreedman:
Thanks for finally getting the change in there, it's a bit of a shame you weren't able to merge but the project has improved so I'll take what I can get.

PR #36, opened 2017-05-01
FileAlternationMonitor#stop() does not stop the thread

anthonyraymond:
The thread if FileAlterationMonitor wasn't stopped by the `stop(int)` method, which forbid application to shutdown unless all `Thread` are exited (if FileAlterationMonitor is part of a DI managed component).

This behavior conflict with the method javadoc `@param stopInterval the amount of time in milliseconds to wait for the thread to finish.`

### Simple example to understand why i changed the code

```java
    Thread t = new Thread(() -> {
        try {
            Thread.sleep(500000);
        } catch (final InterruptedException e) {
        }
    });
    t.start();
    t.join(50);
   // Ok, we reach this point until 500000ms are elapsed, but the thread is still alive.
   //   because Thread#join(int) does not kill the thread. And the thread remains alive.
```

```java
    Thread t = new Thread(() -> {
        try {
            Thread.sleep(500000);
        } catch (final InterruptedException e) {
        }
    });
    t.start();
    t.join(50);
    t.interupt();
   // Thread is exited
```
In this case, we waited the given time BEFORE exiting the `Thread`, as described in the javadoc, and the `Thread` is now finished and killed.

### DIFF
--- a/src/main/java/org/apache/commons/io/monitor/FileAlterationMonitor.java
+++ b/src/main/java/org/apache/commons/io/monitor/FileAlterationMonitor.java
@@ -23,7 +23,7 @@ import java.util.concurrent.ThreadFactory;
 /**
  * A runnable that spawns a monitoring thread triggering any
  * registered {@link FileAlterationObserver} at a specified interval.
- * 
+ *
  * @see FileAlterationObserver
  * @version $Id$
  * @since 2.0
@@ -112,7 +112,7 @@ public final class FileAlterationMonitor implements Runnable {
 
     /**
      * Returns the set of {@link FileAlterationObserver} registered with
-     * this monitor. 
+     * this monitor.
      *
      * @return The set of {@link FileAlterationObserver}
      */
@@ -165,8 +165,8 @@ public final class FileAlterationMonitor implements Runnable {
         running = false;
         try {
             thread.join(stopInterval);
+            thread.interrupt();
         } catch (final InterruptedException e) {
-            Thread.currentThread().interrupt();
         }
         for (final FileAlterationObserver observer : observers) {
             observer.destroy();


anthonyraymond:
Jira issue : https://issues.apache.org/jira/browse/IO-535

src/main/java/org/apache/commons/io/monitor/FileAlterationMonitor.java
```diff
@@ -165,8 +165,8 @@ public synchronized void stop(final long stopInterval) throws Exception {
         running = false;
         try {
             thread.join(stopInterval);
+            thread.interrupt();
         } catch (final InterruptedException e) {
-            Thread.currentThread().interrupt();
```
PascalSchumacher: Is swallowing an `InterruptedException` a really bad practice (at least according to `Java Concurrency in Practice` and https://www.ibm.com/developerworks/library/j-jtp05236/)?

anthonyraymond: Tell me if i'm wrong, because it is kind of obscure.

The monitor embed a thread. But when we call `.stop()` we are calling from another thread (most likely the main thread).
Isn't `Thread.currentThread().interrupt();` going to stop the calling thread? In our case the main thread?

Now, i agree that catching an exception to do nothing sounds dumb. But in our case we want to stop the inner thread. So even if it is interupted while performing the `join#` is kind of a good news. It just stopped earlier than expected.

Am i correct?

garydgregory: Regardless of the implementation of this fix, it would be great to provide a unit test that fails without the main patch. Otherwise, "confidence is low" and regression will happen.

anthonyraymond: Hi @garydgregory .
@SvetlinZarev was faster than i was: https://github.com/apache/commons-io/pull/58

PR #38, opened 2017-07-02
FileUtils.readFileToByteArray - optimize reading of files with known size

ilmarmors:
IO-251 added optimization for reading files with known size as byte arrays,  but it was reverted in IO-453 due to regression, because File.length() method may return 0 for path files denoting system-dependent entities such as devices or pipes

More optimal solution is to treat 0 as unknown size and use IOUtils.toByteArray(input), but if size is known (> 0) use more efficient approach and read content into pre-allocated byte array with exact size by calling IOUtils.toByteArray(input, size).


PascalSchumacher:
Jira issue: https://issues.apache.org/jira/browse/IO-542

PascalSchumacher:
Thanks! :+1:

PR #41, opened 2017-08-02
Missing Test Case for negative offset value

ileontiuc:
The method copyLarge(InputStream input, InputStream output, int offset, buffer) copies the content of input into output. If offset is a positive number, it will skip a number of bytes from input equal to offset value. By performing mutation testing analysis we noticed that the copyLarge method was never tested for a negative value for offset. The added test call copyLarge with a negative value and check that the behavior is the same as for offset = 0, where no bytes are skipped from input.


PascalSchumacher:
Thanks! :+1:

PR #42, opened 2017-08-18
[IO-546] ClosedOutputStream#flush should throw

tjcelaya:
While debugging an issue involving usage of `CloseShieldOutputStream` I discovered that `ClosedOutputStream#flush` is not overridden and will silently succeed. Not sure how much of a breaking change this might be but I think it makes more sense for `ClosedOutputStream#flush` to throw. This is only really meaningful in contexts where multiple streams are being chained together and some of the streams before `CloseShieldOutputStream` perform buffering, but it would make behavior more consistent for these more complex use-cases.

See [IO-546](https://issues.apache.org/jira/browse/IO-546) where @dekobon has attached an example program.


PascalSchumacher:
Thanks!

PR #43, opened 2017-09-15
Make code style of hasBOM() consistent with getBOMCharsetName()

mernst:
This is a minor, stylistic patch.

Background fact:  After the first call to `getBOM()`, the expressions `getBOM()` and `byteOrderMark` are equivalent.  (This fact isn't documented in the code, though.)

`getBOMCharsetName()` first calls `getBOM()` for side effect, then uses `byteOrderMark` within its body.

By contrast, `hasBOM()` mixes the two forms.

This patch changes `hasBOM()` to use the same code style as `getBOMCharsetName()`.  In addition to consistency throughout the file, the refactored code makes it obvious that there is no null pointer exception, without the reader having to re-derive the background fact.


britter:
JIRA ticket: https://issues.apache.org/jira/browse/IO-553

britter:
Thank you!

PR #45, opened 2017-10-09
Add an IfStatement to check the return value of dir.mkdirs().

BruceKuiLiu:
This statement returns a value that is not checked.
The return value should be checked since it can indicate an unusual or unexpected function execution.
The statement returns false if the destination directory could not be successfully created (rather than throwing an Exception).
If you don't check the result, you won't notice if the statement signals an unexpected behavior by returning an atypical return value.
http://findbugs.sourceforge.net/bugDescriptions.html#RV_RETURN_VALUE_IGNORED_BAD_PRACTICE


PascalSchumacher:
Travis build fail because of  a check-style violation:

```
[ERROR] src/main/java/org/apache/commons/io/FileUtils.java:[2962,13] (whitespace) FileTabCharacter: File contains tab characters (this is the first instance).
```

PR #46, opened 2017-10-13
IO-507: Rename ByteOrderUtils class to ByteOrderParser and remove some

britter:
logic for parsing strings "big" and "little", after discussions on the
ML.

### DIFF
--- a/src/main/java/org/apache/commons/io/ByteOrderParser.java
+++ b/src/main/java/org/apache/commons/io/ByteOrderParser.java
@@ -25,38 +25,26 @@ import java.util.Locale;
  *
  * @since 2.6
  */
-public final class ByteOrderUtils {
-
-    private static final Locale ComparisonLocale = Locale.ROOT;
-
-    /**
-     * Big endian.
-     */
-    public static final String BIG_ENDIAN = "Big";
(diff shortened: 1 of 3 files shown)


britter:
@garydgregory what do you think?

garydgregory:
+1

src/main/java/org/apache/commons/io/ByteOrderParser.java
```diff
@@ -25,38 +25,26 @@
...
 
     /**
      * ByteOrderUtils is a static utility class, so prevent construction with a private constructor.
      */
-    private ByteOrderUtils() {
+    private ByteOrderParser() {
     }
 
     /**
      * Parses the String argument as a {@link ByteOrder}, ignoring case.
```
PascalSchumacher: I think the pull request changes this, or I'm I wrong? Maybe we should keep the case insensitivity?

britter: You're right. I've changed the documentation. Let's make it restrictive and see if users want to have a more relaxed handling.

britter:
Fixed in master

PR #47, opened 2017-10-13
Check the return value of lockFile.delete().

Kui-Liu:
This statement returns a value that is not checked.
The return value should be checked since it can indicate an unusual or unexpected function execution.
The statement returns false if the destination file could not be successfully deleted (rather than throwing an Exception).
If you don't check the result, you won't notice if the statement signals an unexpected behavior by returning an atypical return value.
http://findbugs.sourceforge.net/bugDescriptions.html#RV_RETURN_VALUE_IGNORED_BAD_PRACTICE

### DIFF
--- a/src/main/java/org/apache/commons/io/output/LockableFileWriter.java
+++ b/src/main/java/org/apache/commons/io/output/LockableFileWriter.java
@@ -288,7 +288,10 @@ public class LockableFileWriter extends Writer {
         try {
             out.close();
         } finally {
-            lockFile.delete();
+            if (!lockFile.delete()) {
+                throw new IOException("Can't delete file, lock " +
+                         lockFile.getPath() + " exists");
+            }
         }
     }
 


BruceKuiLiu:
How about printing out a warning or error message, not throwing an IOException, if lockFile.delete() fails?

garydgregory:
Commons IO does not log ATM, to the console or anywhere. I do not think it
should for this case...

Gary

garydgregory:
Note that this PR would contradict the class' Javadoc:
```
/**
 * FileWriter that will create and honor lock files to allow simple
 * cross thread file lock handling.
 * <p>
 * This class provides a simple alternative to <code>FileWriter</code>
 * that will use a lock file to prevent duplicate writes.
 * <p>
 * <b>N.B.</b> the lock file is deleted when {@link #close()} is called
 * - or if the main file cannot be opened initially.
 * In the (unlikely) event that the lockfile cannot be deleted,
 * this is not reported, and subsequent requests using
 * the same lockfile will fail.
 * <p>
 * By default, the file will be overwritten, but this may be changed to append.
 * The lock directory may be specified, but defaults to the system property
 * <code>java.io.tmpdir</code>.
 * The encoding may also be specified, and defaults to the platform default.
 *
 */
```

garydgregory:
Closing as this was implemented, but differently.

PR #49, opened 2017-11-05
[IO-554] fixed: prevent input stream close

mmariotti:
[IO-554] FileUtils.copyToFile(InputStream source, File destination) closes input stream.
Fixed removing 'source' from try-with-resources.


jochenw:
For the sake of binary compatibility, I'd suggest an alternate proposal:

- Change the Javadocs of FileUtils.copyToFile to reflect the current situation.

- Introduce a new method copyToFile(InputStream, File, boolean), which is basically doing the same than the old implementation, except that the last parameter decides upon closing, or not.

PR #51, opened 2017-12-18
[IO-557] Perform locale independant upper case conversions.

luccioman:
To handle properly lower cased character encoding name in XML prolog
with any default system locale, notably Turkish which has specific rules for case conversion of dotted and dotless i characters.

Fixes issue [IO-557](https://issues.apache.org/jira/browse/IO-557).


PascalSchumacher:
Thanks! 👍

PR #52, opened 2017-12-21
IO-559 verify hostname part of suspected UNC paths in FileNameUtils

bodewig:
https://issues.apache.org/jira/browse/IO-559

I'm not 100% sure how/if Windows deals with percent encoded hostnames or hostnames using non-ASCII characters. Maybe anybody with deeper Windows knowledge can chime in.

### DIFF
--- a/src/main/java/org/apache/commons/io/FilenameUtils.java
+++ b/src/main/java/org/apache/commons/io/FilenameUtils.java
@@ -19,8 +19,12 @@ package org.apache.commons.io;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
 import java.util.Stack;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * General file name and file path manipulation utilities.
@@ -676,7 +680,9 @@ public class FilenameUtils {
             }
             posUnix = posUnix == NOT_FOUND ? posWin : posUnix;
             posWin = posWin == NOT_FOUND ? posUnix : posWin;
-            return Math.min(posUnix, posWin) + 1;
+            int pos = Math.min(posUnix, posWin) + 1;
+            String hostnamePart = fileName.substring(2, pos - 1);
+            return isValidHostName(hostnamePart) ? pos : NOT_FOUND;
         } else {
             return isSeparator(ch0) ? 1 : 0;
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/io/FilenameUtils.java
```diff
@@ -1490,4 +1496,163 @@ public static boolean wildcardMatch(final String fileName, final String wildcard
...
+     */
+    // mostly copied from org.apache.commons.validator.routines.InetAddressValidator#isValidInet4Address
+    private static boolean isIPv4Address(String name) {
+        Matcher m = IPV4_PATTERN.matcher(name);
+        if (!m.matches() || m.groupCount() != 4) {
+            return false;
+        }
+
+        // verify that address subgroups are legal
+        for (int i = 1; i < 5; i++) {
```
sebbASF: IMO it's more obvious what is happening to put
for (int i = 1; i <= 4; i++) {

src/main/java/org/apache/commons/io/FilenameUtils.java
```diff
@@ -1490,4 +1496,163 @@ public static boolean wildcardMatch(final String fileName, final String wildcard
...
+            String ipSegment = m.group(i);
+            if (ipSegment == null || ipSegment.length() == 0) {
+                return false;
```
sebbASF: Not possible, as the RE requires at least one digit

src/main/java/org/apache/commons/io/FilenameUtils.java
```diff
@@ -1490,4 +1496,163 @@ public static boolean wildcardMatch(final String fileName, final String wildcard
...
+            }
+
+            int iIpSegment = 0;
+
+            try {
+                iIpSegment = Integer.parseInt(ipSegment);
+            } catch(NumberFormatException e) {
```
sebbASF: I don't think a NFE is possible when parsing 1-3 digits

src/main/java/org/apache/commons/io/FilenameUtils.java
```diff
@@ -1490,4 +1496,163 @@ public static boolean wildcardMatch(final String fileName, final String wildcard
...
+                return false;
+            }
+
+            if (iIpSegment > IPV4_MAX_OCTET_VALUE) {
+                return false;
+            }
+
+            if (ipSegment.length() > 1 && ipSegment.startsWith("0")) {
+                return false;
```
sebbASF: Are leading zeros really disallowed?

src/main/java/org/apache/commons/io/FilenameUtils.java
```diff
@@ -1490,4 +1496,163 @@ public static boolean wildcardMatch(final String fileName, final String wildcard
...
+        return isIPv4Address(name) || isIPv6Address(name) || isRFC3986HostName(name);
+    }
+
+    private static final Pattern IPV4_PATTERN =
+        Pattern.compile("^(\\d{1,3})\\.(\\d{1,3})\\.(\\d{1,3})\\.(\\d{1,3})$");
+    private static final int IPV4_MAX_OCTET_VALUE = 255;
+
+    /**
+     * Checks whether a given string represents a valid IPv4 address.
+     *
```
sebbASF: Does the code really need to check for more than just 4 sets of 1-3 digits? If a user enters 999.000.999.000 does that affect the final outcome?

src/main/java/org/apache/commons/io/FilenameUtils.java
```diff
@@ -1490,4 +1496,163 @@ public static boolean wildcardMatch(final String fileName, final String wildcard
...
+
+    private static final int IPV6_MAX_HEX_GROUPS = 8;
+    private static final int IPV6_MAX_HEX_DIGITS_PER_GROUP = 4;
+    private static final int MAX_UNSIGNED_SHORT = 0xffff;
+    private static final int BASE_16 = 16;
+
+    // copied from org.apache.commons.validator.routines.InetAddressValidator#isValidInet6Address
+    /**
+     * Checks whether a given string represents a valid IPv6 address.
+     *
```
sebbASF: Again here: is detailed validation really necessary?

src/main/java/org/apache/commons/io/FilenameUtils.java
```diff
@@ -1490,4 +1496,163 @@ public static boolean wildcardMatch(final String fileName, final String wildcard
...
+     * RFC calls a "reg-name". Percent encoded names don't seem to be
+     * valid names in UNC paths.</p>
+     *
+     * @see "https://tools.ietf.org/html/rfc3986#section-3.2.2"
+     * @param name the hostname to validate
+     * @return true if the given name is a valid host name
+     */
+    private static boolean isValidHostName(String name) {
+        return isIPv4Address(name) || isIPv6Address(name) || isRFC3986HostName(name);
+    }
```
sebbASF: isIPv4Address can return false for input that looks like an IP address but fails a subsequent check, e.g. 999.9.9.9.
In which case it will be subject to two further checks (which must fail if the code is correct).

src/main/java/org/apache/commons/io/FilenameUtils.java
jochenw: Lets concentrate on the aspects, that were causing this PR: IMO, it is perfectly fine to accept the IPV4_PATTERN as harmless, even if there is a possibility, that an error like INVALID_HOSTNAME, or whatever, may come up later.

src/main/java/org/apache/commons/io/FilenameUtils.java
bodewig: I didn't want to modify the Validator code to much, so I didn't really spend any time cleaning it up much. But sure, I'd also think <= would be easier to read.

bodewig:
I've pushed two more commits which should address most of the issues found by @sebbASF

bodewig:
where are we with this? Shall I go ahead and merge the PR or is there anything left that still needs to be addressed?

PR #53, opened 2018-01-12
Modifier 'private' is redundant for enum constructors.

shahabkondri:
If no access modifier is specified for the constructor of an enum type, the constructor is private.


PascalSchumacher:
Thanks! 👍

PR #54, opened 2018-01-14
Code cleanup

shahabkondri:
Simplified conditional expression. Remove duplicate exceptions and redundant throws clause.


kinow:
Looks good to me! Much more readable! I only would need to consult the documentation (and perhaps other devs) regarding removing exception from public methods. I am not sure if that's OK for (binary&feature) backward compatibility.

garydgregory:
@shahabkondri 

PR #59, opened 2018-04-30
Tailer: speed-up by avoiding BAOS.toByteArray()

illes:


jochenw:
The original version is using a variable "charset". Your version ignores that variable?

illes:
No, it should not ignore it. `charset` is used in the overriden `toString()` method, see [Tailer.java:517](https://github.com/apache/commons-io/pull/59/commits/6a2c89c27fd19feb17c3ee76a29b830f23d02a9d#diff-4f30d3c239202f27bd0dbeb93f659378R517).

kinow:
I think I understood the change, and my first impression is that it would do the same avoiding the `toByteArray` call. However, I'd still be -0 for the change (meaning if others prefer to add it, please go ahead). 

My minor nit-picks is that we are overriding a method from `ByteArrayOutputStream`. (as a developer, I like to know whether my libraries are using lower level SDK API's if any, and how they are using). And I also think a developer could accidentally put back the `toByteArray` somewhere else, as it's possible to use both (we'd just have to note the way the current code is doing, but sometimes we accidentally use what we are used to).

But probably nit-picking, so as I said earlier, it's OK to merge it if others are OK :+1:

garydgregory:
Tailer has changed recently, may you please rebase on master?

garydgregory:
Closing, no reply since September of last year.

PR #61, opened 2018-05-31
Strengthen TeeOutputStremTest.testTee with an expected result.

oscarlvp:
Tee should not just repeat what's being written to both outputs, the output should also be the same as expected. For example, if the body of any of the `write` methods in TeeOutputStream is removed,
then the values written are the same (nothing) and the assertions don't fail in the test. This is solved by adding an expected value and checking both outputs against it.


PascalSchumacher:
Thanks! 👍

PascalSchumacher:
Merged in https://github.com/apache/commons-io/commit/9e801d916c3d163a4cfcc822e50dd75ac7cefcc5

It would be nice if you could close this pull request. Thanks!

PR #64, opened 2018-08-31
[IO-634] Make getCause synchronized and use a Deque instead of a Stack

wilx:


src/main/java/org/apache/commons/io/TaggedIOException.java
```diff
@@ -129,7 +129,7 @@ public Serializable getTag() {
      * @return wrapped exception
      */
     @Override
-    public IOException getCause() {
+    public synchronized IOException getCause() {
```
kinow: Interesting! Didn't know `Throwable`'s `getCuase` was `synchronized`.

wilx:
Ping.

src/main/java/org/apache/commons/io/FilenameUtils.java
```diff
@@ -1385,11 +1386,11 @@ public static boolean wildcardMatch(final String fileName, final String wildcard
         boolean anyChars = false;
         int textIdx = 0;
         int wcsIdx = 0;
-        final Stack<int[]> backtrack = new Stack<>();
+        final Deque<int[]> backtrack = new ArrayDeque<>(wcs.length);
```
kinow: Looks reasonable to me. `Stack` is synchronized I think, while `Deque` is not. But `backtrack` is used in this method without any locks, so `Stack` doesn't appear to be used specifically for that.

kinow:
@wilx could you rebase your code on `master`, please? I had a look at Travis, and the build appears to be passing there, while failing for this PR. Hopefully something that was fixed since you created the PR. Then will ping another dev to take a look at the changes (I guess they are small but still deserve an issue for the changelog perhaps?) Thanks

kinow:
Thanks @wilx ! Build passing on Travis. @garydgregory I haven't looked at commons-io in a while. Would you be able to have another look at this PR and double check if it looks OK, and whether we'd need a JIRA for this or not, please? (no worries if busy, I can drop an e-mail to dev-list later this weekend)

garydgregory:
Yes, we should have a JIRA and this PR needs a description.

kinow:
Thanks @garydgregory . @wilx could you create a JIRA and rename the PR, please?

wilx:
I will but TBH, reluctantly.

wilx:
JIRA issue here: <https://issues.apache.org/jira/browse/IO-634>

kinow:
That's just to have the issue in the changelog (when merging we'll add it to `changes.xml`), so other users will be aware of the changes and know what to look for in case of bugs - even if very unlikely, it can be helpful to see what changed/why.

The title is still vague. What if the PR/issue have a title like "Make getCause synchronized and use a Deque instead of Stack" ? Or something along that line.

PR #65, opened 2018-10-10
Performance Enhancement: Call toArray with Zero Array Size

DaGeRe:
toArray should, at least since Java 7, be used with new T[0] instead of size, since it is "faster, safer, and contractually cleaner" (https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion)

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -454,7 +454,7 @@ public class FileUtils {
      * @return an array of java.io.File
      */
     public static File[] convertFileCollectionToFileArray(final Collection<File> files) {
-        return files.toArray(new File[files.size()]);
+        return files.toArray(new File[0]);
     }
 
     //-----------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


kinow:
Hi @DaGeRe ! I don't know if anything changed in Java 10+, but I think this is a good change, and will mark it as approved. Alas I can't merge it straight away due to the failing Travis-CI builds. I am not certain that they are related to your change, but I suspect they are not (master is broken too, but there are issues in progress working to fix it).

Thanks!
Bruno

namannigam:
@kinow In Java-11, there has been an introduction to an API `default <T> T[] toArray(IntFunction<T[]> generator)` which does internally calls the same 0 sized array initialisation. 
The construct for example would in that case be:
`list.toArray(String[]::new)`
instead of 
`list.toArray(new String[0])`

kinow:
Thanks @namannigam . So I believe so far the overall consensus here is that the change in the pull request is OK. Let's give it some more time to wait for at least another developer to review it.

PascalSchumacher:
Merged. Thank you very much for the pull request and sorry for the delay.

PR #66, opened 2018-10-14
Avoid Code Duplication: Reuse Sleep from ThreadMonitor

DaGeRe:
It makes no sense to duplicate code, like the sleep method, just to keep the interface private.

### DIFF
--- a/src/main/java/org/apache/commons/io/ThreadMonitor.java
+++ b/src/main/java/org/apache/commons/io/ThreadMonitor.java
@@ -37,7 +37,7 @@ package org.apache.commons.io;
  * </pre>
  *
  */
-class ThreadMonitor implements Runnable {
+public class ThreadMonitor implements Runnable {
 
     private final Thread thread;
     private final long timeout;
@@ -121,7 +121,7 @@ class ThreadMonitor implements Runnable {
      * @param ms the number of milliseconds to sleep for
      * @throws InterruptedException if interrupted
      */
-    private static void sleep(final long ms) throws InterruptedException {
+    public static void sleep(final long ms) throws InterruptedException {
         final long finishAt = System.currentTimeMillis() + ms;
         long remaining = ms;
         do {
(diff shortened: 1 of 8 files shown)


kinow:
Hi @DaGeRe !

Agree that code duplication is bad. But the change could perhaps be changed a little bit to make it easier to accept it in my opinion.

In `ThreadMonitor`, when you make it public, it also means that we are telling users that they are free to use this class in their code as well. Instead of it being hidden in the package level (which doesn't necessarily stops them, but discourages and at least raises the question of whether they should use it).

But I agree with the remaining changes (except by one incorrect indentation that may raise a checkstyle issue).

Would you like to try, perhaps, creating a test class in the same package, but under the test sources? That way you can either wrap or extend the origin `ThreadMonitor` and still remove the duplication in the `sleep` method. Also making it easier for us to maintain the code :-)

Cheers
Bruno

garydgregory:
> Hi @DaGeRe !
> 
> Agree that code duplication is bad. But the change could perhaps be changed a little bit to make it easier to accept it in my opinion.
> 
> In `ThreadMonitor`, when you make it public, it also means that we are telling users that they are free to use this class in their code as well. Instead of it being hidden in the package level (which doesn't necessarily stops them, but discourages and at least raises the question of whether they should use it).
> 
> But I agree with the remaining changes (except by one incorrect indentation that may raise a checkstyle issue).
> 
> Would you like to try, perhaps, creating a test class in the same package, but under the test sources? That way you can either wrap or extend the origin `ThreadMonitor` and still remove the duplication in the `sleep` method. Also making it easier for us to maintain the code :-)
> 
> Cheers
> Bruno

Right, this class should stay package private.
You can either subclass it in the test tree as a public class or provide a public test method that accesses the package private one.

garydgregory:
I've made simpler changes. Please see git master.

garydgregory:
@DaGeRe ping?

PR #67, opened 2018-10-16
IO-170: Scalable Iterator for files, better than FileUtils.iterateFiles

detinho:
I took an aproach of implementing a separate class with the iterator logic.
While this brings a little code duplication, it leads to a more scalable approach, as the iterator class processes only the necessary files and directories to return the next file.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -51,6 +51,7 @@ import org.apache.commons.io.filefilter.FileFilterUtils;
 import org.apache.commons.io.filefilter.IOFileFilter;
 import org.apache.commons.io.filefilter.SuffixFileFilter;
 import org.apache.commons.io.filefilter.TrueFileFilter;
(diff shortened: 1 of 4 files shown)


garydgregory:
Why would we one these "old" APIs instead of java.nio.file.Files.walk(Path, int, FileVisitOption...) and friends? Should our Javadocs be updated instead?

detinho:
@garydgregory in that case it would be Files.walkTree, as Files.walk was introduced on Java 1.8.
Regarding Files.walkTree, it would be possible to use it instead FileUtils.iterateFiles, but not with an Iterator pattern.
I think it would be nice to update the documentation too, so users can be aware of Files.walk in case they are using Java 1.8 already and use the Stream pattern, what do you think?

garydgregory:
Hi @detinho 
This code is likely obsolete due to (1) the git master rework that uses Java NIO's streaming APIs and (2) this code does not use NIO streaming APIs.
May you please review git master, comment and/or update/close this PR?
TY!

garydgregory:
Closing, see my previous comment.

PR #62, opened 2018-06-01
IO-578: Support java.nio.Path and non-default file systems for ReversedLinesFileReader

mches:
JIRA Issue: [IO-578](https://issues.apache.org/jira/browse/IO-578) "ReversedLinesFileReader cannot be used with non-default file systems on Java 7+"

It's not currently possible to use `ReversedLinesFileReader` with a non-default file system like Jimfs. The file would first have to be copied to the default file system.

Non-default file systems can be supported by preferring `java.nio.file.Path` and `java.nio.channels.SeekableByteChannel` to `java.io.File` and `java.io.RandomAccessFile`.

PR #68, opened 2018-10-29
Output all parameters in ReversedLinesFileReaderTestParamFile

august782:
In ReversedLinesFileReaderTestParamFile, the parameterized tests have three parameters, but currently the test class is configured to only output the first two. Outputting only the first two parameters is not enough to distinguish the different parameterized tests, particularly for the ones whose first two parameters are "test-file-utf8-win-linebr.bin", "UTF-8". The proposed fix is to output all three parameters as part of the reporting of the tests.


mches:
@august782 I noticed that too and fixed [here ](https://github.com/apache/commons-io/pull/62/files)while adding a 4th parameter

garydgregory:
@august782 
Thank you for your PR but it no longer seems to apply based on git master. Closing.

PR #72, opened 2018-12-26
[IO-596] Add DeleteFiles utility class

jvz:
This class provides configurable strategies to delete files and directories.

Adapted from the work done in https://github.com/jenkinsci/jenkins/pull/3812.

Signed-off-by: Matt Sicker <boards@gmail.com>

@reviewbybees @garydgregory @jeffret-b

### DIFF
--- a/src/main/java/org/apache/commons/io/CompositeIOException.java
+++ b/src/main/java/org/apache/commons/io/CompositeIOException.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.commons.io;
+
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Combines one or more IOExceptions into a single IOException.
+ */
+public class CompositeIOException extends IOException {
+    private static final long serialVersionUID = 1L;
+
+    private final IOException[] causes;
+
+    public CompositeIOException(final String message, final List<IOException> causes) {
+        super(message);
+        this.causes = causes.toArray(new IOException[0]);
+    }
+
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/commons/io/DeleteFiles.java
```diff
@@ -0,0 +1,339 @@
...
+            final List<IOException> errors = tryDeleteRecursive(path);
+            if (errors.isEmpty()) {
+                return;
+            }
+            accumulatedErrors.addAll(errors);
+            if (waitToRetry(retriesAttempted)) {
+                throw new CompositeIOException(failureMessage(path, retriesAttempted, true), accumulatedErrors);
+            }
+        }
+        throw new CompositeIOException(failureMessage(path, maxRetries, false), accumulatedErrors);
```
asciborek: From what I can see this code will always cause throwing a new exception

jvz: Line 80 with the `return` will terminate the loop. This only returns successfully if none of the subpaths threw an exception.

src/main/java/org/apache/commons/io/CompositeIOException.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public CompositeIOException(final String message, final IOException... causes) {
+        super(message);
+        this.causes = causes.clone();
+    }
+
+    public IOException[] getCauses() {
+        return causes.clone();
+    }
+}
```
garydgregory, line 44: Note that we have `IOExceptionList`.

garydgregory:
Some of this PR is likely obsolete due to the newer PathUtils delete methods. Please review these and perhaps propose enhancements?

garydgregory:
Closing this PR, use version 2.7's `PathUtils` methods:
- `delete(Path)`
- `delete(Path, DeleteOption...)`
- `deleteDirectory(Path)`
- `deleteDirectory(Path, DeleteOption...)`
- `deleteFile(Path)`
- `deleteFile(Path, DeleteOption...)`

jvz:
Breaking up into execution strategies sounds like a good idea! Such a strategy could likely adopt a similar API to retry libraries like retry4j.

garydgregory:
@jvz 
Nice idea Matt :-) let's discuss on the ML where a retry API belongs.

garydgregory:
On Wed, Sep 2, 2020 at 10:57 AM Matt Sicker <notifications@github.com>
wrote:

> Breaking up into execution strategies sounds like a good idea! Such a
> strategy could likely adopt a similar API to retry libraries like retry4j.
>

I am cutting 2.8.0-RC2 now, so let's think about this for 2.9.0.

Gary

garydgregory:
Another item: values like `waitBetweenRetriesMillis` are error prone in my experience. Over at HttpComponents, we have classes called TimeValue and Timeout that wrap a `long` and a `TimeUnit`, and I'd like to bring that in to Apache Commons, someplace...

garydgregory:
Some of this is superseded by the new(ish) code in PathUtils.delete methods and associated utilities. If we want a retry (count/duration delay) feature, let's build it on top of the newer code.

PR #75, opened 2019-03-16
Addition of tests for FileUtils

Kled9:
Add coverage for FileUtils.openInputStream, FileUtils.copyFile, FileUtils.openOutputStream, FileUtils.copyDirectoryToDirectory.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -350,7 +350,7 @@ public class FileUtils {
             if (file.isDirectory()) {
                 throw new IOException("File '" + file + "' exists but is a directory");
             }
-            if (file.canWrite() == false) {
+            if (!file.canWrite()) {
                 throw new IOException("File '" + file + "' cannot be written to");
             }
         } else {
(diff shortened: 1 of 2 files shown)


garydgregory:
Please rebase on master.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -1176,20 +1178,20 @@ private static void doCopyFile(final File srcFile, final File destFile, final bo
...
-     * @throws NullPointerException if source or destination is {@code null}
+     * @throws IllegalArgumentException if source or destination is {@code null}
      * @throws IOException          if source or destination is invalid
      * @throws IOException          if an IO error occurs during copying
      * @since 1.2
      */
     public static void copyDirectoryToDirectory(final File srcDir, final File destDir) throws IOException {
         if (srcDir == null) {
-            throw new NullPointerException("Source must not be null");
+            throw new IllegalArgumentException("Source must not be null");
```
garydgregory: Do not change these please. Since Java 8 and the broad use of [java.util.Objects.requireNonNull(T, String)](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.lang.String-), any edits to current throws of NPE should use the aforementioned API.

PR #80, opened 2019-07-01
Remove throws IOException in method isSymlink()

ryanthara:
The `IOException` is caught in `Files.isSymbolicLink(Path path)` and therefore not thrown anymore in `public static boolean isSymlink(final File file)`.


garydgregory:
-1. This commit does not even compile: https://travis-ci.org/apache/commons-io/jobs/552735671

PR #84, opened 2019-08-01
Added ClosedReader and CloseShieldReader.

robtimus:
Use ClosedInputStream.CLOSED_INPUT_STREAM in CloseShieldInputStream instead of creating a new one.


garydgregory:
Tracking https://issues.apache.org/jira/browse/IO-613

PR #85, opened 2019-08-01
Added TaggedReader, ClosedReader and BrokenReader.

robtimus:


robtimus:
The new code has 100% code coverage except for the BrokenReader constructor that creates its own IOException.

PR #86, opened 2019-08-01
Added TaggedWriter, ClosedWriter and BrokenWriter.

robtimus:


robtimus:
The new code has 100% code coverage except for the BrokenWriter constructor that creates its own IOException.

PR #87, opened 2019-08-01
Added AppendableWriter.

robtimus:


robtimus:
The new code has 100% code coverage except for the close() and flush() methods, which have no body and therefore should not need any testing.

PR #89, opened 2019-08-02
Added TeeReader.

robtimus:


src/main/java/org/apache/commons/io/input/TeeReader.java
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * Reader proxy that transparently writes a copy of all characters read
+ * from the proxied reader to a given Reader. Using {@link #skip(long)}
+ * or {@link #mark(int)}/{@link #reset()} on the reader will result on some
+ * characters from the reader being skipped or duplicated in the writer.
+ * <p>
+ * The proxied reader is closed when the {@link #close()} method is
+ * called on this proxy. It is configurable whether the associated writer
+ * will also closed.
+ *
```
garydgregory: Please close HTML tags. IOW </p>

src/main/java/org/apache/commons/io/input/TeeReader.java
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * Reader proxy that transparently writes a copy of all characters read
+ * from the proxied reader to a given Reader. Using {@link #skip(long)}
+ * or {@link #mark(int)}/{@link #reset()} on the reader will result on some
+ * characters from the reader being skipped or duplicated in the writer.
+ * <p>
+ * The proxied reader is closed when the {@link #close()} method is
+ * called on this proxy. It is configurable whether the associated writer
+ * will also closed.
```
garydgregory: Replace:

 It is configurable whether the associated writer
		 * will also closed.

With:

You may configure whether the reader closes the writer.

src/main/java/org/apache/commons/io/input/TeeReader.java
```diff
@@ -0,0 +1,180 @@
...
+public class TeeReader extends ProxyReader {
+
+    /**
+     * The writer that will receive a copy of all characters read from the
+     * proxied reader.
+     */
+    private final Writer branch;
+
+    /**
+     * Flag for closing also the associated writer when this
```
garydgregory: Remove "also"

src/main/java/org/apache/commons/io/input/TeeReader.java
```diff
@@ -0,0 +1,180 @@
...
+     * @param st start offset within the buffer
+     * @param end maximum number of characters to read
+     * @return number of characters read, or -1 if the reader has ended
+     * @throws IOException if the reader could not be read (or written)
+     */
+    @Override
+    public int read(final char[] chr, final int st, final int end) throws IOException {
+        final int n = super.read(chr, st, end);
+        if (n != -1) {
+            branch.write(chr, st, n);
```
garydgregory: Use EOF in the test.

src/main/java/org/apache/commons/io/input/TeeReader.java
robtimus: Fixed, also in TeeInputStream where I copied the text from.

src/main/java/org/apache/commons/io/input/TeeReader.java
robtimus: Fixed, also in TeeInputStream where I copied the text from.

src/main/java/org/apache/commons/io/input/TeeReader.java
robtimus: Fixed, also in TeeInputStream where I copied the text from.

src/main/java/org/apache/commons/io/input/TeeReader.java
robtimus: Fixed, also in TeeInputStream where I copied the text from.

garydgregory:
@robtimus 

If you run rebase your PR on master, run 'mvn clean install site', and then look at the JaCoCo report you will see that the `close()` method has 0% code coverage by the tests.
May you address this please?

You should follow these steps for each PR. It is sometimes hard to get 100% code coverage for new code but it is a worthy goal nonetheless and in this case at least, easily achievable :-)

Thank you!
Gary

robtimus:
I've added some unit tests for both TeeReader and TeeInputStream, because this too had 0% code coverage on close().

src/test/java/org/apache/commons/io/input/TeeInputStreamTest.java
```diff
@@ -30,6 +32,52 @@
  */
 public class TeeInputStreamTest  {
 
+    private static class ExceptionOnCloseByteArrayInputStream extends ByteArrayInputStream {
```
garydgregory: @robtimus ,

Instead of creating these one-off classes, you should reuse and/or possibly refactor `YellOnFlushAndCloseOutputStream`, `ClosedOutputStream`, or `CloseShieldOutputStream`.

May you please give that a try?

Gary

src/test/java/org/apache/commons/io/input/TeeReaderTest.java
```diff
@@ -30,6 +34,54 @@
  */
 public class TeeReaderTest  {
 
+    // Cannot use StringReader or CharSequenceReader as super class because these removed the throws clause from close()
```
garydgregory: @robtimus 
Same kind of comment as above, maybe classes these can be made as "Reader" versions of the "OutputStream" classes mentioned above for reuse.

src/test/java/org/apache/commons/io/input/TeeInputStreamTest.java
robtimus: These classes are based on similar classes in `TeeOutputStreamTest`. While I could change the code to use `YellOnFlushAndCloseOutputStream`, I don't think that `ClosedOutputStream` or `CloseShieldOutputStream` is feasible, as these don't give access to its state; the test becomes more difficult to read because the check on whether or not `close()` has been called would require a try-fail-catch statement.

garydgregory:
Hi @robtimus ,

I pushed a modified version of this patch to `master`. Please rebase your other PRs on master and note the use in my push of Mockito instead of the addition of the static reader/writer/inputstream/outputstream. Also note the addition of new test helpers in `org.apache.commons.io.testtools`. Please adapt your other PRs if need be.

Tracked in Apache's Jira as https://issues.apache.org/jira/browse/IO-612

Thank you!
Gary

robtimus:
I have merged master into my other PRs, and updated a few tests to use Mockito and the Yell classes. I've also merged master into this PR but it's now identical (just for completeness sake).

PR #90, opened 2019-08-10
Use getChars method for String, StringBuilder and StringBuffer in Cha…

robtimus:
…rSequenceReader.read(char[], int, int)


garydgregory:
You will likely want to rebase #91 from master.

PR #92, opened 2019-08-12
simply code

Mixpa:
hello I found two places that can be simplified when I scanned the code with idea.


garydgregory:
This feels to me like a style issue. If we really want to use this style, we should do it for the whole component, not just a these seemingly random two files.

Mixpa:
Ok, I just think this can increase the readability of the code.

PR #93, opened 2019-08-15
Control flow style

Mixpa:
simplify code

### DIFF
--- a/src/main/java/org/apache/commons/io/IOUtils.java
+++ b/src/main/java/org/apache/commons/io/IOUtils.java
@@ -778,11 +778,11 @@ public class IOUtils {
 
         String line1 = br1.readLine();
         String line2 = br2.readLine();
-        while (line1 != null && line2 != null && line1.equals(line2)) {
+        while (line1 != null && line1.equals(line2)) {
             line1 = br1.readLine();
             line2 = br2.readLine();
         }
-        return line1 == null ? line2 == null ? true : false : line1.equals(line2);
+        return line1 == null && line2 == null;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -778,11 +778,11 @@ public static boolean contentEqualsIgnoreEOL(final Reader input1, final Reader i
 
         String line1 = br1.readLine();
         String line2 = br2.readLine();
-        while (line1 != null && line2 != null && line1.equals(line2)) {
+        while (line1 != null && line1.equals(line2)) {
             line1 = br1.readLine();
             line2 = br2.readLine();
         }
-        return line1 == null ? line2 == null ? true : false : line1.equals(line2);
+        return line1 == null && line2 == null;
```
garydgregory: What about the `line1.equals(line2)` bit?

Mixpa: yes,  you way is better than me

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -778,11 +778,11 @@ public static boolean contentEqualsIgnoreEOL(final Reader input1, final Reader i
 
         String line1 = br1.readLine();
         String line2 = br2.readLine();
-        while (line1 != null && line2 != null && line1.equals(line2)) {
+        while (line1 != null && line1.equals(line2)) {
             line1 = br1.readLine();
             line2 = br2.readLine();
         }
-        return line1 == null ? line2 == null ? true : false : line1.equals(line2);
+        return line1.equals(line2);
```
Stzx: line1 may be null.

Mixpa: how about Objects.equals

Mixpa:
@garydgregory

PR #96, opened 2019-09-10
Fixed a mistake in the FilenameUtils.concat()'s Javadoc about an absolute path.

yuji-k64613:
ng
/foo/a/ + ../bar     -->   foo/bar

ok
/foo/a/ + ../bar     -->   /foo/bar

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -134,7 +134,7 @@ The <action> type attribute can be add,update,fix,remove.
       <action issue="IO-625" dev="ggregory" type="fix" due-to="Mikko Maunu">
         Corrected misleading exception message for FileUtils.copyDirectoryToDirectory.
       </action>
-      <action issue="IO-626" dev="Yuji Konishi" type="fix" due-to="Yuji Konishi">
+      <action issue="IO-626" dev="ggregory" type="fix" due-to="Yuji Konishi">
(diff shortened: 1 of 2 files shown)


yuji-k64613:
The Travis CI build failed, what should I do?

Stzx:
This should be a fake failure.

garydgregory:
This could be a random failure and certainly should not be affected by a Javadoc changes.

@yuji-k64613 : If you want to be credited, please update `src/changes/changes.xml`, otherwise, please say if you do not here.

src/changes/changes.xml
```diff
@@ -134,6 +134,9 @@ The <action> type attribute can be add,update,fix,remove.
       <action issue="IO-625" dev="ggregory" type="fix" due-to="Mikko Maunu">
         Corrected misleading exception message for FileUtils.copyDirectoryToDirectory.
       </action>
+      <action issue="IO-626" dev="Yuji Konishi" type="fix" due-to="Yuji Konishi">
```
Stzx: Normally the **dev** is a committer, and **due-to** is the PR contributor name/alias.

yuji-k64613: Sorry.

I thought that a committer and the PR contributor are me.
Which is my mistake, a committer or the PR contributor?

PR #97, opened 2019-09-30
IO-628: Migration to JUnit Jupiter

mureinik:
This PR upgrades the project's testing framework from JUnit 4.12 to the modern JUnit Jupiter 5.5.4.

Since JUnit 5 Jupiter is not backwards compatible to JUnit 4.x (or even JUnit Vintage), this PR is a bit large, even though a lot of the changes are merely cosmetic (such as changing the argument order,
see details below). In order to make the reviewer's task as easy as possible, this PR does not presume to use JUnit Jupiter's best practices and all its new functionality, but only to migrate the existing tests with as little change as possible. Following PRs may want to improve the tests by using some of JUnit Jupiter's new features.

This PR includes the following changes:

1. Maven dependency changes:
    1. `junit:junit` was replaced with `org.junit.jupiter:junit-jupiter`.
    1. `org.junit-pioneer:junit-pioneer` was introduced (see details below in section 2.vii.).
    1. The JaCoCo report was disabled for JDK-EA by adding the `jacoco.skip` property via an environment variable set by Travis CI. Note that this report fails on the master branch too (e.g., see errors inline in https://travis-ci.org/apache/commons-io/jobs/591248951), but does not fail the build. With JUnit Jupiter there seem to be more of these errors, and the jdk-ea build fails since the log file is exhausted (although all the test pass!). In order to allow the build to pass, this (already broken) report was disabled.

2. Annotations:
    1. `org.junit.jupiter.api.Test` was used as a drop in replacement for `org.juit.Test` without arguments. See 3.ii. and 3.iii. for handling of `@Test` annotations with `expected` and `timeout` arguments respectively.
    1. `org.junit.jupiter.params.ParameterizedTest` in conjunction with `org.junit.jupiter.params.provider.MethodSource` was used to replace `org.juit.Test` in test classes run with `org.junit.runners.Parameterized`.
    1. `org.junit.runners.Parameterized.Parameters` annotations were removed, and the methods annotated with them were written to match the signature expected of a `@MethodSource`.
    1. `org.junit.jupiter.api.BeforeEach` was used as an drop in replacement for `org.junit.Before`.
    1. `org.junit.jupiter.api.AfterEach` was used as a drop in replacement for `org.junit.After`.
    1. `org.junit.jupiter.api.Disabled` was used as a drop in replacement for `org.junit.Ignore`.
    1. `org.junitpioneer.jupiter.DefaultLocale` was used as a drop in replacement for `org.apache.commons.io.testtools.SystemDefaults`. This annotation comes from the JUnit Pioneer project, and while it isn't part of the actual JUnit project, it's widely accepted as an extension library for JUnit, and is used in other Apache Commons projects, such as Apache Commons Lang.

3. Assertions:
    1. `org.junit.jupiter.api.Assertions`' methods were used as drop in replacements for `org.junit.Assert`'s methods with the same name in the simple case of an assertion without a message.  In the case of an assertion with a message, `org.junit.jupiter.api.Assertions`' methods were used, but the argument order was changed - `Assert`'s methods take the message as the first argument, while `Assertions`' methods take the message as the last argument.
    1. `org.junit.jupiter.api.Assertions#assertThrows` was used to assert that a specific exception was throws instead of an `org.junit.Test` annotation with an `expected` argument. This technique has a side bonus of making the tests slightly stricter, as now they can assert the exception was thrown from a specific line and prevent false positives where the test's "set-up" code accidentally threw that exception. The `throws` clauses of these methods were cleaned up from exceptions that can no longer be thrown in order to avoid compilation warnings.
    1. `org.junit.jupiter.api.Assertions#assertTimeout` was used to assert a specific block of code completes before a timeout is reach instead of an `org.junit.Test` annotation with a `timeout` argument.
    1. `org.junit.jupiter.api.Assertions#fail` was used in order to explicitly fail a test instead of throwing a `junit.framework.AssertionFailedError`.

4. Assumptions
    1. `org.junit.jupiter.api.Assumptions`' methods were used as drop in replacements for `org.junit.Assume`'s methods with the same name, but the argument order was changed - `Assume`'s methods take the message as the first argument, while `Assumptions`' methods take the message as the last argument.

5. Rules
    1. `org.junit.jupiter.api.io.TempDir` was used instead of the `org.junit.rules.TemporaryFolder` `@Rule`, and the code that used them was adapted accordingly. Note that `TempDir` is stricter than the old `TemporaryFolder` rule, and will fail the test if any file in the temporary directory cannot be deleted, and thus some tests required specific changes. See 6.ii. and 6.iii. for additional details.
    1. `org.apache.commons.io.testtools.SystemDefaultsSwitch` was removed from the project, and its functionality replaced by `org.junitpioneer.jupiter.DefaultLocale`.

6. Specific changes:
    1. Parameters in JUnit Jupiter (such as those produced by `MethodSource`) can only be applied to `@Test` methods, not to other lifecycle methods, such as `@BeforeEach`. Thus, the design of `ReversedLinesFileReaderTestParamFile` had to be changed - instead of receiving constructor arguments and storing a state in data members, it was rewritten to be stateless, and manage all its resources in the single test method it has, `testDataIntegrityWithBufferedReader`. As a side bonus, this change allows to narrow the scope of open resources, and close them immediately after they are no longer needed, instead of preserving them in data members until the `@After` method is called.
    1.  `FileUtilsCleanDirectoryTestCase`'s methods `testThrowsOnNullList` and `testThrowsOnCannotDeleteFile` chmod the temporary directory and thus prevent files in it from being deleted at the end of the test. A `finally` block was added to these tests in order to chmod the temporary directory back to 755 and allow it to be deleted. In addition, the restriction of not running the test on Windows was changed to use JUnit Jupiter's `@DisabledOnOs` annotation, in order to make the test a tad easier to follow.
    1. `TailerTest` has a similar issue to 6.iii., where threads opened during the test are only closed in the `@AfterEach` method (by calling `Tailer#stop`), which is too late, since the `@TempDir`'s tear down is called (and fails) beforehand. Using a shared temp dir (by declaring the member `static`) causes the directory to be deleted only when the entire class is finished.


mureinik:
reviewers - the CI fails, so this PR is clearly not OK, even though the test work on my machine. Looking in to this now.

mureinik:
Reviewers - while disabling the JaCoCo report (change 1.iii described above) is required in order to get the jdk-ea build to pass, it seems to be a good idea regardless, as, IIUC, this report is broken on master anyway. If you prefer this patch in a separate PR, just say so.

garydgregory:
Please try to get this to work with JaCoCo, we cannot loose this feature for our site.

mureinik:
> Please try to get this to work with JaCoCo, we cannot loose this feature for our site.

@garydgregory as I noted, as far as I can see, the JaCoCo report is broken regardless of this PR (the master build I linked to has a bunch of the same errors).
The project reports section on the site (https://commons.apache.org/proper/commons-io/project-reports.html) doesn't have a JaCoCo report (unlike, e.g., Commons Lang), unless it's located somewhere else and I'm missing something basic.

garydgregory:
The build you point to https://travis-ci.org/apache/commons-io/jobs/591248951 is a Java 14-EA build, I do not think we need to worry that JaCoCo does not support Java 14-EA at this point because it shadows a version of ASM that does not support Java 14-EA.

mureinik:
> The build you point to https://travis-ci.org/apache/commons-io/jobs/591248951 is a Java 14-EA build, I do not think we need to worry that JaCoCo does not support Java 14-EA at this point because it shadows a version of ASM that does not support Java 14-EA.

@garydgregory that's a good point, thanks. I originally overlooked that. I messed around with Travis CI a bit, and redid the last patch so JaCoCo is only disabled for the JDK EA build.

garydgregory:
Hi @mureinik,

Thank you for your update.
It seems like in `TailerTest` there should be a cleaner way to tear down tests than by calling `Thead.interrupt()`. Can `org.apache.commons.io.input.Tailer.stop()` be used with or without changes to `Tailer` to clean up tests in a nicer manner?

Gary

mureinik:
> Hi @mureinik,
> 
> Thank you for your update.
> It seems like in `TailerTest` there should be a cleaner way to tear down tests than by calling `Thead.interrupt()`. Can `org.apache.commons.io.input.Tailer.stop()` be used with or without changes to `Tailer` to clean up tests in a nicer manner?
> 
> Gary

@garydgregory The original JUnit 4 test tried doing that in the `tearDown()` method, which is wrong (since it's called after the temp dir deletion is attempted). Note this isn't new to JUnit Jupiter - it was too late in JUnit 4 too, but it was pretty harmless there (the deletion failed, and just silently left junk on the filesystem, whereas in Jupiter it fails the test).

By using a shared temp directory, we can delay the deletion till after the end of the entire class, and won't have to try something dirty like explicitly interrupting the thread, which would be arguably cleaner. I've updated the PR with this approach..

garydgregory:
> > Hi @mureinik,
> > Thank you for your update.
> > It seems like in `TailerTest` there should be a cleaner way to tear down tests than by calling `Thead.interrupt()`. Can `org.apache.commons.io.input.Tailer.stop()` be used with or without changes to `Tailer` to clean up tests in a nicer manner?
> > Gary
> 
> @garydgregory The original JUnit 4 test tried doing that in the `tearDown()` method, which is too late (since it's called after the temp dir deletion is attempted). Note this isn't new to JUnit Jupiter - it was too late in JUnit 4 too, but it was pretty harmless there (the deletion silently failed, and just left some junk on the temp filesystem, whereas in Jupiter it fails the test).
> 
> By using a shared temp directory, we can delay the deletion till after the end of the entire class, and won't have to try something dirty like explicitly interrupting the thread, which would be arguably cleaner. I've updated the PR with this approach..

@mureinik 
I still see calls to `thread.interrupt();`; did you forget to remove those?

mureinik:
> > > Hi @mureinik,
> > > Thank you for your update.
> > > It seems like in `TailerTest` there should be a cleaner way to tear down tests than by calling `Thead.interrupt()`. Can `org.apache.commons.io.input.Tailer.stop()` be used with or without changes to `Tailer` to clean up tests in a nicer manner?
> > > Gary
> > 
> > 
> > @garydgregory The original JUnit 4 test tried doing that in the `tearDown()` method, which is too late (since it's called after the temp dir deletion is attempted). Note this isn't new to JUnit Jupiter - it was too late in JUnit 4 too, but it was pretty harmless there (the deletion silently failed, and just left some junk on the temp filesystem, whereas in Jupiter it fails the test).
> > By using a shared temp directory, we can delay the deletion till after the end of the entire class, and won't have to try something dirty like explicitly interrupting the thread, which would be arguably cleaner. I've updated the PR with this approach..
> 
> @mureinik
> I still see calls to `thread.interrupt();`; did you forget to remove those?

@garydgregory The original JUnit 4 test has some calls to `thread.interupt()` - in `testTailer()` and `testInterrupt()` which I didn't touch (as they are part fo the test logic, not the migration to Jupiter). The diff (https://github.com/apache/commons-io/pull/97/files#diff-8f25c6c9a344f344c3266085f09adeb5) doesn't contain any new calls to `interrupt`, unless I'm missing something really obvious. Can you elaborate what calls you mean?

garydgregory:
> > > > Hi @mureinik,
> > > > Thank you for your update.
> > > > It seems like in `TailerTest` there should be a cleaner way to tear down tests than by calling `Thead.interrupt()`. Can `org.apache.commons.io.input.Tailer.stop()` be used with or without changes to `Tailer` to clean up tests in a nicer manner?
> > > > Gary
> > > 
> > > 
> > > @garydgregory The original JUnit 4 test tried doing that in the `tearDown()` method, which is too late (since it's called after the temp dir deletion is attempted). Note this isn't new to JUnit Jupiter - it was too late in JUnit 4 too, but it was pretty harmless there (the deletion silently failed, and just left some junk on the temp filesystem, whereas in Jupiter it fails the test).
> > > By using a shared temp directory, we can delay the deletion till after the end of the entire class, and won't have to try something dirty like explicitly interrupting the thread, which would be arguably cleaner. I've updated the PR with this approach..
> > 
> > 
> > @mureinik
> > I still see calls to `thread.interrupt();`; did you forget to remove those?
> 
> @garydgregory The original JUnit 4 test has some calls to `thread.interupt()` - in `testTailer()` and `testInterrupt()` which I didn't touch (as they are part fo the test logic, not the migration to Jupiter). The diff (https://github.com/apache/commons-io/pull/97/files#diff-8f25c6c9a344f344c3266085f09adeb5) doesn't contain any new calls to `interrupt`, unless I'm missing something really obvious. Can you elaborate which calls you're referring to?

@mureinik ,
Ah, I incorrectly assumed that your change would remove all calls to `thread.interupt()` - in `testTailer()`. If they were already there, it fair to leave them as is for now. I will complete the review tonight.

PR #98, opened 2019-10-10
IO-629: Providing more meaningful exceptions on file delete

malfist:
Applying patch found here: https://issues.apache.org/jira/browse/IO-629

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -34,6 +34,7 @@ import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


garydgregory:
@malfist 
Please rebase on master to allow the Travis build to proceed; this will bring in checkstyle fixes.
Gary

malfist:
@garydgregory I think I did that, not used to working with forks.

garydgregory:
@malfist You did the right thing. I made the fix in master after you created your PR.

garydgregory:
I think the better path here is to add a delete API to the PathUtils class.

malfist:
> I think the better path here is to add a delete API to the PathUtils class.

So are you thinking of moving these functions to PathUtils?
- cleanDirectory
- deleteDirectory
- deleteQuietly
- forceDelete
- forceDeleteOnExit 

For backwards compatibility, I'm assuming I should leave these functions in FileUtils, mark them as deprecated, and have them call the new PathUtils functions?

garydgregory:
No, not quite. Nothing would move in my view. FileUtils can keep on working on File instances only and PathUtils on Path instances. It might be possible for one to delegate to the other in some cases but that is not a given.

malfist:
I'm not following what you're asking for? Can you help me understand?

This function doesn't use Path objects except in the File.delete(Path) call. Java's Path class does not provide a mechanism to delete anything.

garydgregory:
I think what I am saying is that we should add `PathUtils.delete()` method(s) that work from a `Path` using the `Files` API and not muck with `FileUtils`.

garydgregory:
Please see git master for the latest.

malfist:
Rebased and used new PathUtils.delete function

garydgregory:
Something went wrong with your git work, master changes since your PR appear as new code here and it is now a pain to see what this PR truly provides. May you please clean up this PR?

malfist:
> 
> 
> Something went wrong with your git work, master changes since your PR appear as new code here and it is now a pain to see what this PR truly provides. May you please clean up this PR?

Got it cleaned up

PR #99, opened 2019-12-22
IO-458: Add a SequenceReader

RobberPhex:
<https://issues.apache.org/jira/browse/IO-458>

### DIFF
--- a/src/main/java/org/apache/commons/io/input/SequenceReader.java
+++ b/src/main/java/org/apache/commons/io/input/SequenceReader.java
@@ -1,0 +1,124 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.io.input;
+
+import java.io.IOException;
+import java.io.Reader;
+import java.util.List;
+
+import static org.apache.commons.io.IOUtils.EOF;
+
+/**
+ * Provides the contents of multiple Readers in sequence.
+ *
+ * @since 2.7
+ */
+public class SequenceReader extends Reader {
+    private Reader[] fReaders;
+    private int fReadersOffset;
+
+    /**
+     * Construct a new instance with readers
+     *
+     * @param readers the readers to read
+     */
+    public SequenceReader(List<? extends Reader> readers) {
+        final int size = readers.size();
+        fReaders = new Reader[size];
+        for (int i = 0; i < size; i++) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/commons/io/input/SequenceReader.java
```diff
@@ -0,0 +1,120 @@
...
+ * @since 2.7
+ */
+public class SequenceReader extends Reader {
+    private Reader[] fReaders;
+    private int fReadersOffset;
+
+    /**
+     * @param readers the readers from which to read
+     */
+    public SequenceReader(List<? extends Reader> readers) {
```
garydgregory: @RobberPhex 
If you make the input a `Collection`, then you can take a `Set` as input.

You should also consider using a List internally instead of an array. It might make the above simpler.

src/main/java/org/apache/commons/io/input/SequenceReader.java
```diff
@@ -0,0 +1,120 @@
...
+ * Provides the contents of multiple Readers in sequence.
+ *
+ * @since 2.7
+ */
+public class SequenceReader extends Reader {
+    private Reader[] fReaders;
+    private int fReadersOffset;
+
+    /**
+     * @param readers the readers from which to read
```
garydgregory: Incomplete Javadoc.

src/main/java/org/apache/commons/io/input/SequenceReader.java
```diff
@@ -0,0 +1,120 @@
...
+    public SequenceReader(List<? extends Reader> readers) {
+        int size = readers.size();
+        fReaders = new Reader[size];
+        for (int i = 0; i < size; i++) {
+            fReaders[i] = readers.get(i);
+        }
+    }
+
+    /**
+     * @param readers the readers from which to read
```
garydgregory: Incomplete Javadoc.

src/main/java/org/apache/commons/io/input/SequenceReader.java
```diff
@@ -0,0 +1,120 @@
...
+import static org.apache.commons.io.IOUtils.EOF;
+
+/**
+ * Provides the contents of multiple Readers in sequence.
+ *
+ * @since 2.7
+ */
+public class SequenceReader extends Reader {
+    private Reader[] fReaders;
+    private int fReadersOffset;
```
garydgregory: No need for the "f" prefix which would means what? "field"?

src/main/java/org/apache/commons/io/input/SequenceReader.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * @param readers the readers from which to read
+     */
+    public SequenceReader(List<? extends Reader> readers) {
+        int size = readers.size();
```
garydgregory: make `size` final.

src/test/java/org/apache/commons/io/input/SequenceReaderTest.java
```diff
@@ -0,0 +1,140 @@
...
+ */
+public class SequenceReaderTest {
+    private static final char NONE = (new char[1])[0];
+
+    @Test
+    public void testClose() throws IOException {
+        final Reader reader = new CharSequenceReader("FooBar");
+        checkRead(reader, "Foo");
+        reader.close();
+        checkRead(reader, "Foo");
```
garydgregory: What does this test?

src/test/java/org/apache/commons/io/input/SequenceReaderTest.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Test
+    public void testMarkSupported() throws Exception {
```
garydgregory: What does this test?

PR #103, opened 2020-03-01
IO-652 spring clean .gitignore

nhojpatrick:
Handles https://issues.apache.org/jira/browse/IO-652


sebbASF:
I think some of the files should not be ignored.
For example pom.* (apart from pom.xml) should not be present normally.

I wonder if OS-specific files should even be listed - I would expect them to be defined by the developer at system level.

PR #104, opened 2020-03-01
IO-656 update to junit v5.6.0

nhojpatrick:
Handles https://issues.apache.org/jira/browse/IO-656


garydgregory:
JUnit is at version 5.6.2 in master.

PR #106, opened 2020-03-02
IO-655 add hamcrest v2.2

nhojpatrick:
Handles https://issues.apache.org/jira/browse/IO-655


garydgregory:
Not sure this is relevant now that we are on JUnit 5 and nothing depends on Hamcrest directly.

sebbASF:
Won't do

PR #108, opened 2020-04-03
Refactor ByteArrayOutputStream into synchronized and non-synchronized versions

adamretter:
This abstracts the functionality of `ByteArrayOutputStream` into `AbstractByteArrayOutputStream`.  Synchronized and non-synchronized (for performance) implementations are then provided in `ByteArrayOutputStream` and `FastByteArrayOutputStream`.

Tests are also provided.

p.s. I have just submitted by ICLA by email.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -209,6 +209,10 @@ file comparators, endian transformation classes, and much more.
       <email>alban.peignier at free.fr</email>
     </contributor>
     <contributor>
+      <name>Adam Retter</name>
+      <organization>Evolved Binary</organization>
+    </contributor>
+    <contributor>
       <name>Ian Springer</name>
     </contributor>
     <contributor>
@@ -226,6 +230,11 @@ file comparators, endian transformation classes, and much more.
   </contributors>
 
   <dependencies>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+      <version>3.0.2</version>
+    </dependency>
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -231,11 +231,6 @@ file comparators, endian transformation classes, and much more.
 
   <dependencies>
     <dependency>
-      <groupId>com.google.code.findbugs</groupId>
-      <artifactId>jsr305</artifactId>
-      <version>3.0.2</version>
-    </dependency>
-    <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter</artifactId>
       <version>5.5.2</version>
--- a/src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java
+++ b/src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java
@@ -1,0 +1,391 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.io.output;
+
+import org.apache.commons.io.input.ClosedInputStream;
+
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.SequenceInputStream;
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import static org.apache.commons.io.IOUtils.EOF;
+
+/**
+ * This is the base class for implementing an output stream in which the data
+ * is written into a byte array. The buffer automatically grows as data
+ * is written to it.
+ * <p>
+ * The data can be retrieved using <code>toByteArray()</code> and
+ * <code>toString()</code>.
+ * <p>
+ * Closing an {@code AbstractByteArrayOutputStream} has no effect. The methods in
+ * this class can be called after the stream has been closed without
+ * generating an {@code IOException}.
+ * <p>
+ * This is the base for an alternative implementation of the
+ * {@link java.io.ByteArrayOutputStream} class. The original implementation
+ * only allocates 32 bytes at the beginning. As this class is designed for
+ * heavy duty it starts at 1024 bytes. In contrast to the original it doesn't
+ * reallocate the whole memory block but allocates additional buffers. This
+ * way no buffers need to be garbage collected and the contents don't have
+ * to be copied to the new buffer. This class is designed to behave exactly
+ * like the original. The only exception is the deprecated
+ * {@link java.io.ByteArrayOutputStream#toString(int)} method that has been
+ * ignored.
+ */
+public abstract class AbstractByteArrayOutputStream extends OutputStream {
+
+    static final int DEFAULT_SIZE = 1024;
+
+    /** A singleton empty byte array. */
+    private static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
+
+    /** The list of buffers, which grows and never reduces. */
+    private final List<byte[]> buffers = new ArrayList<>();
+    /** The index of the current buffer. */
+    private int currentBufferIndex;
+    /** The total count of bytes in all the filled buffers. */
+    private int filledBufferSum;
+    /** The current buffer. */
+    private byte[] currentBuffer;
+    /** The total count of bytes written. */
+    protected int count;
+    /** Flag to indicate if the buffers can be reused after reset */
+    private boolean reuseBuffers = true;
+
+    /**
+     * Makes a new buffer available either by allocating
+     * a new one or re-cycling an existing one.
+     *
+     * @param newcount  the size of the buffer if one is created
+     */
+    protected void needNewBuffer(final int newcount) {
+        if (currentBufferIndex < buffers.size() - 1) {
+            //Recycling old buffer
+            filledBufferSum += currentBuffer.length;
+
+            currentBufferIndex++;
+            currentBuffer = buffers.get(currentBufferIndex);
+        } else {
+            //Creating new buffer
+            int newBufferSize;
+            if (currentBuffer == null) {
+                newBufferSize = newcount;
+                filledBufferSum = 0;
+            } else {
+                newBufferSize = Math.max(
+                    currentBuffer.length << 1,
+                    newcount - filledBufferSum);
+                filledBufferSum += currentBuffer.length;
+            }
+
+            currentBufferIndex++;
+            currentBuffer = new byte[newBufferSize];
+            buffers.add(currentBuffer);
+        }
+    }
+
+    /**
+     * Write the bytes to byte array.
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -226,6 +230,11 @@ file comparators, endian transformation classes, and much more.
   </contributors>
 
   <dependencies>
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
```
aherbert, line 235: Is this just to bring in the @javax.annotation.concurrent.ThreadSafe annotation? I don't think this necessary since you already document the thread safety in the javadoc. Nothing else in the code base uses this non-standard annotation support.

src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTestCase.java
```diff
@@ -211,13 +217,30 @@ public void testStream() throws Exception {
...
         assertSame(baos1.toByteArray(), baos2.toByteArray());
         baos1.close();
         baos2.close();
         baout.close();
         baout1.close();
     }
+
+    private static Stream<Arguments> baosFactories() {
+        final BAOSFactory syncBaosFactory = size -> new ByteArrayOutputStream(size);
+        final BAOSFactory nonSyncBaos = size -> new FastByteArrayOutputStream(size);
```
aherbert, line 231: rename nonSyncBaosFactory

pom.xml
adamretter: Well yes it is to bring in the annotation... but these annotations can also be used by various tools like FindBugs and others to detect issues.

I think it adds value, and jsr305 is tiny (19KB). I can remove it if people prefer? It also adds useful annotations like `@Nullable` and `@NotNull` for similar purposes.

src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTestCase.java
adamretter: Done.

pom.xml
aherbert: I would remove it from this PR. Adding an annotation dependency should be in another PR that perhaps then makes more use of it than a single location. The Guava codebase uses these throughout and in that case they have value for a bug detection tool. I don't think a single annotation of `@ThreadSafe` is likely to detect many (any?) bugs given that synchronized code should be less of a problem than unsynchronized code. If you do have a bug detection tool that flags this class as a problem when used in a concurrent scenario a brief look at the javadoc (which notes this is safe for concurrent use) will allow you to mark the bug as false positive.

adamretter: Okie dokie. Done.

garydgregory:
> This looks like a nice conversion and maintains the previous support. I'd be wary of using a new dependency just to mark the code as Threadsafe with an annotation on an existing class. Bringing in an annotation framework is a different topic.
> 
> Using the name FastXXX is presumptuous but I cannot think of something better. NonSyncXXX is not as friendly. Others may have comments on this.

I am a big -1 on calling a new class "Fast" since it is not fast, it just removes thread-safety. You can hope that it is "faster" than the standard class but that's it.

The JRE has been calling some thread-safe classes `Concurrent` like `ConcurrentHashMap` but other JRE thread-safe classes just have names that reflect behavior like `CopyOnWriteArrayList`, it therefore seem OK to me to call a classes that is not synchronized one of:
- `UnsynchronizedByteArrayOutputStream` (to the point, precise)
- `SimpleByteArrayOutputStream` (not great)
- `BasicByteArrayOutputStream` (not great)
- `DefaultByteArrayOutputStream` (not great)

My preference is `UnsynchroniziedByteArrayOutputStream`.

adamretter:
@garydgregory Okay I understand. Your preferred option is very long though. How do you feel about: 

1. `UnsyncByteArrayOutputStream`.
2. `UByteArrayOutputStream`.
3. `USByteArrayOutputStream`.
4. `UnsafeByteArrayOutputStream`.

garydgregory:
> @garydgregory Okay I understand. Your preferred option is very long though. How do you feel about:
> 
> 1. `UnsyncByteArrayOutputStream`.
The least worst IMO.

> 2. `UByteArrayOutputStream`.
Oh, the horror! ;-) There is no convention that equates "U" to "Unsynchronized", if anything "U" usually stands for "Unsigned".

> 3. `USByteArrayOutputStream`.
Even worse IMO! Same comment as above except US usually means... United States?

> 4. `UnsafeByteArrayOutputStream`.
Nope, same as for "Fast", unsafe compared in what area? Will this be a CVE?

I do not thing there should be an issue spelling out Unsynchronized, if you look around, you'll see similar names spelled out:
- javax.transaction.Synchronization 
- java.util.concurrent.locks.AbstractOwnableSynchronizer
- java.util.concurrent.locks.AbstractQueuedLongSynchronizer
- java.util.concurrent.locks.AbstractQueuedSynchronizer
- java.util.Collections.SynchronizedCollection
- java.util.Collections.SynchronizedList
- java.util.Collections.SynchronizedMap
- java.util.Collections.SynchronizedNavigableMap
- java.util.Collections.SynchronizedNavigableSet
- java.util.Collections.SynchronizedRandomAccessList
- java.util.Collections.SynchronizedSet
- java.util.Collections.SynchronizedSortedMap<
- java.util.Collections.SynchronizedSortedSet
- Too many to list in Google Guava.
- Too many to list in our own Apache Commons Collections.

So, IMO, just say what it does.

src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java
```diff
@@ -0,0 +1,391 @@
...
+ * {@link java.io.ByteArrayOutputStream} class. The original implementation
+ * only allocates 32 bytes at the beginning. As this class is designed for
+ * heavy duty it starts at 1024 bytes. In contrast to the original it doesn't
+ * reallocate the whole memory block but allocates additional buffers. This
+ * way no buffers need to be garbage collected and the contents don't have
+ * to be copied to the new buffer. This class is designed to behave exactly
+ * like the original. The only exception is the deprecated
+ * {@link java.io.ByteArrayOutputStream#toString(int)} method that has been
+ * ignored.
+ */
```
garydgregory, line 56: Please Add `@since 2.7`.

src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java
```diff
@@ -0,0 +1,391 @@
...
+import static org.apache.commons.io.IOUtils.EOF;
+
+/**
+ * This is the base class for implementing an output stream in which the data
+ * is written into a byte array. The buffer automatically grows as data
+ * is written to it.
+ * <p>
+ * The data can be retrieved using <code>toByteArray()</code> and
+ * <code>toString()</code>.
+ * <p>
```
garydgregory, line 41: Please close HTML tags.

src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java
```diff
@@ -0,0 +1,391 @@
...
+            }
+
+            currentBufferIndex++;
+            currentBuffer = new byte[newBufferSize];
+            buffers.add(currentBuffer);
+        }
+    }
+
+    /**
+     * Write the bytes to byte array.
```
garydgregory, line 110: "Write" -> "Writes".

src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java
```diff
@@ -0,0 +1,391 @@
...
+                needNewBuffer(currentBuffer.length);
+                inBufferPos = 0;
+            }
+            n = in.read(currentBuffer, inBufferPos, currentBuffer.length - inBufferPos);
+        }
+        return readCount;
+    }
+
+    /**
+     * Return the current size of the byte array.
```
garydgregory, line 204: "Return" -> "Returns".

src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java
```diff
@@ -0,0 +1,391 @@
...
+        }
+        return newbuf;
+    }
+
+    /**
+     * Gets the current contents of this byte stream as a string
+     * using the platform default charset.
+     * @return the contents of the byte array as a String
+     * @see java.io.ByteArrayOutputStream#toString()
+     * @deprecated 2.5 use {@link #toString(String)} instead
```
garydgregory, line 356: I do not think it makes sense to deprecate `toString()` since it is such a basic JRE method that will never go away.

src/main/java/org/apache/commons/io/output/UnsynchronizedByteArrayOutputStream.java
```diff
@@ -0,0 +1,161 @@
...
+ * limitations under the License.
+ */
+package org.apache.commons.io.output;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+/**
+ * This class implements a version of {@link AbstractByteArrayOutputStream}
```
garydgregory, line 24: "This class implements..." -> "Implements..."

src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java
adamretter: This was how it was previously in `ByteArrayOutputStream`. I just left it as I found it!

adamretter:
@garydgregory I have incorporated your further feedback now.

adamretter:
@garydgregory Adding the closing `<p>` tags as you suggested breaks the javadoc build, so I will go ahead and remove them again:

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  06:02 min
[INFO] Finished at: 2020-04-05T12:41:14Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.1:javadoc (default-cli) on project commons-io: An error has occurred in Javadoc report generation: 
[ERROR] Exit code: 1 - /home/travis/build/adamretter/commons-io/src/main/java/org/apache/commons/io/output/ByteArrayOutputStream.java:115: error: unexpected end tag: </p>
[ERROR]      * </p>
[ERROR]        ^
[ERROR] /home/travis/build/adamretter/commons-io/src/main/java/org/apache/commons/io/output/ByteArrayOutputStream.java:142: error: unexpected end tag: </p>
[ERROR]      * </p>
[ERROR]        ^
[ERROR] /home/travis/build/adamretter/commons-io/src/main/java/org/apache/commons/io/output/UnsynchronizedByteArrayOutputStream.java:112: error: unexpected end tag: </p>
[ERROR]      * </p>
[ERROR]        ^
[ERROR] /home/travis/build/adamretter/commons-io/src/main/java/org/apache/commons/io/output/UnsynchronizedByteArrayOutputStream.java:138: error: unexpected end tag: </p>
[ERROR]      * </p>
[ERROR]        ^
[ERROR] 
[ERROR] Command line was: /usr/local/lib/jvm/openjdk11/bin/javadoc @options @packages
```

aherbert:
Looking at the line errors for your `</p>` markers they are all at lines following a `<ul>` element that cannot be nested within a paragraph. Move the `</p>` to before the `<ul>`.

adamretter:
@aherbert Good point. I have incorporated the changes as you suggested.

garydgregory:
Hi @adamretter 
Thank you for your updates.
JaCoCo reports that the code coverage for `UnsynchronizedByteArrayOutputStream` is less than it is for `ByteArrayOutputStream`.java. May you address this please?
Thank you!
You can run this: `mvn -V clean install site -P jacoco -P japicmp`

adamretter:
@garydgregory Hmm, from what I read the code coverage after this PR is `+0.07%` so it is a net improvement, no?

garydgregory:
> @garydgregory Hmm, from what I read in coveralls.io the code coverage after this PR is `+0.07%` so it is a net improvement, no?

Please run the command I provided and look at the JaCoCo report. You will see that the new class has less coverage than the current (refactored) one. We should strive to give the best possible code coverage for changes and new code.

adamretter:
@garydgregory Sure I see it now. I have just pushed a commit to further improve coverage.

adamretter:
@garydgregory @aherbert Yay! Thanks :-)

PR #110, opened 2020-04-08
Add functional interfaces IOFunction and IOSupplier

adamretter:
1. Adds `IOFunction` which it turns out I had already implemented in `ByteArrayOutputStreamTestCase` of #108 .
2. Adds `IOSupplier`.

These are designed to complement the existing `IOConsumer` class, and come with tests for 100% code coverage.

p.s. on a separate side-note... if you like these, I have a much larger library of Java Functional Interfaces that I wrote in the past that can work with exceptions here - https://github.com/adamretter/j8fu/tree/master/src/main/java/com/evolvedbinary/j8fu/function
Perhaps Apache commons would be interested in such a donation under a different scope?

### DIFF
--- a/src/main/java/org/apache/commons/io/function/IOFunction.java
+++ b/src/main/java/org/apache/commons/io/function/IOFunction.java
@@ -1,4 +1,170 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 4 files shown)


garydgregory:
@adamretter Non-IO functional interfaces likely belong in Apache Commons Lang. Please have a look there.

adamretter:
@garydgregory Okay cool. This PR does add two functional interfaces which are IO related. Are you happy with that?

PR #112, opened 2020-04-13
IO-665  ensure that passing a null InputStream results in NPE with tests

ottobackwards:
This pr adds a null check for the InputStream constructor of XmlStreamReader.  This makes it so that each contractor of type ( File, URL, URLConnection, InputStream ) throws NPE if passed a NULL for that type.

Tests were added as well for each type.

---
This change is [Reviewable](https://reviewable.io/reviews/apache/commons-io/112)

### DIFF
--- a/src/test/java/org/apache/commons/io/input/XmlStreamReaderTest.java
+++ b/src/test/java/org/apache/commons/io/input/XmlStreamReaderTest.java
@@ -84,7 +84,7 @@ public class XmlStreamReaderTest {
 
     @Test
     protected void testNullFileInput() throws IOException {
-        assertThrows(NullPointerException.class, ()-> new XmlStreamReader((File)null));
+        assertThrows(NullPointerException.class, () -> new XmlStreamReader((File)null));
     }
 
     @Test
@@ -94,12 +94,12 @@ public class XmlStreamReaderTest {
 
     @Test
     protected void testNullURLInput() throws IOException {
-         assertThrows(NullPointerException.class, ()-> new XmlStreamReader((URL)null));
+         assertThrows(NullPointerException.class, () -> new XmlStreamReader((URL)null));
     }
 
     @Test
     protected void testNullURLConnectionInput() throws IOException {
-          assertThrows(NullPointerException.class, ()-> new XmlStreamReader((URLConnection)null, "US-ASCII"));
+          assertThrows(NullPointerException.class, () -> new XmlStreamReader((URLConnection)null, "US-ASCII"));
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/io/input/XmlStreamReaderTest.java
```diff
@@ -78,6 +82,26 @@ protected void _testRawNoBomInvalid(final String encoding) throws Exception {
         }
     }
 
+    @Test
+    protected void testNullFileInput() throws IOException {
+        assertThrows(NullPointerException.class, ()-> new XmlStreamReader((File)null));
```
garydgregory, line 87: Hi @ottobackwards 
The spacing in "() ->" is inconsistent, it should be "() ->" everywhere.

ottobackwards:
The other contructors, already throw the NPE so I didn't change them. We could add the check to ensure that thread is always a message though.

ottobackwards:
Actually, since they use

` this(object.somethingThatReturnsAStream())`  we would have to change the initialization pattern from this() to some sort of initialize pattern.

Do you have a preference in that area?

They _behave_ correctly, but there is no message.

src/test/java/org/apache/commons/io/input/XmlStreamReaderTest.java
```diff
@@ -78,6 +82,26 @@ protected void _testRawNoBomInvalid(final String encoding) throws Exception {
         }
     }
 
+    @Test
+    protected void testNullFileInput() throws IOException {
+        assertThrows(NullPointerException.class, () -> new XmlStreamReader((File)null));
```
elharo, line 87: Has commons-io decided to move  Java 8 minimum? Maven still requires Java 7 and could not upgrade to this.

ottobackwards:
```xml
  <properties>
    <maven.compiler.source>1.8</maven.compiler.source>
    <maven.compiler.target>1.8</maven.compiler.target>
    <commons.componentid>io</commons.componentid>
    <commons.module.name>org.apache.commons.io</commons.module.name>
    <commons.rc.version>RC1</commons.rc.version>
    <commons.bc.version>2.6</commons.bc.version>
    <commons.release.version>2.7</commons.release.version>
    <commons.release.desc>(requires Java 8)</commons.release.desc>
```

garydgregory:
> Actually, since they use
> 
> ` this(object.somethingThatReturnsAStream())` we would have to change the initialization pattern from this() to some sort of initialize pattern.
> 
> Do you have a preference in that area?
> 
> They _behave_ correctly, but there is no message.

I do not see why you'd need anything more than changing methods like:
```
    public XmlStreamReader(final URL url) throws IOException {
        this(url.openConnection(), null);
    }
```

to:
```
    public XmlStreamReader(final URL url) throws IOException {
        this(Objects.requireNonNull(url, "url").openConnection(), null);
    }
```

ottobackwards:
That is because you @garydgregory know java better than I lol.  Will do

ottobackwards:
@garydgregory done

garydgregory:
> That is because you @garydgregory know java better than I lol. Will do

;-)

PR #114, opened 2020-05-24
Remove C-style array declaration

danygold:
Remove C-style array declaration.
This is the only point in all project that use this instead of classic Java-style array declaration.


danygold:
In #115 is fixed this "bug" and also others. 
Close this pull request because it's now redundant

PR #115, opened 2020-05-24
[IO-669] Fix code smells; fix typos

XenoAmess:


XenoAmess:
Also, in function NullPrintStreamTest.process(final NullPrintStream nos),
We use function void PrintStream.write(byte buf[]), which is @since 14
Should we really use a java 14 function in a java 8 project?
And is it neccessary?

garydgregory:
> Also, in function NullPrintStreamTest.process(final NullPrintStream nos),
> We use function void PrintStream.write(byte buf[]), which is @SInCE 14
> Should we really use a java 14 function in a java 8 project?
> And is it neccessary?

I just started a release candidate for 2.7 so this will have to wait until another RC if something goes wrong or more likely the next release.

The API you refer to is defined in Java 8 as `java.io.FilterOutputStream.write(byte[])`.

XenoAmess:
> > Also, in function NullPrintStreamTest.process(final NullPrintStream nos),
> > We use function void PrintStream.write(byte buf[]), which is @SInCE 14
> > Should we really use a java 14 function in a java 8 project?
> > And is it neccessary?
> 
> I just started a release candidate for 2.7 so this will have to wait until another RC if something goes wrong or more likely the next release.
> 
> The API you refer to is defined in Java 8 as `java.io.FilterOutputStream.write(byte[])`.

OK, none of them is bugfix nor other important things so you can review/merge it anytime you want sooner or later. 

I will find some time to test if j2se8 can run jdk14's compile output FilterOutputStream classfile.

src/main/java/org/apache/commons/io/file/Counters.java
```diff
@@ -92,8 +92,8 @@ public int hashCode() {
 
         @Override
         public String toString() {
-            return String.format("%,d files, %,d directories, %,d bytes", Long.valueOf(fileCounter.get()),
-                    Long.valueOf(directoryCounter.get()), Long.valueOf(byteCounter.get()));
+            return String.format("%,d files, %,d directories, %,d bytes", fileCounter.get(),
+                    directoryCounter.get(), byteCounter.get());
```
garydgregory, line 96: Same comment as in your other PR. Let's leave boxing and unboxing as obvious since we are in a low level library and I prefer to see where this happens so we can think about consequences of creating objects.

src/main/java/org/apache/commons/io/file/Counters.java
```diff
@@ -135,7 +135,7 @@ public BigInteger getBigInteger() {
 
         @Override
         public Long getLong() {
-            return Long.valueOf(value.longValueExact());
+            return value.longValueExact();
```
garydgregory, line 138: See other comment.

src/main/java/org/apache/commons/io/file/Counters.java
```diff
@@ -245,7 +245,7 @@ public BigInteger getBigInteger() {
 
         @Override
         public Long getLong() {
-            return Long.valueOf(value);
+            return value;
```
garydgregory, line 248: See other comment.

src/test/java/org/apache/commons/io/filefilter/FileFilterTestCase.java
```diff
@@ -1406,7 +1406,7 @@ public void testFilterListNullParameters() {
 
         final IOFileFilter filter = FileFilterUtils.trueFileFilter();
         try {
-            FileFilterUtils.filterList(filter, Arrays.asList((File) null));
+            FileFilterUtils.filterList(filter, Collections.singletonList((File) null));
```
garydgregory, line 1409: It is worth testing both kinds?

XenoAmess: @garydgregory it is only used as a Iterable there thus I think it has no actual difference IMO, exluding running speed.

src/main/java/org/apache/commons/io/file/Counters.java
XenoAmess: OK, will not happen next time :-)

src/main/java/org/apache/commons/io/file/Counters.java
XenoAmess: done.

src/main/java/org/apache/commons/io/file/Counters.java
XenoAmess: done.

PR #116, opened 2020-05-26
refine travis-ci scripts

XenoAmess:


sebbASF:
What is the purpose of the change?

XenoAmess:
@sebbASF 
> What is the purpose of the change?

to add cache for m2 folder, means dependencies jars, to avoid downloading them times and times.

PR #119, opened 2020-06-01
Add a MarkShieldInputStream

adamretter:
This is similar to the `CloseShieldInputStream` but it focuses on the mark, markSupported, and reset methods.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/MarkShieldInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/MarkShieldInputStream.java
@@ -31,7 +31,7 @@ import java.io.InputStream;
  * the stream, but it it is not desirable to allow marking of the stream.
  * </p>
  *
- * @since 2.8
+ * @since 2.8.0
  */
 public class MarkShieldInputStream extends ProxyInputStream {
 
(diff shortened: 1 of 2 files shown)


garydgregory:
Before we need one more and one more shield types of ProxyInputStreams, shouldn't we consider a more general shield class that takes an enum {CLOSE, MARK, THIS, THAT} so we can just have a single one?

adamretter:
@garydgregory Hmm I had not thought of that. Personally I prefer to have the separate classes... but I can refactor it if that's what is needed?

What would happen to the existing `CloseShieldInputStream`? Would it just be removed?

garydgregory:
We work hard to avoid breaking binary compatibility. Instead, we mark code as deprecated and that code can then be considered for removal for the next major version, which would include a package name change along with a matching Maven coordinate change. 
Let me poke around later...

garydgregory:
Yeah, this gets tricky and arbitrary: the class is called "MarkShield" which reimplements `mark()` as a noop but it also reimplements `reset()` to throw an `IOException`, which feels like a surprise to me (not in a good way.) Why should one method be a noop and the other throw an exception? It seems to be they both should be noops. WDYT?

adamretter:
@garydgregory If you look at java.io.InputStream source code - the default is to throw an exception when reset is called if mark is unsupported... so I am just following the defacto Java standard.

adamretter:
@garydgregory Could I get an update on this please?

garydgregory:
> @garydgregory If you look at `java.io.InputStream` source code - the default is to throw an exception when reset is called if mark is unsupported... so I am just following the defacto Java standard.

Yes, but `InputStream` is an abstract class, so there are no instances of it flying around. Why should this class blow up that call instead of passing it on? Or, does it make sense to have this class shield _both_ `mark()` _and_ `reset()`?

PR #122, opened 2020-07-08
Implemented CharSequenceReader.ready

robtimus:
`Reader.ready()` should return
> True if the next read() is guaranteed not to block for input, false otherwise

The default implementation is to return false, and that's what `CharSequenceReader` inherits. This change overrides `ready()` to return `true` if there characters can still be returned.


robtimus:
I've seen that the Travis build has failed, but that's a test unrelated to this change. I don't know why it fails though.

garydgregory:
I've restarted the build, that tailer test has failed 'randomly' in the past.

robtimus:
Thanks, it's green now as it should be.

PR #123, opened 2020-07-08
CharSequenceReader.skip should return 0 instead of EOF on stream end

robtimus:
`Reader.skip(long n)` should return
> The number of characters actually skipped

The current implementation of `CharSequenceReader` instead returns `EOF`, which is -1. This should be 0, which is just what this PR changes.


jochenw:
Approving. This introduces a minor incompatibility, but I guess, that the skip method is rarely used, and the spec rules.

PR #124, opened 2020-07-12
[IO-676] Add isFileNewer() and isFileOlder() methods that support the Java 8 Date/Time API.

Isira-Seneviratne:
Add `isFileNewer()` and `isFileOlder()` methods to `FileUtils` that accept the Java 8 Date/Time API classes, such as `Instant`, as parameters.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -38,10 +38,11 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.time.Instant;
-import java.time.LocalDate;
-import java.time.LocalDateTime;
+import java.time.LocalTime;
 import java.time.ZoneId;
-import java.time.ZonedDateTime;
+import java.time.chrono.ChronoLocalDate;
+import java.time.chrono.ChronoLocalDateTime;
+import java.time.chrono.ChronoZonedDateTime;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
@@ -89,9 +90,6 @@ import org.apache.commons.io.output.NullOutputStream;
  * </p>
  */
 public class FileUtils {
-    private static final String LOCAL_DATE = "localDate";
-    private static final String ZONE_ID = "zoneId";
-
     /**
      * The number of bytes in a kilobyte.
      */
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -1624,6 +1631,119 @@ public static boolean isFileNewer(final File file, final long timeMillis) {
...
+     * <code>ZonedDateTime</code>.
+     *
+     * @param file          the <code>File</code> of which the modification date
+     *                      must be compared, must not be {@code null}
+     * @param zonedDateTime the date reference, must not be {@code null}
+     * @return true if the <code>File</code> exists and has been modified
+     * after the given <code>ZonedDateTime</code>.
+     * @throws IllegalArgumentException if the file or zoned date time is {@code null}
+     */
+    public static boolean isFileNewer(final File file, final ZonedDateTime zonedDateTime) {
```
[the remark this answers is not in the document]
Isira-Seneviratne: The Javadoc for `ChronoZonedDateTime` includes this statement: "Most applications should declare method signatures, fields and variables as `ZonedDateTime`, not this interface."

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -1690,6 +1810,119 @@ public static boolean isFileOlder(final File file, final long timeMillis) {
...
+     *
+     * @param file    the <code>File</code> of which the modification date
+     *                must be compared, must not be {@code null}
+     * @param instant the date reference, must not be {@code null}
+     * @return true if the <code>File</code> exists and has been modified
+     * before the given <code>Instant</code>.
+     * @throws IllegalArgumentException if the file or instant is {@code null}
+     */
+    public static boolean isFileOlder(final File file, final Instant instant) {
+        if (instant == null) {
```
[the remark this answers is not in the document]
Isira-Seneviratne: Maybe the other methods could be updated to use `Objects.requireNonNull` as well in a separate PR?

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -1624,6 +1631,119 @@ public static boolean isFileNewer(final File file, final long timeMillis) {
...
+    public static boolean isFileNewer(final File file, final LocalDate localDate) {
+        if (localDate == null) {
+            throw new IllegalArgumentException(NO_SPECIFIED_LOCALDATE);
+        }
+        return isFileNewer(file, localDate.atStartOfDay());
+    }
+
+    /**
+     * Tests if the specified <code>File</code> is newer than the specified
+     * <code>LocalDate</code> at the specified <code>ZoneId</code>.
```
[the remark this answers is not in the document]
Isira-Seneviratne: Okay.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -1624,6 +1631,119 @@ public static boolean isFileNewer(final File file, final long timeMillis) {
...
+     *
+     * @param file          the <code>File</code> of which the modification date
+     *                      must be compared, must not be {@code null}
+     * @param localDateTime the date reference, must not be {@code null}
+     * @param zoneId        the time zone, must not be {@code null}
+     * @return true if the <code>File</code> exists and has been modified
+     * after the given <code>LocalDateTime</code> at the given <code>ZoneId</code>.
+     * @throws IllegalArgumentException if the file, local date time or zone ID is {@code null}
+     */
+    public static boolean isFileNewer(final File file, final LocalDateTime localDateTime, final ZoneId zoneId) {
```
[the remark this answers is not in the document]
Isira-Seneviratne: The Javadoc for `ChronoLocalDateTime` includes this statement: "Most applications should declare method signatures, fields and variables as `LocalDateTime`, not this interface."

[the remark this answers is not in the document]
garydgregory: What happens if I want to use a `JapaneseDate`, `HijrahDate`, `MinguoDate`, or `ThaiBuddhistDate`?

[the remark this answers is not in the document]
Isira-Seneviratne: Ah, I see what you mean. I'll update the method to use `ChronoLocalDateTime` :)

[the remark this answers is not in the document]
Isira-Seneviratne: Actually, those classes implement the `ChronoLocalDate` interface, so I'll use that instead.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -1603,93 +1598,272 @@ public static boolean isFileNewer(final File file, final File reference) {
...
+     *
+     * @param file                the {@code File} of which the modification date must be compared
+     * @param chronoZonedDateTime the date reference
+     * @return true if the {@code File} exists and has been modified after the given
+     * {@code ChronoZonedDateTime}.
+     * @throws NullPointerException if the file or zoned date time is {@code null}
+     *
+     * @since 2.8
+     */
+    public static boolean isFileNewer(final File file, final ChronoZonedDateTime<? extends ChronoLocalDate> chronoZonedDateTime) {
```
garydgregory, line 1643: You can replace `? extends ChronoLocalDate` with `?` since `ChronoZonedDateTime` is already defined with a generic that extends `ChronoLocalDate`. Same for other APIs I think.

src/test/java/org/apache/commons/io/FileUtilsTestCase.java
```diff
@@ -1015,6 +1021,12 @@ public void testIsFileNewerOlder() throws Exception {
 
         final Date date = new Date();
```
garydgregory, line 1022: In general, you want to use hard coded dates in order to avoid tests failing on the two times a year the build runs in a locale that happens to perform a daylight-saving-time switch. Yes, I've actually seen this happen (at work.)

Isira-Seneviratne: > 
> 
> In general, you want to use hard coded dates in tests

The files being tested seem to be created at the moment the test is run, so how should it be handled?

> Yes, I've actually seen this happen (at work.)

Wow.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -1603,93 +1598,272 @@ public static boolean isFileNewer(final File file, final File reference) {
...
+     * Tests if the specified {@code File} is newer than the specified {@code Instant}.
+     *
+     * @param file    the {@code File} of which the modification date must be compared
+     * @param instant the date reference
+     * @return true if the {@code File} exists and has been modified after the given {@code Instant}.
+     * @throws NullPointerException if the file or instant is {@code null}
+     *
+     * @since 2.8
+     */
+    public static boolean isFileNewer(final File file, final Instant instant) {
```
chtompki, line 1627: Curious here...what do you think happens if the TimeZone of the local datetime of the machine on which the code is running differs from the `TimeZone` embedded in the Instant. I worry that `instant.toEpochMilli();` might return the `Epoch` from the timezone of the instant which could be off by multiples of 30 minutes, and thus could yield differing results depending upon a user error here. Maybe we should convert the instant to the TimeZone of the machine before running `instant.toEpochMilli();`??

garydgregory:
@Isira-Seneviratne @chtompki 

Here is another thought which I think will give us more bang for the buck:

All of these `FileUtils` time comparison APIs should convert their input `File` object to a`Path` with `toPath()` and forward the calls to `PathUtils` which opens the feature to any backing NIO file system, not old school `File` objects.

Isira-Seneviratne:
> 
> 
> @Isira-Seneviratne @chtompki
> 
> Here is another thought which I think will give us more bang for the buck:
> 
> All of these `FileUtils` time comparison APIs should convert their input `File` object to a`Path` with `toPath()` and forward the calls to `PathUtils` which opens the feature to any backing NIO file system, not old school `File` objects.

That sounds like a good idea.

However, there's a potential issue in that those methods would become unusable on most versions of Android, since the `java.nio.file` package is only supported on Android Oreo and higher (admittedly, I'm not sure if any Android apps will need to make use of those methods, and it's possible that the package [will be made available](https://issuetracker.google.com/issues/154309654) to all versions of Android via [desugaring](https://developer.android.com/studio/write/java8-support-table)).

Other than that, sounds good to me :)

garydgregory:
> > @Isira-Seneviratne @chtompki
> > Here is another thought which I think will give us more bang for the buck:
> > All of these `FileUtils` time comparison APIs should convert their input `File` object to a`Path` with `toPath()` and forward the calls to `PathUtils` which opens the feature to any backing NIO file system, not old school `File` objects.
> 
> That sounds like a good idea.
> 
> However, there's a potential issue in that those methods would become unusable on most versions of Android, since the `java.nio.file` package is only supported on Android Oreo and higher (admittedly, I'm not sure if any Android apps will need to make use of those methods, and it's possible that the package [will be made available](https://issuetracker.google.com/issues/154309654) to all versions of Android via [desugaring](https://developer.android.com/studio/write/java8-support-table)).
> 
> Other than that, sounds good to me :)

Hi @Isira-Seneviratne 

After playing around a bit, I think I will proceed with this PR as is and leave a refactoring for later.

TBH, I had not considered Android. Would you have time to research if there is a way to build and test for Android on GitHub Actions and/or Travis-CI?

src/main/java/org/apache/commons/io/FileUtils.java
garydgregory: @Isira-Seneviratne Would you please reply to @chtompki ?

Isira-Seneviratne:
> 
> 
> > > @Isira-Seneviratne @chtompki
> > > Here is another thought which I think will give us more bang for the buck:
> > > All of these `FileUtils` time comparison APIs should convert their input `File` object to a`Path` with `toPath()` and forward the calls to `PathUtils` which opens the feature to any backing NIO file system, not old school `File` objects.
> > 
> > 
> > That sounds like a good idea.
> > However, there's a potential issue in that those methods would become unusable on most versions of Android, since the `java.nio.file` package is only supported on Android Oreo and higher (admittedly, I'm not sure if any Android apps will need to make use of those methods, and it's possible that the package [will be made available](https://issuetracker.google.com/issues/154309654) to all versions of Android via [desugaring](https://developer.android.com/studio/write/java8-support-table)).
> > Other than that, sounds good to me :)
> 
> Hi @Isira-Seneviratne
> 
> After playing around a bit, I think I will proceed with this PR as is and leave a refactoring for later.

Great!

> 
> TBH, I had not considered Android. Would you have time to research if there is a way to build and test for Android on GitHub Actions and/or Travis-CI?

Sure :)

On that note, I've made use of some of the methods in some apps I've contributed to, and they work great:

https://github.com/BOINC/boinc/tree/master/android/BOINC
https://invent.kde.org/network/kdeconnect-android

Isira-Seneviratne:
This article might be helpful: https://dev.to/ychescale9/running-android-emulators-on-ci-from-bitrise-io-to-github-actions-3j76

src/main/java/org/apache/commons/io/FileUtils.java
Isira-Seneviratne: @garydgregory My bad, I'll do so now.

@chtompki How should the conversion be done?

garydgregory:
@Isira-Seneviratne Would you please rebase on master, this should cause all CI builds to run.

src/main/java/org/apache/commons/io/FileUtils.java
uschindler: This is not true, an Instant never has a timezone. Unfortunately this method now goes on the forbiddenAPI list of Lucene because of this.
Instant is by definition not related to any time zone. toEpochMillies returns a UNIX timestamp.

uschindler: The big risk of this forward/backward transformation is now that at the DST changes hour, the calculation is wrong: During DST change the same local date time exists 2 times. When you convert the instant to a local time (after local time went backwards due to DST change) and then you convert the local time back, you're off by an hour.
The correct way here is to just call toEpochMillies(). PLEASE DO NOT CONVERT FORWARD BACKWARDS from to/local time.

uschindler: Because of this the whole methods added here are not useable in international application, except the one taking an explicit ZoneId. The one which would have been correct (the one taking an Instant) was corrupted by this forward/backward transformation.

uschindler (commented):
Please see the bugs introduced by this.

uschindler:
I opened https://issues.apache.org/jira/browse/IO-689 about the bug instroduced by https://github.com/apache/commons-io/pull/124/commits/2eb549873470844c88681e92c64631f796002020

PR #125, opened 2020-07-20
Simplify expression

bhawna94:
Simplify expressions to enhance code readability.

### DIFF
--- a/src/main/java/org/apache/commons/io/EndianUtils.java
+++ b/src/main/java/org/apache/commons/io/EndianUtils.java
@@ -54,8 +54,8 @@ public class EndianUtils {
      * @return the converted value
      */
     public static short swapShort(final short value) {
-        return (short) ( ( ( ( value >> 0 ) & 0xff ) << 8 ) +
-            ( ( ( value >> 8 ) & 0xff ) << 0 ) );
+        return (short) ( ( ( (value) & 0xff ) << 8 ) +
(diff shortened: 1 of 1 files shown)


namannigam:
Looks more like formatting changes. If not, can you revert all such formatting differences to specifically point out which expression is simplified.

sebbASF:
Whilst the operations with 0 don't actually do anything, they help to make the algorithm clearer, so I don't think the code should be changed. Or, if the code is changed, there needs to be documentation to note that the changed code is part of the same sequence.
In any case, won't the compiler optimise the operations away?

garydgregory:
-1 Same issue as @sebbASF mentions: the code follows a pattern for obvious clarity.

jochenw:
I would support a change, that splits the code into multiple statements. To give an idea of what I am thinking:

        byte byte0 = /* Whatever's necessary to calculate one byte */;
        byte byte1 = /* Calculate another byte */;
        byte byte2 = /* Calculate the third byte */
        byte byte3 = /* Calculate the lasr´t byte */
        long result = byte0 | (byte1 << 8) | (byte2 << 16) | (byte3 << 24);

In my opinion, that would be actually readable. The current code as well as this proposed change aren't.

garydgregory:
Unless someone update the PR, we should close it IMO.

garydgregory:
No feedback, closing.

PR #132, opened 2020-07-29
[IO-679] input.AbstractCharacterFilterReader passes count of chars read

proneel:
A bug in the read(final char[] cbuf, final int off, final int len)
method causes the filter function to receive the count of chars read
instead of iteratively getting each char code point

This affects subclasses CharacterFilterReader and
CharacterSetFilterReader also.

Added to tests to cover the bug


garydgregory:
@proneel Thank you!

PR #118, opened 2020-06-01
refine IOUtils.contentEquals(Reader, Reader)

XenoAmess:

PR #137, opened 2020-08-05
[IO-680] add more tests for IOUtils.contentEqualsIgnoreEOL

XenoAmess:


sebbASF:
What conditions are missed by the existing tests?
In any case, if new tests are needed, they should follow the existing syntax.
Or better, such tests should be independent.

XenoAmess:
@sebbASF 
> What conditions are missed by the existing tests?
> In any case, if new tests are needed, they should follow the existing syntax.
> Or better, such tests should be independent.

In short, it lacks situations I added.
In details, it lacks:

1. "" and "\n"
because "" is very tricky, and it is hard to say it for sure whether we think them equal.

2. "a\n" and "a\n\n"
This might sounds stupid, but actually very tricky. because. this function think "a" equals "a\n", but it thinks "a\n" not equals "a\n\n".
Only a non-CRLF char can be follow a ending CRLF and cause no difference to original, in this function.

3. some Readers who same to itself.
The existing tests only have some examples for a same Object, who will be killed exit at the first == check.
But it lacks some real same input test.

4. for every tests, it lacks verse.
whenever check `contentEqualsIgnoreEOF(input1,input2)` be true, must make sure it `contentEqualsIgnoreEOF(input2,input1)` also be true.

5. some \r, \n, \r\n examples.
the original only have one \r\n example, but have no tests for \r and \n.

6. non-equal-length examples.
like "123" and "1234" be false.
they share common prefix, but a longer one is not same.

7. some really not equal examples.
like "1235" and "1234", they really differ on some char, and should return false.

garydgregory:
When I run `mvn site -P jacoco` on git. master, this method already has 100% code coverage, so more tests are not needed IMO.

XenoAmess:
@garydgregory 
> When I run `mvn site -P jacoco` on git. master, this method already has 100% code coverage, so more tests are not needed IMO.

Hi.
They are for pr https://github.com/apache/commons-io/pull/118
Sebb think I should not change test codes in that pr (in order not to break any test, which I agree somehow), so I have to split them out to a new pr.
So this pr SHOULD be reviewed only after https://github.com/apache/commons-io/pull/118

garydgregory:
JaCoCo shows this method's coverage is already at 100%/100%, so this does not test anything new. Especially since all EOL handling is handled by the `Reader`'s `readLine()` method, not by our method, so I say let's keep it simple and not test the JRE itself. This PR can just be closed IMO.

XenoAmess:
> JaCoCo shows this method's coverage is already at 100%/100%, so this does not test anything new. Especially since all EOL handling is handled by the `Reader`'s `readLine()` method, not by our method, so I say let's keep it simple and not test the JRE itself. This PR can just be closed IMO.

coverages can lie, sir.

garydgregory:
Hi @XenoAmess 
Would you rebase on master? I can bring this in then.

XenoAmess:
> Hi @XenoAmess Would you rebase on master? I can bring this in then.

@garydgregory done.

PR #139, opened 2020-08-08
[IO-682] Add lines() methods for InputStream and Reader.

Isira-Seneviratne:
Add `lines()` methods that return a `Stream` object for `InputStream` and `Reader`.


sebbASF:
Thanks, but the extra code is not worth the minimal reduction in user code (see the JIRA issue)

PR #150, opened 2020-09-01
Add isFileNewer() and isFileOlder() methods that accept OffsetTime and OffsetDateTime.

Isira-Seneviratne:


garydgregory:
Close: Merged manually to git master. TY @Isira-Seneviratne.

PR #163, opened 2020-10-08
Tiny performance improvements

michielgkalkman:
I found that copyDirectory(final File srcDir, final File destDir, final FileFilter filter, final boolean preserveFileDate) executes getCanonicalPath() multiple times on the same path. I might be wrong, but that seems wasteful. So, I created this pull request.

I created a unit test based on jmh and it seemed to indicate that the performance indeed improves. However, I am not an expert on benchmarking and profiling and I cannot explain all results. Don't hesitate to show where I went wrong and/or might improve.

See https://github.com/michielgkalkman/PmdPlus/tree/master/src/test/java/org/taHjaj/wo/pmdplus/dup/jmh/org/apache/commons/io (or: https://github.com/michielgkalkman/PmdPlus/commit/10d44959c4441480fa6d1051a9c0fd65bfba4eb2) for some background on what I did test..


garydgregory:
Looks reasonable even without the performance minor issue, for me, the code is less cluttered with this refactoring. TY! :-)

PR #171, opened 2020-11-26
Add QueueInput/OutputStream as simpler alternatives to PipedInput/OutputStream

maxxedev:
Add QueueInput/OutputStream as simpler alternatives to PipedInput/OutputStream

PipedInput/OutputStream in JDK can have surprisingly complex behavior with respect
to how threads need to be arranged. QueueInput/OutputStream are much simpler
alternatives that is easier to use correctly.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/QueueInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/QueueInputStream.java
@@ -21,10 +21,9 @@ import org.apache.commons.io.output.QueueOutputStream;
 import java.io.InputStream;
 import java.io.PipedInputStream;
 import java.io.PipedOutputStream;
+import java.util.Objects;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
-
-import static java.util.Objects.requireNonNull;
 
 /**
  * Simple alternative to JDK {@link java.io.PipedInputStream}; queue input stream provides what's written in queue output stream.
@@ -65,7 +64,7 @@ public class QueueInputStream extends InputStream {
      * @param queue backing queue for the stream
      */
     public QueueInputStream(final BlockingQueue<Integer> queue) {
-        this.queue = requireNonNull(queue, "queue is required");
+        this.queue = Objects.requireNonNull(queue, "queue is required");
     }
 
     /**
--- a/src/main/java/org/apache/commons/io/input/QueueInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/QueueInputStream.java
@@ -52,14 +52,14 @@ public class QueueInputStream extends InputStream {
     private final BlockingQueue<Integer> queue;
 
     /**
-     * Create a {@link QueueInputStream} with no limit to internal buffer size
+     * Constructs a QueueInputStream with no limit to internal buffer size
      */
     public QueueInputStream() {
         this(new LinkedBlockingQueue<>());
     }
 
     /**
-     * Create a {@link QueueInputStream} with given buffer
+     * Constructs a QueueInputStream with given buffer
      * 
      * @param queue backing queue for the stream
      */
@@ -68,7 +68,7 @@ public class QueueInputStream extends InputStream {
     }
 
     /**
-     * Create an QueueOutputStream connected to this. Writes to the output stream will be visible to this input stream.
+     * Creates a new QueueOutputStream instance connected to this. Writes to the output stream will be visible to this input stream.
      * 
      * @return QueueOutputStream connected to this stream
      */
@@ -77,9 +77,9 @@ public class QueueInputStream extends InputStream {
     }
 
     /**
-     * Read a single byte.
+     * Reads a single byte.
      *
-     * @return either the byte read or <code>-1</code> if the end of the stream has been reached
+     * @return either the byte read or {@code -1} if the end of the stream has been reached
      */
     @Override
     public int read() {
--- a/src/main/java/org/apache/commons/io/input/QueueInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/QueueInputStream.java
@@ -1,0 +1,87 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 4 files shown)


garydgregory:
IMO, you need better tests with edge cases like input size of 0, 1, 2, 4, 8, and the "usual" buffer sizes from other streams like 4096 or the more recent 8192 (private java.io.BufferedInputStream.DEFAULT_BUFFER_SIZE).
You can parameterize the tests with normal JUnit features. Someone (like me ;-) will invariably look at the current implementation and not like the one byte at a time implementations since normal streams usually work on buffers, but that's a different story, let's start with better tests.

src/main/java/org/apache/commons/io/output/QueueOutputStream.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.commons.io.input.QueueInputStream;
+
+import java.io.InterruptedIOException;
+import java.io.OutputStream;
+import java.io.PipedInputStream;
+import java.io.PipedOutputStream;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+
+import static java.util.Objects.requireNonNull;
```
garydgregory, line 28: We don't use static imports for this kind of call in the rest of the main code base. In unit tests, we use statics for JUnit methods.

src/main/java/org/apache/commons/io/output/QueueOutputStream.java
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * @throws InterruptedIOException if the thread is interrupted while writing to the queue.
+     */
+    @Override
+    public void write(final int b) throws InterruptedIOException {
+        try {
+            queue.put(0xFF & b);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new InterruptedIOException();
```
garydgregory, line 92: Is there a reason, the catch block creates a new exception instead of rethrowing the one it caught?

maxxedev: InterruptedException does not extend IOException but InterruptedIOException does.

garydgregory: Then we shouldn't we still preserve the original exception by calling `initCause()` on the new exception?

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -0,0 +1,90 @@
...
+     * @return QueueOutputStream connected to this stream
+     */
+    public QueueOutputStream newQueueOutputStream() {
+        return new QueueOutputStream(queue);
+    }
+
+    /**
+     * Read a single byte.
+     *
+     * @return either the byte read or <code>-1</code> if the end of the stream has been reached
```
garydgregory, line 82: You can abbreviate ` <code>-1</code>` as `{@code -1}`.
You could also avoid using a magic number and do a static import of `IOUtils.EOF` in the code.

garydgregory (changes requested):

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -0,0 +1,90 @@
...
+ * 
+ * @see QueueOutputStream
+ * @since 2.9.0
+ */
+public class QueueInputStream extends InputStream {
+
+    private final BlockingQueue<Integer> queue;
+
+    /**
+     * Create a {@link QueueInputStream} with no limit to internal buffer size
```
garydgregory, line 55: `Create` -> `Constructs` since a constructor constructs ;-)

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -0,0 +1,90 @@
...
+     * Create a {@link QueueInputStream} with given buffer
+     * 
+     * @param queue backing queue for the stream
+     */
+    public QueueInputStream(final BlockingQueue<Integer> queue) {
+        this.queue = Objects.requireNonNull(queue, "queue is required");
+    }
+
+    /**
+     * Create an QueueOutputStream connected to this. Writes to the output stream will be visible to this input stream.
```
garydgregory, line 71: `an QueueOutputStream` -> `an instance`.

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -0,0 +1,90 @@
...
+     * 
+     * @return QueueOutputStream connected to this stream
+     */
+    public QueueOutputStream newQueueOutputStream() {
+        return new QueueOutputStream(queue);
+    }
+
+    /**
+     * Read a single byte.
```
garydgregory, line 80: `Read` -> `Reads`.

src/main/java/org/apache/commons/io/output/QueueOutputStream.java
```diff
@@ -0,0 +1,96 @@
...
+ * 
+ * @see QueueInputStream
+ * @since 2.9.0
+ */
+public class QueueOutputStream extends OutputStream {
+
+    private final BlockingQueue<Integer> queue;
+
+    /**
+     * Create a {@link QueueOutputStream} with no limit to internal buffer size
```
garydgregory, line 57: -> `Creates a new instance...` 

I am not a fan of creating a hyperlink to the class we are already in, it feels misleading, like we are sending someone elsewhere, when we are not.

src/main/java/org/apache/commons/io/input/QueueInputStream.java
maxxedev: -1 as magic number comes from [InputStream::read](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#read--). I think it'd better to keep it that way to be consistent with super method documentation.

src/main/java/org/apache/commons/io/input/QueueInputStream.java
maxxedev: This Queue**Input**Stream method creates a Queue**Output**Stream instance. Just saying "an instance" might be confusing?

garydgregory: Ah, yes of course!

garydgregory:
I will review all of this tomorrow (Thursday).

PR #174, opened 2020-12-15
Tiny performance improvement in FileUtils#moveDirectoryToDirectory()

michielgkalkman:
I think this version performs better, especially when the directory to move to already exists (2 calls to exists are skipped).


garydgregory:
Hi @michielgkalkman 
Don't you mean ONE call to `exists()` skipped instead of 2?

michielgkalkman:
No, in the original code, when the destination directory already exists, you do an exists() on this directory in lines 2173 and 2178 before isDirectory() is executed.

However, isDirectory() can be executed right away (it does an exists() itself) and thus you skip two calls to exists().

PR #178, opened 2020-12-28
IO-600: fix getPrefixLength for Linux filename

abchaubey:


pom.xml
```diff
@@ -269,7 +269,6 @@ file comparators, endian transformation classes, and much more.
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
       <version>3.11</version>
-      <scope>test</scope>
```
garydgregory, line 272: -1 we don't want this runtime dependency.

src/main/java/org/apache/commons/io/FilenameUtils.java
```diff
@@ -665,7 +666,10 @@ public static int getPrefixLength(final String fileName) {
         if (ch1 == ':') {
             ch0 = Character.toUpperCase(ch0);
             if (ch0 >= 'A' && ch0 <= 'Z') {
-                if (len == 2 || isSeparator(fileName.charAt(2)) == false) {
+                if (len == 2 && SystemUtils.IS_OS_LINUX) {
```
IchbinkeinReh, line 669: I think you can use "!FilenameUtils.isSystemWindows()" instead. see https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/FilenameUtils.java#L146

abchaubey: thanks, updated to use  "!FilenameUtils.isSystemWindows()"

pom.xml
abchaubey: thanks, remove runtime dependency

src/main/java/org/apache/commons/io/FilenameUtils.java
garydgregory: I though about this a little more and since other OSs support driver letters, I've abstracted this away in a new method `FileSystem#supportsDriveLetter()`, so please rebase on master and use that method instead of testing for the OS directly.

abchaubey: FileSystem.supportsDriveLetter() is non-static method getPrefixLength is static

garydgregory: So?

garydgregory:
@abchaubey 
Please rebase on master.
TY.

abchaubey:
@garydgregory  I will create new PR for this from new fork

abchaubey:
@garydgregory new PR :https://github.com/apache/commons-io/pull/179 , closing this one

PR #179, opened 2020-12-30
[IO-600] Fix getPrefixLength method for Linux filename

abchaubey:


src/main/java/org/apache/commons/io/FilenameUtils.java
```diff
@@ -665,7 +665,10 @@ public static int getPrefixLength(final String fileName) {
         if (ch1 == ':') {
             ch0 = Character.toUpperCase(ch0);
             if (ch0 >= 'A' && ch0 <= 'Z') {
-                if (len == 2 || isSeparator(fileName.charAt(2)) == false) {
+                if (len == 2 && !FileSystem.getCurrent().supportsDriveLetter()) {
```
garydgregory, line 668: The last step is to document this in the Javadoc for the method.

abchaubey: adding this in under Unix section : 
 * C:                  --&gt; ""          --&gt; valid filename as only null byte and / are reserved characters
 Let me know your thoughts, thanks

garydgregory: Well, our current Javadoc is lame because while this methods returns an `int`, the Javadoc shows `String` examples as 'output' which is not so helpful IMO, so if you can fix the Javadoc to show what the `int` value is now for all the example strings, that would be appreciated.

abchaubey: Updated Javadoc

PR #181, opened 2021-01-04
IO-696: Fix IOUtils.toByteArray exception

abchaubey:


garydgregory:
Hi @abchaubey 
Thank you for your PR.
I think we need to consider taking a look at all of Commons IO and consider if this is a task we want to do consistently: When do we throw an IOException vs IllegalArgumentException vs IllegalStateException vs UnsupportedOperationException and so on. 
So IOW, let's pause and all think about this instead of doing a one-off PR here...
This case in particular is interesting because the exception is thrown because the expected input does not match the actual file, so either the input is wrong (IAE) OR something went wrong while reading the file (IOEx). So either might be OK here. One general rule could be that IOEx all come from the JRE... needs further thought...

abchaubey:
@garydgregory thanks for comments,  let me know if any further analysis needs to be done to identify all such scenarios in commons-io

abchaubey:
Deleting this PR as changes are already taken care of

PR #187, opened 2021-01-15
IO-702 FileUtils.forceDelete does not delete invalid links. Testcase …

boris-unckel:
…and fix of check.
Testcase and first fix for [https://issues.apache.org/jira/browse/IO-702](https://issues.apache.org/jira/browse/IO-702)

### DIFF
--- a/src/test/java/org/apache/commons/io/FileUtilsDeleteDirectoryBaseTestCase.java
+++ b/src/test/java/org/apache/commons/io/FileUtilsDeleteDirectoryBaseTestCase.java
@@ -195,6 +195,29 @@ public abstract class FileUtilsDeleteDirectoryBaseTestCase {
     }
 
     @Test
+    public void testDeleteInvalidLinks() throws Exception {
+        File aFile = new File(top, "realParentDirA");
+        assertTrue(aFile.mkdir());
+        File bFile = new File(aFile, "realChildDirB");
+        assertTrue(bFile.mkdir());
+
+        File cFile = new File(top, "realParentDirC");
+        assertTrue(cFile.mkdir());
+        File dFile = new File(cFile, "realChildDirD");
+        assertTrue(dFile.mkdir());
+
+        File linkToC = new File(bFile, "linkToC");
+        Files.createSymbolicLink(linkToC.toPath(), cFile.toPath());
+
+        File linkToB = new File(dFile, "linkToB");
+        Files.createSymbolicLink(linkToB.toPath(), bFile.toPath());
+
+        FileUtils.deleteDirectory(aFile);
+        FileUtils.deleteDirectory(cFile);
+        assertEquals(0, top.list().length);
+    }
+
+    @Test
     public void testDeletesNested() throws Exception {
         final File nested = new File(top, "nested");
         assertTrue(nested.mkdirs());
--- a/src/main/java/org/apache/commons/io/file/DeletingPathVisitor.java
+++ b/src/main/java/org/apache/commons/io/file/DeletingPathVisitor.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.LinkOption;
+import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Arrays;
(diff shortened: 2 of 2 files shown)


boris-unckel:
@garydgregory
The deleteDirectory story is still not final, but with different root cause. Please review the testcase and first fix. Do you have an idea whats wrong with forceDelete?

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -309,6 +309,9 @@ public static void cleanDirectory(final File directory) throws IOException {
         if (!causeList.isEmpty()) {
             throw new IOExceptionList(causeList);
         }
+        if (listFiles(directory, null).length > 0) {
```
garydgregory, line 312: Hi @boris-unckel 

Thank you for your PR.

That is an expensive way to test for emptiness if there are a lot of files. Please rebase on master and call `isEmptyDirectory(File)` in this class.

TY.

src/test/java/org/apache/commons/io/FileUtilsDeleteDirectoryBaseTestCase.java
```diff
@@ -194,6 +194,29 @@ public void testDeleteParentSymlink2() throws Exception {
         assertEquals(1, randomDirectory.list().length, "Contents of sym link should not have been removed");
     }
 
+    @Test
+    public void testDeleteInvalidLinks() throws Exception {
+        File AFile = new File(top, "realParentDirA");
```
garydgregory, line 199: These local variable names are unconventional, `AFile` -> `fileA` as local variable names should start with a lower case chaarcter.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -1342,6 +1345,9 @@ public static void forceDelete(final File file) throws IOException {
             // didn't find a file to delete.
             throw new FileNotFoundException("File does not exist: " + file);
         }
+        if (file.exists()) {
```
garydgregory, line 1348: This belongs in the unit test.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -309,6 +309,9 @@ public static void cleanDirectory(final File directory) throws IOException {
         if (!causeList.isEmpty()) {
             throw new IOExceptionList(causeList);
         }
+        if (listFiles(directory, null).length > 0) {
+            throw new IOException("Directory " + directory + " is not empty, cleaning attempt failed.");
```
garydgregory, line 313: This belongs in the unit test IMO.

src/test/java/org/apache/commons/io/FileUtilsDeleteDirectoryBaseTestCase.java
boris-unckel: thnx, corrected.

src/main/java/org/apache/commons/io/FileUtils.java
boris-unckel: removed

src/main/java/org/apache/commons/io/FileUtils.java
boris-unckel: removed

src/main/java/org/apache/commons/io/FileUtils.java
boris-unckel: obsolete - it's removed, see later comment

boris-unckel:
@garydgregory 
Now it fails again in deleteDirectory (which is not the cause)
`Error:  org.apache.commons.io.FileUtilsDeleteDirectoryLinuxTestCase.testDeleteInvalidLinks  Time elapsed: 0.003 s  <<< ERROR!
java.nio.file.DirectoryNotEmptyException: /tmp/junit16834268871756684963/realParentDirC
	at java.base/sun.nio.fs.UnixFileSystemProvider.implDelete(UnixFileSystemProvider.java:247)
	at java.base/sun.nio.fs.AbstractFileSystemProvider.delete(AbstractFileSystemProvider.java:105)
	at java.base/java.nio.file.Files.delete(Files.java:1141)
	at org.apache.commons.io.FileUtils.delete(FileUtils.java:1170)
	at org.apache.commons.io.FileUtils.deleteDirectory(FileUtils.java:1189)
	at org.apache.commons.io.FileUtilsDeleteDirectoryBaseTestCase.testDeleteInvalidLinks(FileUtilsDeleteDirectoryBaseTestCase.java:216)`

IMO This stacktrace is misleading because originally forceDelete fails to delete the broken link. Maybe it's the Visitor in the nio TreeWalker.

boris-unckel:
@garydgregory
Fixed the DeletingPathVisitor.visitFile with an additional check for symlinks. Please review again.

PR #193, opened 2021-01-27
[IO-708] Refactor o.a.c.io.FileUtils to NIO

boris-unckel:
Fixes https://issues.apache.org/jira/browse/IO-708

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -27,6 +27,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.Reader;
+import java.io.UncheckedIOException;
 import java.math.BigInteger;
 import java.net.URL;
 import java.net.URLConnection;
@@ -1229,7 +1230,7 @@ public class FileUtils {
             return false;
         }
         try {
-            if (file.isDirectory()) {
+            if (FileUtils.isDirectory(file)) {
                 cleanDirectory(file);
             }
         } catch (final Exception ignored) {
@@ -1237,7 +1238,8 @@ public class FileUtils {
         }
 
         try {
-            return file.delete();
+            FileUtils.delete(file);
+            return true;
         } catch (final Exception ignored) {
             return false;
         }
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -1838,6 +1841,36 @@ public static boolean isSymlink(final File file) {
...
+     * <p>
+     * This method delegates to {@link Files#isRegularFile(Path, LinkOption...)}
+     * </p>
+     *
+     * @param file the file to test.
+     * @return true if the file is a regular file, see {@link Files#isRegularFile(Path, LinkOption...)}.
+     * @since 2.9
+     * @see Files#isRegularFile(Path, LinkOption...)
+     */
+    public static boolean isRegularFile(final File file) {
```
garydgregory, line 1855: If you want to do this kind of rework, add "primitive" methods to PathUtils and then have this method just pass the call forward with the File#toPath() result. The PathUtils method then just becomes a null-safe version of the underlying JRE Files API.

garydgregory (changes requested):
Hi @boris-unckel 

Thank you for your PR.

This PR might need to be broken up in smaller PRs. See my other comment for the new API(s) for example. 

Please be aware that there are subtle differences between IO and NIO which has caused a few bug reports after we released 2.8.0 such that this kind of rework needs to be done quite carefully and therefore in small increments. You can scan changes.xml and Jira for these problems, for example in dealing with symbolic links among other issues.

src/main/java/org/apache/commons/io/FileUtils.java
garydgregory, line 1855: IMO you could have a separate PR that only add "is" methods that cover the Files.is() APIs.

boris-unckel:
Smaller refactorings startet here [https://github.com/apache/commons-io/commit/4caff5d7b7f114d3ca300abb76b6e275167794c6](https://github.com/apache/commons-io/commit/4caff5d7b7f114d3ca300abb76b6e275167794c6) and here [https://github.com/apache/commons-io/pull/196](https://github.com/apache/commons-io/pull/196)

PR #196, opened 2021-01-28
[IO-711] Use Objects.requireNotNull for fail fast method/constructors

boris-unckel:
Fixes https://issues.apache.org/jira/browse/IO-711

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -867,6 +867,8 @@ public class FileUtils {
      * @since 2.1
      */
     public static long copyFile(final File input, final OutputStream output) throws IOException {
+        Objects.requireNonNull(input, "input");
+        Objects.requireNonNull(output, "output");
         try (FileInputStream fis = new FileInputStream(input)) {
             return IOUtils.copyLarge(fis, output);
         }
@@ -1038,6 +1040,8 @@ public class FileUtils {
      * @since 2.5
      */
     public static void copyToFile(final InputStream inputStream, final File file) throws IOException {
+        Objects.requireNonNull(inputStream, "inputStream");
+        Objects.requireNonNull(file, "file");
         try (OutputStream out = openOutputStream(file)) {
             IOUtils.copy(inputStream, out);
         }
@@ -1064,6 +1068,8 @@ public class FileUtils {
      * @throws IOException if an IO error occurs during copying
      */
     public static void copyURLToFile(final URL source, final File destination) throws IOException {
+        Objects.requireNonNull(source, "source");
+        Objects.requireNonNull(destination, "destination");
         try (final InputStream stream = source.openStream()) {
             copyInputStreamToFile(stream, destination);
         }
@@ -1090,6 +1096,8 @@ public class FileUtils {
      */
     public static void copyURLToFile(final URL source, final File destination,
         final int connectionTimeoutMillis, final int readTimeoutMillis) throws IOException {
+        Objects.requireNonNull(source, "source");
+        Objects.requireNonNull(destination, "destination");
         final URLConnection connection = source.openConnection();
         connection.setConnectTimeout(connectionTimeoutMillis);
         connection.setReadTimeout(readTimeoutMillis);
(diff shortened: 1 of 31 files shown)


boris-unckel:
@garydgregory Please review and mark the cases you don't want to change for compatibility and/or design reasons. I'll change then all occurrences accordingly.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -866,6 +868,8 @@ public static void copyFile(final File srcFile, final File destFile, final CopyO
      * @since 2.1
      */
     public static long copyFile(final File input, final OutputStream output) throws IOException {
+        Objects.requireNonNull(input, "input");
+        Objects.requireNonNull(output, "output");
```
garydgregory, line 872: -1: no need to check on `output`, this already done in `copyLarge`.

I'll stop this type of comment since these checks over the top in this PR. IOW, let's not check for NPEs when we do not need to.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -1200,6 +1210,7 @@ public static void deleteDirectory(final File directory) throws IOException {
      * @throws IOException          in case deletion is unsuccessful
      */
     private static void deleteDirectoryOnExit(final File directory) throws IOException {
+        Objects.requireNonNull(directory,"directory");
```
garydgregory, line 1213: Fix formatting.

src/main/java/org/apache/commons/io/FilenameUtils.java
```diff
@@ -1510,6 +1510,9 @@ public static boolean wildcardMatch(final String fileName, final String wildcard
      * @return true if the given name is a valid host name
      */
     private static boolean isValidHostName(final String name) {
+        if(Objects.isNull(name)) {
```
garydgregory, line 1513: Fix formatting and no need to call an API meant for use in lambdas when `name == null` would do.

src/main/java/org/apache/commons/io/monitor/FileAlterationObserver.java
```diff
@@ -200,12 +201,9 @@ public FileAlterationObserver(final File directory, final FileFilter fileFilter,
...
     protected FileAlterationObserver(final FileEntry rootEntry, final FileFilter fileFilter,
                                      final IOCase caseSensitivity) {
-        if (rootEntry == null) {
-            throw new IllegalArgumentException("Root entry is missing");
-        }
-        if (rootEntry.getFile() == null) {
-            throw new IllegalArgumentException("Root directory is missing");
-        }
+        Objects.requireNonNull(rootEntry, "rootEntry");
+        Objects.requireNonNull(rootEntry.getFile(),"rootDirectory");
```
garydgregory, line 205: Fix formatting.

garydgregory:
> @garydgregory Please review and mark the cases you don't want to change for compatibility and/or design reasons. I'll change then all occurrences accordingly.

Hi @boris-unckel 

Sorry but I don't want to take the time to review each change in each of 31 files. If you do not want to revise the PR to minimize changes and duplication of null checks, that's fine as well. I might get to these kinds of changes later. Thank you for your contributions so far though :-)

src/main/java/org/apache/commons/io/FileUtils.java
boris-unckel: @garydgregory The other cases before I understood the pattern behind "-1". In this case a FileInputStream is created, afterwards copyLarge calls copy calls copyLarge which then does the check.
I'll fix the former cases and wait with removal for cases like this on your comment.

boris-unckel:
@garydgregory 
- Fixed the formatting (is there a official eclipse configuration you can recommend?)
- Removed single parameter checks in case of direct usage, 
- Kept single parameter checks when prior (different) Object creation happens (see case above)
- Kept multiple parameter checks
- Removed double checks in constructors, even if objects are created
- Kept all checks in constructors, where fail will be in post constructor usage of methods

garydgregory:
@boris-unckel 

PR #201, opened 2021-02-07
Fix Typos in JavaDoc, Comments and Tests

FelixRilling:
This PR fixes a number of smaller typos (none affecting compiled results).


FelixRilling:
A question regarding some other typos that are left: What is the stance on messages of exceptions and backwards compatibility? while *technically* fixing a typo in an exception message could break existing code testing the message for a substring but I would argue that that is a rather exotic case.

garydgregory:
> A question regarding some other typos that are left: What is the stance on messages of exceptions and backwards compatibility? while _technically_ fixing a typo in an exception message could break existing code testing the message for a substring but I would argue that that is a rather exotic case.

I say fix the typos.

PR #203, opened 2021-02-09
IO-719: Fixed error of copying directories between different file systems

ashcheglov:
Also fixed relativize error for different types of paths
(absolute/relative)


garydgregory (changes requested):
No tabs please. See broken build.

ashcheglov:
Looks like all checks passed, excluding early access environments

src/main/java/org/apache/commons/io/file/CopyDirectoryVisitor.java
```diff
@@ -153,7 +155,7 @@ public FileVisitResult preVisitDirectory(final Path directory, final BasicFileAt
 
     @Override
     public FileVisitResult visitFile(final Path sourceFile, final BasicFileAttributes attributes) throws IOException {
-        final Path targetFile = targetDirectory.resolve(sourceDirectory.relativize(sourceFile));
+        final Path targetFile = targetDirectory.resolve(sourceDirectory.relativize(sourceFile).toString());
```
garydgregory, line 158: As above...

garydgregory:
@ashcheglov 
The jar file you refer to in the new tests does not exist, you should put in in the same place as the other test fixtures: `/src/test/resources/org/apache/commons/io`.

garydgregory:
@ashcheglov 
Thank  you for your PR, I applied the patch in two separate commits with additional clean ups and minor refactoring. Please see git master and close this PR if your use case is satified.

ashcheglov:
Cool, thanks!

PR #204, opened 2021-02-11
Changed IndexOOB to ArrayIndexOOB

MartensCedric:
This is slightly more precise : https://stackoverflow.com/questions/34266174/difference-between-arrayindexoutofboundsexception-and-indexoutofboundsexception


garydgregory:
-1: Probably a sood idea but not in this form. Most (if not all) of Commons IO uses `IndexOutOfBoundsException` for this type of failure, so if you want to make such a change, it should be done consistently _throughout_ the code, not in just one random location. Currently I see 10 locations that throw  `IndexOutOfBoundsException`.

garydgregory:
Closing: No reply since my February 16 comment.

PR #205, opened 2021-02-19
[IO-720] Fix error about usage of DirectBuffer in JRE 16/17

XenoAmess:
usage of DirectBuffer will cause IllegalAccessError.
This pr aims to fix it.
https://issues.apache.org/jira/browse/IO-720


XenoAmess:
NOTICE: 
jdk 17 will still fail CI, but that is another problem to fix, and shall no in this pr. 
At least 16 should be OK now.

XenoAmess:
@garydgregory @sebbASF please review. thanks.

PR #208, opened 2021-02-23
fix ci error on jdk17

XenoAmess:
As title.
Not a big thing, so do not think need a JIRA ticket.,

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -256,7 +256,25 @@ file comparators, endian transformation classes, and much more.
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-inline</artifactId>
-      <version>3.7.7</version>
+      <version>3.8.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>net.bytebuddy</groupId>
+      <artifactId>byte-buddy</artifactId>
+      <version>${byte-buddy.version}</version>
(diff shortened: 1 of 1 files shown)


XenoAmess:
@garydgregory please review. thanks. this one should be very easy to review.

garydgregory:
Is the mix of changing mockito AND adding a Maven profile intentional?

garydgregory:
Hi @XenoAmess 

-1 for several reasons:

Nits: The title of this PR ("fix ci error on jdk17") is misleading as (1) the PR does not fix anything, it just ignores a part of the build and (2) this is also for an Early Access Java build (jdk17-ea) and not a released Java version of 17 which is not there yet. 

Main: The 17-ea build is marked in our GitHub action as "experimental" and is allowed to fail by definition. The whole point (in my mind at least) for testing on EA builds is exactly what we are seeing now: a failure. In this case, it is quite nice to see that JaCoCo is not ready for Java 17 yet; this lets anyone who cares about the FOSS ecosystem go and help JaCoCo get up to speed on Java 17. If by these efforts we can get a new release of JaCoCo out sooner for us EA adventurers, then we are all better off.

XenoAmess:
> Is the mix of changing mockito AND adding a Maven profile intentional?

yes. that is needed as mockito use older version of buddy which cannot deal with jdk17 binary

PR #215, opened 2021-03-22
IO-726 Add MemoryMappedInputStream.

shollander:
Adds a `MemoryMappedInputStream` class bridging memory mapped files and `InputStream`s. Makes it easy to use memory mapped files and turbocharge disk IO.

Fixes IO-726.

### DIFF
--- a/src/main/java/org/apache/commons/io/ByteBufferCleaner.java
+++ b/src/main/java/org/apache/commons/io/ByteBufferCleaner.java
@@ -1,0 +1,144 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.commons.io;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.nio.ByteBuffer;
+
+/**
+ * Utility to manually clean a direct {@link ByteBuffer}. Without manual
+ * intervention, direct ByteBuffers will be cleaned eventually upon garbage
+ * collection. However, this should be be relied upon since it may not occur in
+ * a timely fashion - especially since off heap ByeBuffers don't put pressure on
+ * the garbage collector.
+ * 
+ * <p>
+ * <b>Warning:</b> Do not attempt to use a direct {@link ByteBuffer} that has
+ * been cleaned or bad things will happen. Don't use this class unless you can
+ * ensure that the cleaned buffer will not be accessed anymore.
+ * </p>
+ * <p>
+ * See <a href=https://bugs.openjdk.java.net/browse/JDK-4724038>JDK-4724038</a>
+ * </p>
+ * 
+ * @since 2.9.0
+ */
+public class ByteBufferCleaner {
+
+    private static final Cleaner cleaner = getCleaner();
+
+    private static Cleaner getCleaner() {
+        boolean isJava8;
+        try {
+            Class.forName("sun.misc.Cleaner");
+            isJava8 = true;
+        } catch (ClassNotFoundException e) {
+            isJava8 = false;
+        }
+        Cleaner cleaner = null;
+        try {
+            cleaner = isJava8 ? new Java8Cleaner() : new Java9plusCleaner();
+        } catch (Exception e) {
+            System.err.println("Failed to load ByteBuffer cleaner");
+            e.printStackTrace();
+        }
+        return cleaner;
+    }
+
+    /**
+     * Release memory held by the given {@link ByteBuffer}
+     * 
+     * @param buffer to release
+     */
+    public static void clean(final ByteBuffer buffer) {
+        cleaner.clean(buffer);
+    }
+
+    /**
+     * Report if were able to load a suitable cleaner for the current JVM.
+     * Attempting to call {@link #clean(ByteBuffer)} when this method returns false
+     * will result in an exception
+     * 
+     * @return {@code true} if cleaning is supported, {@code false} otherwise
+     */
+    public static boolean isSupported() {
+        return cleaner != null;
+    }
+
+    private static interface Cleaner {
+        void clean(ByteBuffer buffer);
+    }
+
+    private static class Java8Cleaner implements Cleaner {
+        private final Method cleanerMethod;
+        private final Method cleanMethod;
+
+        private Java8Cleaner() {
+            try {
+                Class<?> directBufferClass = Class.forName("sun.nio.ch.DirectBuffer");
+                Class<?> cleanerClass = Class.forName("sun.misc.Cleaner");
+                this.cleanerMethod = directBufferClass.getMethod("cleaner");
+                this.cleanMethod = cleanerClass.getMethod("clean");
+            } catch (ClassNotFoundException | NoSuchMethodException | SecurityException e) {
+                throw new IllegalStateException("Failed to initialize Java8Cleaner", e);
+            }
+        }
+
+        @Override
+        public void clean(final ByteBuffer buffer) {
+            try {
+                Object cleaner = cleanerMethod.invoke(buffer);
+                if (cleaner != null) {
+                    cleanMethod.invoke(cleaner);
+                }
+            } catch (IllegalAccessException | IllegalArgumentException | InvocationTargetException e) {
+                throw new RuntimeException("Failed to clean direct buffer", e);
+            }
+
+        }
+    }
+
+    private static class Java9plusCleaner implements Cleaner {
+        private final Object theUnsafe;
+        private final Method invokeCleaner;
+
+        private Java9plusCleaner() {
+            try {
+                Class<?> unsafeClass = Class.forName("sun.misc.Unsafe");
+                Field field = unsafeClass.getDeclaredField("theUnsafe");
+                field.setAccessible(true);
+                this.theUnsafe = field.get(null);
+                this.invokeCleaner = unsafeClass.getMethod("invokeCleaner", ByteBuffer.class);
+            } catch (ClassNotFoundException | IllegalAccessException | NoSuchFieldException | SecurityException
+                    | NoSuchMethodException e) {
+                throw new IllegalStateException("Failed to initialize Java9plusCleaner", e);
+            }
+        }
+
+        @Override
+        public void clean(final ByteBuffer buffer) {
+            try {
+                this.invokeCleaner.invoke(theUnsafe, buffer);
+            } catch (IllegalAccessException | IllegalArgumentException | InvocationTargetException e) {
+                throw new RuntimeException("Failed to clean direct buffer", e);
+            }
+
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


garydgregory:
@shollander 
The Java 17-EA builds but I can't see why; may you please rebase on master, all builds are green there.

shollander:
@garydgregory Issues addressed.

src/main/java/org/apache/commons/io/input/MemoryMappedFileInputStream.java
```diff
@@ -0,0 +1,169 @@
...
+ * An {@link InputStream} that utilizes memory mapped files to improve
+ * performance. A sliding window of the file is mapped to memory to avoid
+ * mapping the entire file to memory at one time. The size of the sliding buffer
+ * is user configurable.
+ * 
+ * For most operating systems, mapping a file into memory is more expensive than
+ * reading or writing a few tens of kilobytes of data. From the standpoint of
+ * performance it is generally only worth mapping relatively large files into
+ * memory. Use of this class can provide approximately a 25% increase in
+ * throughput when reading large files. <br>
```
garydgregory, line 42: Close all HTML tags. You might as well use a paragraph here.

src/main/java/org/apache/commons/io/input/MemoryMappedFileInputStream.java
```diff
@@ -0,0 +1,169 @@
...
+ * may still further improve performance. For example:
+ * 
+ * <pre>
+ * new BufferedInputStream(new GzipInputStream(new MemoryMappedFileInputStream(path))))
+ * </pre>
+ * 
+ * will greatly outperform:
+ * 
+ * <pre>
+ * new GzipInputStream(new MemoryMappedFileInputStream(path))
```
garydgregory, line 54: This comparison does not make sense. You should compare using this class with a code fragment that does not use it.

src/main/java/org/apache/commons/io/input/MemoryMappedFileInputStream.java
garydgregory, line 42: I would not make any performance claims without backing it up with evidwnce like a test anyone can run on their combination of OS/JRE/file. IOW, provide a test where the user can specify the file size in a sys prop or command line for example.

src/main/java/org/apache/commons/io/ByteBufferCleaner.java
```diff
@@ -0,0 +1,144 @@
...
+ */
+package org.apache.commons.io;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.nio.ByteBuffer;
+
+/**
+ * Utility to manually clean a direct {@link ByteBuffer}. Without manual
```
garydgregory, line 25: Move this class to the input package and make it package private to reduce the public API surface, we can make it public later.

src/main/java/org/apache/commons/io/input/MemoryMappedFileInputStream.java
```diff
@@ -0,0 +1,169 @@
...
+ * </pre>
+ * 
+ * @since 2.9.0
+ */
```
garydgregory, line 58: We are up to 2.10.0 now 😉

src/main/java/org/apache/commons/io/input/MemoryMappedFileInputStream.java
```diff
@@ -0,0 +1,169 @@
...
+            nextBuffer();
+            if (!buffer.hasRemaining()) {
+                return EOF;
+            }
+        }
+        return Short.toUnsignedInt(buffer.get());
+    }
+
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
```
garydgregory, line 99: Use final where you can.

src/test/java/org/apache/commons/io/ByteBufferCleanerTest.java
```diff
@@ -0,0 +1,49 @@
...
+package org.apache.commons.io;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.nio.ByteBuffer;
+
+import org.apache.commons.lang3.RandomUtils;
+import org.junit.jupiter.api.Test;
+
+class ByteBufferCleanerTest {
```
garydgregory, line 26: The style of this component is to make test classes and methods public.

garydgregory (changes requested):
@shollander 
Thank you for your update. Please see my 7 comments.

src/test/java/org/apache/commons/io/ByteBufferCleanerTest.java
shollander: Seems like a holdover from legacy JUnit. Why bother when JUnit doesn't require it anymore?

src/main/java/org/apache/commons/io/input/MemoryMappedFileInputStream.java
shollander: This comment has nothing to do with the performance of this class. It is just noting that user's shouldn't think there is no need for buffering after using this class. In the example given, buffering is needed to improve the performance of the gzip compression in memory.

src/main/java/org/apache/commons/io/ByteBufferCleaner.java
shollander: Done

src/main/java/org/apache/commons/io/input/MemoryMappedFileInputStream.java
shollander: Done

src/main/java/org/apache/commons/io/input/MemoryMappedFileInputStream.java
shollander: Fixed.

src/test/java/org/apache/commons/io/ByteBufferCleanerTest.java
garydgregory: @shollander
 Because you want to match the existing style we use.

src/test/java/org/apache/commons/io/input/MemoryMappedFileInputStreamTest.java
```diff
@@ -0,0 +1,248 @@
...
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.RandomUtils;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class MemoryMappedFileInputStreamTest {
```
garydgregory, line 39: Make public.

garydgregory:
@shollander 

garydgregory:
@shollander

src/test/java/org/apache/commons/io/input/MemoryMappedFileInputStreamTest.java
shollander: Done.

src/test/java/org/apache/commons/io/ByteBufferCleanerTest.java
shollander: Fixed.

shollander:
> 
> 
> @shollander
> May you please rebase on master?

Done.

src/main/java/org/apache/commons/io/input/ByteBufferCleaner.java
```diff
@@ -0,0 +1,144 @@
...
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.nio.ByteBuffer;
+
+/**
+ * Utility to manually clean a direct {@link ByteBuffer}. Without manual
+ * intervention, direct ByteBuffers will be cleaned eventually upon garbage
+ * collection. However, this should be be relied upon since it may not occur in
```
garydgregory, line 27: "this should be be relied"  -> "this should NOT be relied"

src/main/java/org/apache/commons/io/input/ByteBufferCleaner.java
```diff
@@ -0,0 +1,144 @@
...
+ * <p>
+ * <b>Warning:</b> Do not attempt to use a direct {@link ByteBuffer} that has
+ * been cleaned or bad things will happen. Don't use this class unless you can
+ * ensure that the cleaned buffer will not be accessed anymore.
+ * </p>
+ * <p>
+ * See <a href=https://bugs.openjdk.java.net/browse/JDK-4724038>JDK-4724038</a>
+ * </p>
+ * 
+ * @since 2.10.0
```
garydgregory, line 40: You do need a since tag since this is package private. Also new public and protected elements should be since 2.12.0 ;-)

src/main/java/org/apache/commons/io/input/MemoryMappedFileInputStream.java
```diff
@@ -0,0 +1,169 @@
...
+ * new BufferedInputStream(new GzipInputStream(new MemoryMappedFileInputStream(path))))
+ * </pre>
+ * 
+ * will greatly outperform:
+ * 
+ * <pre>
+ * new GzipInputStream(new MemoryMappedFileInputStream(path))
+ * </pre>
+ * 
+ * @since 2.10.0
```
garydgregory, line 57: "2.10.0" -> "2.12.0"

src/main/java/org/apache/commons/io/input/MemoryMappedFileInputStream.java
```diff
@@ -0,0 +1,169 @@
...
+    private final int bufferSize;
+    private final FileChannel channel;
+    private ByteBuffer buffer = EMPTY_BUFFER;
+    private boolean closed = false;
+    /**
+     * The starting position (within the file) of the next sliding buffer.
+     */
+    private long nextBufferPosition = 0;
+
+    public MemoryMappedFileInputStream(final Path file) throws IOException {
```
garydgregory, line 77: Please add a Javadoc comment: "Constructs..."

src/main/java/org/apache/commons/io/input/MemoryMappedFileInputStream.java
```diff
@@ -0,0 +1,169 @@
...
+        this(file, DEFAULT_BUFFER_SIZE);
+    }
+
+    public MemoryMappedFileInputStream(final Path file, final int bufferSize) throws IOException {
```
garydgregory, line 81: Please add a Javadoc comment: "Constructs..."

garydgregory (changes requested):
Thank you for the update. Please see my few remaining (new) comments.

src/main/java/org/apache/commons/io/input/ByteBufferCleaner.java
shollander: fixed.

src/main/java/org/apache/commons/io/input/ByteBufferCleaner.java
shollander: fixed.

src/main/java/org/apache/commons/io/input/MemoryMappedFileInputStream.java
shollander: fixed.

garydgregory:
@shollander 

I modified this PR 215 for git master with:
- Do not write the console.
- Simplify cleaner check support.
- Simplify exception handling.
- Don't initialize instance variables to their default values.
- Un-nest else clause.
- Declaring interfaces as static is redundant.
- Use EOF constant instead of magic number.
- Don't throw RuntimeException, use IllegalStateException.
- Pacakge private class does not need public method.
- Use final.
- Improve Javadocs.
- Javadoc sentences should end in a period.
- Javadoc do not need first sentence to be in an HTML p element.
- No need to use `this.` when you do not need to.
- In-line some single use variables.
- Sort members.

Please check it out. Feel free to rebase this PR on master or provide a new PR if you need more for your use case.
TY!

garydgregory:
Code in git master, no reply from the original author, closing.

PR #217, opened 2021-03-26
IO-727: FilenameUtils directoryContains() should handle files with the same prefix

trungPa:
- Fix `FilenameUtils.directoryContains`. Add a file separator to the end of the parent directory to avoid treating files having the same prefix as subdirectories. `directoryContains` will return false when either parent or child is empty.
- Explicitly state in the docs that the input paths are expected to be canonical.
- Add a new test case for FilenameUtils.directoryContains.
- Replace `testIO466` with a new test case to demonstrate the bug (the current `testIO466` will always pass because fooFile doesn't exist)

Jira: https://issues.apache.org/jira/browse/IO-727

### DIFF
--- a/src/test/java/org/apache/commons/io/FileUtilsDirectoryContainsTestCase.java
+++ b/src/test/java/org/apache/commons/io/FileUtilsDirectoryContainsTestCase.java
@@ -147,6 +147,12 @@ public class FileUtilsDirectoryContainsTestCase {
     }
 
     @Test
+    public void testIO466() throws IOException {
+        final File fooFile = new File(directory1.getParent(), "directory1.txt");
+        assertFalse(FileUtils.directoryContains(directory1, fooFile));
+    }
+
+    @Test
     public void testFileHavingSamePrefixBug() throws IOException {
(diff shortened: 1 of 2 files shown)


garydgregory:
@trungPa
May you please rebase on master? I see failures on Java 17 but the logs have expired.

src/test/java/org/apache/commons/io/FileUtilsDirectoryContainsTestCase.java
```diff
@@ -134,9 +134,16 @@ public void testSameFile() throws IOException {
     }
 
     @Test
-    public void testIO466() throws IOException {
```
garydgregory, line 137: Please don't remove an existing test, it makes reviewing harder.

trungPa: The fix has been applied in d42b9286eb7a534f1b2b6243858ff7ef6fcac14c so isn't it time to close this pr?
anyway, I've rebased and restored the test.

PR #221, opened 2021-04-09
Replace construction of FileInputStream and FileOutputStream objects with Files NIO APIs.

arturobernalg:


garydgregory:
@arturobernalg 
May you please rebase on master which should allow the Java 17-EA build to pass.

arturobernalg:
> @arturobernalg
> May you please rebase on master which should allow the Java 17-EA build to pass.

HI @garydgregory 
Done.
TY

src/test/java/org/apache/commons/io/IOUtilsTestCase.java
```diff
@@ -118,7 +116,7 @@ public void setUp() {
             if (!testFile.getParentFile().exists()) {
                 throw new IOException("Cannot create file " + testFile + " as the parent directory does not exist");
             }
-            final BufferedOutputStream output = new BufferedOutputStream(new FileOutputStream(testFile));
+            final BufferedOutputStream output = new BufferedOutputStream(Files.newOutputStream(testFile.toPath()));
```
garydgregory, line 119: In these test methods, you keep calling `toPath()` over and over, why not just call it once?

arturobernalg: HI @garydgregory 
you're right. Changed.
TY

jochenw:
I am unsure about this one. First of all, it changes way too many files, in my opinion.
Then, I see no real benefit, when comparing Files.newInputstream() to new FileInputStream(). As far as I know, there is no real difference between those streams. In particular, neither is buffered.

arturobernalg:
> Files.newInputstream() to new FileInputStream()

HI @jochenw 
This is the reason --> 
'Basically, FileInputStream and FileOutputStream both use a finalizer to ensure that the resources are closed.
Even with programmers do call close() after using FileInputStream, its finalize() method will still be called. In other words, still get the side effect of the FinalReference being registered at FileInputStream allocation time, and also reference processing to reclaim the FinalReference during GC (any GC solution has to deal with this).'

garydgregory:
> 
> 
> > Files.newInputstream() to new FileInputStream()
> 
> HI @jochenw
> This is the reason -->
> 'Basically, FileInputStream and FileOutputStream both use a finalizer to ensure that the resources are closed.
> Even with programmers do call close() after using FileInputStream, its finalize() method will still be called. In other words, still get the side effect of the FinalReference being registered at FileInputStream allocation time, and also reference processing to reclaim the FinalReference during GC (any GC solution has to deal with this).'

Indeed, @jochenw do notice `java.io.FileInputStream.finalize()` and `java.io.FileOutputStream.finalize()`. I'll review further but I am incline to merged based on the `finalize()` issue. While in the long term I think finalization is going away, we are still on Java 8 for now.

PR #226, opened 2021-05-10
Prevent infinite loop with AbstractCharacterFilterReader if EOF is filtered out

robtimus:
`AbstractCharacterFilterReader.read()` only aborts its loop if `filter(ch)` returns false. If that method returns `true` for EOF, you'll get an infinite loop if the backing `read()` call keeps returning EOF. That's really easy to achieve with `CharacterFilterReader`.

This small fix aborts the loop not only if `filter(ch)` returns false, but also if the backing `read()` call returns EOF.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/AbstractCharacterFilterReader.java
+++ b/src/main/java/org/apache/commons/io/input/AbstractCharacterFilterReader.java
@@ -42,7 +42,7 @@ public abstract class AbstractCharacterFilterReader extends FilterReader {
         int ch;
         do {
             ch = in.read();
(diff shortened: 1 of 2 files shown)


garydgregory:
@robtimus 
I pushed an alternate commit to master, please verify and close this PR if it works for you.

garydgregory:
What is duplicated?

robtimus:
I had added the `testReadFilteringEOF` method at the bottom, whereas you added it a bit higher. That didn't trigger any conflicts during rebase, it just added the same method, leading to two methods with the same name and almost the same contents.

PR #235, opened 2021-05-19
[IO-732] CharArrayWriter

pjfanning:
A draft implementation of https://issues.apache.org/jira/browse/IO-732.
* needs more tests
* needs better javadoc

### DIFF
--- a/src/main/java/org/apache/commons/io/IOUtils.java
+++ b/src/main/java/org/apache/commons/io/IOUtils.java
@@ -342,7 +342,7 @@ public class IOUtils {
      * Returns a new char array of size {@link #DEFAULT_BUFFER_SIZE}.
      *
      * @return a new char array of size {@link #DEFAULT_BUFFER_SIZE}.
-     * @since 2.9.0
+     * @since 2.10.0
      */
     public static char[] charArray() {
         return charArray(DEFAULT_BUFFER_SIZE);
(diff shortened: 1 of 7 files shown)


garydgregory:
@pjfanning 
Thank you for your contribution. I will look at this in the context of 2.10.0 as I am about done preparing the 2.9.0 release candidate.

pjfanning:
@garydgregory thanks, I can update the javadocs in my PR later to say 2.10.0. My PR uses the IOUtils charArray method - would it be possible to make that public in the 2.9.0 release? Accoring the the javadoc, the charArray methods were only recently added and it might be tidier to make them public now than making them public later.

garydgregory:
@pjfanning 
Sorry, I am in the middle of the release candidate process ATM.

jochenw:
I'd like to see an explanation for this statement:



        It is also a good alternative to java.io.StringWriter class.

pjfanning:
@jochenw  I guess the point is that this CharArrayWriter can also be used to generate a String but I suppose commons-io already has StringBuilderWriter and maybe that is a better choice if building a String is what you want. I can remove the line from the Javadoc - I'm neutral abut whether it is needed.

jochenw:
The point I am trying to make: Right now, I wouldn't know when, and why, to use which? And, my guess is: If I don't know, other users won't know, either.

garydgregory:
@jochenw May we close this PR then?

garydgregory:
Closing per  @jochenw's comment.

PR #251, opened 2021-07-25
[IO-744] FileWriterWithEncoding for an existing file no longer truncates the file.

rbri:


garydgregory:
@rbri 
Thank you for your PR. See comments.

src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
```diff
@@ -227,7 +227,12 @@ private static Writer initWriter(final File file, final Object encoding, final b
         OutputStream stream = null;
         final boolean fileExistedAlready = file.exists();
         try {
-            stream = Files.newOutputStream(file.toPath(),  append ? StandardOpenOption.APPEND : StandardOpenOption.CREATE);
+            if (append) {
```
garydgregory, line 230: Not sure why the fix has to be exploded into so much code when all you need is an additional argument to the JRE API. Please see git master for a much simpler fix.

src/test/java/org/apache/commons/io/output/FileWriterWithEncodingTest.java
```diff
@@ -197,6 +198,25 @@ public void constructor_File_nullFile() {
         assertFalse(file1.exists());
     }
 
+    @Test
```
garydgregory, line 201: This test is now in git master, TY. You are credited in `changes.xml`.

src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
rbri: @garydgregory Thanks for the warm welcome as someone trying to help.
Have done another PR that shows, that your optimized version of the fix now breaks the ctor with the append option set to true.

PR #252, opened 2021-07-26
[IO-744] test that shows the last fix did not work

rbri:
one more test - this time the test shows that the 'optimized' fix for issue_IO-744 breaks the append constructor option

### DIFF
--- a/src/test/java/org/apache/commons/io/output/FileWriterWithEncodingTest.java
+++ b/src/test/java/org/apache/commons/io/output/FileWriterWithEncodingTest.java
@@ -86,6 +86,20 @@ public class FileWriterWithEncodingTest {
     }
 
     @Test
+    public void constructorAppend_File_existingFile_withContent() throws Exception {
+        try (FileWriter fw1 = new FileWriter(file1);) {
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/io/output/FileWriterWithEncodingTest.java
```diff
@@ -85,6 +85,20 @@ public void constructor_File_existingFile_withContent() throws Exception {
         assertEquals(4, file1.length());
     }
 
+    @Test
+    public void constructorAppend_File_existingFile_withContent() throws Exception {
+        try (FileWriter fw1 = new FileWriter(file1);) {
```
garydgregory, line 90: The extra semi colon is not needed FYI.

rbri: ups :-D, have fixed this (also for the other test) - i think you can squash the two commits

garydgregory:
@rbri 
Thank you for your PR. I've brought in the test and added a fix in git master. I did not merge this PR as I did not want to bring in a PR that breaks the build.

PR #253, opened 2021-07-29
Minor changes

arturobernalg:
* Remove Unused import.
* Make Method void.


arturobernalg:
> Let's leave these internal API sigatures alone for now.

Hi @garydgregory 
understood
TY

PR #254, opened 2021-07-29
Simplify Bitwise operation with equivalent but more simple.

arturobernalg:


garydgregory:
I am guessing this code was carefully and purposefully written to be easy to read and maintain. Some of these changes work away from that goal. The original author, if still around, can comment.

arturobernalg:
> I am guessing this code was carefully and purposefully written to be easy to read and maintain. Some of these changes work away from that goal. The original author, if still around, can comment.

HI @garydgregory 
OK,
TY

PR #265, opened 2021-08-28
Adopt is no more, long live Zulu!

geertjanw:


garydgregory:
You'll need to provide more information as to why this change is necessary. Adopt is moving to Eclipse per the announcement on https://adoptopenjdk.net/ and https://blog.adoptopenjdk.net/2021/03/transition-to-eclipse-an-update/ 
I'd rather stay with Eclipsw.

geertjanw:
Yes, in that case, "temurin" would be the one to use, rather than "adopt" or "zulu". You certainly need to move away from "adopt" -- and so now the choice is between either "temurin" or "zulu". What's clear is that Zulu has been around for a good while longer than Temurin, which is completely new (not simply a rename, but a new distribution, which may or may not have all the JDK versions you need, while Zulu has had all of them all along). Full disclosure: I work for Azul, which produces Zulu, though which also provides support for Temurin, see the differences here: https://www.azul.com/support-for-temurin/

PR #269, opened 2021-09-03
Fix usage of assertNotNull

dfa1:
Signature of assertNoNull() is taking "actual" as first parameter:

    public static void assertNotNull(Object actual, String message) {
        AssertNotNull.assertNotNull(actual, message);
    }

Signed-off-by: Davide Angelocola <davide.angelocola@gmail.com>


dfa1:
@garydgregory can you please review this as well?

PR #270, opened 2021-09-11
IO-748 Documentation correction in FileUtils

elangoravi:
FileNotFoundException included in FileUtils.moveToDirectory() Javadoc as per comments in IO-748.


garydgregory:
https://issues.apache.org/jira/browse/IO-748

garydgregory:
This patch is incomplete because it does not take into account all callers of `validateMoveParameters(File, File)`. Closing this PR in favor of a better patch in git master.

PR #277, opened 2021-09-25
Refactoring

caetanobca:


src/main/java/org/apache/commons/io/filefilter/WildcardFilter.java
```diff
@@ -114,7 +118,9 @@ public WildcardFilter(final String wildcard) {
      * @throws IllegalArgumentException if the pattern array is null
      */
     public WildcardFilter(final String... wildcards) {
-        requireNonNull(wildcards, "wildcards");
+        if (wildcards == null) {
```
garydgregory, line 121: We now use `requireNonNull` in the style Java's own.

garydgregory:
Hi @caetanobca 
You'll need to run `mvn` before you submit a PR and check if the build passes. This PR will not pass as it breaks binary compatibility.

src/main/java/org/apache/commons/io/filefilter/WildcardFilter.java
garydgregory: Ping

garydgregory:
Closing, no feedback from OP.

PR #291, opened 2021-10-31
Improve `ReaderInputStream` documentation

Marcono1234:
Resolves:
- https://issues.apache.org/jira/browse/IO-713
- https://issues.apache.org/jira/browse/IO-715

Please let me know if you want anything changed, or if I should split this pull request.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/ReaderInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/ReaderInputStream.java
@@ -243,16 +243,6 @@ public class ReaderInputStream extends InputStream {
     }
 
     /**
-     * The implementation by {@link ReaderInputStream} always returns 0.
-     */
-    @Override
-    public int available() throws IOException {
-        // Could implement this using Reader.ready(), but would be rather cumbersome
-        // and probably not very efficient
-        return 0;
-    }
-
-    /**
      * Fills the internal char buffer from the reader.
      *
      * @throws IOException If an I/O error occurs
--- a/src/main/java/org/apache/commons/io/input/ReaderInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/ReaderInputStream.java
@@ -128,8 +128,10 @@ public class ReaderInputStream extends InputStream {
      * Constructs a new {@link ReaderInputStream} with a default input buffer size of {@value #DEFAULT_BUFFER_SIZE}
      * characters.
      *
-     * <p>The encoder created for the specified charset will use {@link CodingErrorAction#REPLACE} for malformed input
+     * <p>
+     * The encoder created for the specified charset will use {@link CodingErrorAction#REPLACE} for malformed input
      * and unmappable characters.
+     * </p>
      *
      * @param reader the target {@link Reader}
      * @param charset the charset encoding
@@ -141,8 +143,10 @@ public class ReaderInputStream extends InputStream {
     /**
      * Constructs a new {@link ReaderInputStream}.
      *
-     * <p>The encoder created for the specified charset will use {@link CodingErrorAction#REPLACE} for malformed input
+     * <p>
+     * The encoder created for the specified charset will use {@link CodingErrorAction#REPLACE} for malformed input
      * and unmappable characters.
+     * </p>
      *
      * @param reader the target {@link Reader}.
      * @param charset the charset encoding.
@@ -161,8 +165,10 @@ public class ReaderInputStream extends InputStream {
     /**
      * Constructs a new {@link ReaderInputStream}.
      *
-     * <p>This constructor does not call {@link CharsetEncoder#reset() reset} on the provided encoder. The caller
+     * <p>
+     * This constructor does not call {@link CharsetEncoder#reset() reset} on the provided encoder. The caller
      * of this constructor should do this when providing an encoder which had already been in use.
+     * </p>
      *
      * @param reader the target {@link Reader}
      * @param charsetEncoder the charset encoder
@@ -175,8 +181,10 @@ public class ReaderInputStream extends InputStream {
     /**
      * Constructs a new {@link ReaderInputStream}.
      *
-     * <p>This constructor does not call {@link CharsetEncoder#reset() reset} on the provided encoder. The caller
+     * <p>
+     * This constructor does not call {@link CharsetEncoder#reset() reset} on the provided encoder. The caller
      * of this constructor should do this when providing an encoder which had already been in use.
+     * </p>
      *
      * @param reader the target {@link Reader}
      * @param charsetEncoder the charset encoder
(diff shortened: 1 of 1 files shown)


garydgregory:
@Marcono1234 Do not change `@link` to `@code`.

Marcono1234:
Not even when referring to the class itself (`{@link ReaderInputStream }`)?
To me it appears that this does not add any value; when you are already reading the documentation of `ReaderInputStream` you don't want to open its documentation again.

See also https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#inlinelinks (in case you have not noticed it in the commit message); though it might be slightly dated.

garydgregory:
> Not even when referring to the class itself (`{@link ReaderInputStream }`)? To me it appears that this does not add any value; when you are already reading the documentation of `ReaderInputStream` you don't want to open its documentation again.

To me, the value is that you get a shortcut to the class comment.

garydgregory:
> See also https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#inlinelinks (in case you have not noticed it in the commit message); though it might be slightly dated.

Obviously dated indeed since this document talks about old style HTML links which were underlined, and which Javadoc no longer generates. You'll notice on our Javadoc pages, the hyperlink coloring is rather (IMO) subtle and no longer the harsh bright blue with underline style from the Web 1.0 days.

Marcono1234:
Ok, I have omitted the changes to replace `{@link ...}`.

Linking to the class itself could also cause confusion because you might be wondering why there is a link (e.g. is that a different class and have I misread its name?, is this linking to a specific subsection?), but your point makes sense as well.

Marcono1234:
Also unrelated to this, but it looks like many of your commits use an incorrect (?) e-mail address, for example https://github.com/apache/commons-io/commit/8f7a6430d535b0227fc25cb917af94b9339c3b19.patch: <code>gar<del>y</del>dgregory@...</code> (the first `y` is missing).
If you actually own that e-mail address as well, it might make sense to [add it to your GitHub profile](https://docs.github.com/en/account-and-profile/setting-up-and-managing-your-github-user-account/managing-email-preferences/setting-your-commit-email-address#setting-your-commit-email-address-on-github) to get the commits attributed to you.

src/main/java/org/apache/commons/io/input/ReaderInputStream.java
```diff
@@ -129,6 +128,9 @@ public ReaderInputStream(final Reader reader) {
      * Constructs a new {@link ReaderInputStream} with a default input buffer size of {@value #DEFAULT_BUFFER_SIZE}
      * characters.
      *
+     * <p>The encoder created for the specified charset will use {@link CodingErrorAction#REPLACE} for malformed input
+     * and unmappable characters.
+     *
```
garydgregory, line 133: Close HTML tags.

src/main/java/org/apache/commons/io/input/ReaderInputStream.java
```diff
@@ -156,6 +161,9 @@ public ReaderInputStream(final Reader reader, final Charset charset, final int b
     /**
      * Constructs a new {@link ReaderInputStream}.
      *
+     * <p>This constructor does not call {@link CharsetEncoder#reset() reset} on the provided encoder. The caller
+     * of this constructor should do this when providing an encoder which had already been in use.
+     *
```
garydgregory, line 166: Close HTML tags.

src/main/java/org/apache/commons/io/input/ReaderInputStream.java
```diff
@@ -185,6 +196,9 @@ public ReaderInputStream(final Reader reader, final CharsetEncoder charsetEncode
      * Constructs a new {@link ReaderInputStream} with a default input buffer size of {@value #DEFAULT_BUFFER_SIZE}
      * characters.
      *
+     * <p>The encoder created for the specified charset will use {@link CodingErrorAction#REPLACE} for malformed input
+     * and unmappable characters.
+     *
```
garydgregory, line 201: Close HTML tags.

src/main/java/org/apache/commons/io/input/ReaderInputStream.java
```diff
@@ -213,6 +230,16 @@ public void close() throws IOException {
...
+    /**
+     * The implementation by {@link ReaderInputStream} always returns 0.
+     */
+    @Override
+    public int available() throws IOException {
+        // Could implement this using Reader.ready(), but would be rather cumbersome
+        // and probably not very efficient
+        return 0;
+    }
+
```
garydgregory, line 242: This PR is only supposed to be about docs as stated in the PR title. Furthermore this change is not matched with a unit test.

src/main/java/org/apache/commons/io/input/ReaderInputStream.java
```diff
@@ -195,6 +209,9 @@ public ReaderInputStream(final Reader reader, final String charsetName) {
     /**
      * Constructs a new {@link ReaderInputStream}.
      *
+     * <p>The encoder created for the specified charset will use {@link CodingErrorAction#REPLACE} for malformed input
+     * and unmappable characters.
+     *
```
garydgregory, line 214: Close HTML tags.

garydgregory (changes requested):

garydgregory:
You have 2 PRs with same title, this is confusing and wasted reviewers' time. Please pick one and close the other.

src/main/java/org/apache/commons/io/input/ReaderInputStream.java
Marcono1234: This class directly extends `InputStream` and the [`InputStream.available()` documentation](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/InputStream.html#available()) says:
> The available method of InputStream always returns 0.

I could have used `return super.available();` instead, but then it would be less obvious to the reader why the documentation of this method here says that it always returns 0 but calls `super.available()`. So I wanted to avoid requiring them to look up the super implementation to verify that this method behaves correctly.

src/main/java/org/apache/commons/io/input/ReaderInputStream.java
Marcono1234: Have also moved the `<p>` on a separate line to match the style used above for the class documentation.

src/main/java/org/apache/commons/io/input/ReaderInputStream.java
garydgregory: Please drop this method, it duplicates the superclass' implementation.

garydgregory: ping.

Marcono1234: Sorry; have removed this method again now in the latest commit.

Marcono1234:
Can I resolve the corresponding Jira issues ([IO-713](https://issues.apache.org/jira/browse/IO-713) and [IO-715](https://issues.apache.org/jira/browse/IO-715)) now, or do you want to do that? (in case there is anything to consider when resolving them)

jochenw:
> Thanks a lot for merging the changes!
> 
> Can I resolve the corresponding Jira issues ([IO-713](https://issues.apache.org/jira/browse/IO-713) and [IO-715](https://issues.apache.org/jira/browse/IO-715)) now, or do you want to do that? (in case there is anything to consider when resolving them)

Feel free to close them. Thank you!

PR #294, opened 2021-11-01
Update cleanDirectory(final File) method in org.apache.commons.io.FileUtils.

Krone9891:
I have made a change to the method cleanDirectory(final File) in class org.apache.commons.io.FileUtils.

It first checks if the directory exists, then it cleans the directory by iterating though the files and deleting them. If the
given File is a simple file, it deletes it.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -339,19 +339,26 @@ public class FileUtils {
      * @throws IOException if an I/O error occurs.
      * @see #forceDelete(File)
      */
-    public static void cleanDirectory(final File directory) {
+    public static void cleanDirectory(final File directory) throws IOException {
+        if (directory == null) {
+            return;
+        }
         if (directory.exists()) {
             if (directory.isDirectory()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -340,19 +340,26 @@ public static long checksumCRC32(final File file) throws IOException {
...
+            if (directory.isDirectory()) {
+                if (Objects.requireNonNull(directory.listFiles(), "File is not directory").length != 0) {
+                    Arrays.stream(Objects.requireNonNull(directory.listFiles(), "File is not directory")).forEach(file -> {
+                        try {
+                            if (file.isDirectory() && Objects.requireNonNull(file.listFiles()).length != 0) {
+                                cleanDirectory(file);
+                            }
+                            Files.delete(directory.toPath());
+                        } catch (IOException e) {
+                            e.printStackTrace();
```
garydgregory, line 356: Never output to the console.

garydgregory (commented):
-1 The behavior is so different that I don't even need to run a build to know it will break tests; just run `mvn` from the command line to see the failures.

PR #296, opened 2021-11-02
IO-756: Update FileWriterWithEncoding to extend ProxyWriter

wodencafe:
This pull request simplifies org.apache.commons.io.output.FileWriterWithEncoding by removing the decorated object and extending org.apache.commons.io.output.ProxyWriter directly.
Update: Ticket is here https://issues.apache.org/jira/browse/IO-756

### DIFF
--- a/src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
+++ b/src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
@@ -46,9 +46,7 @@ import org.apache.commons.io.IOUtils;
  *
  * @since 1.4
  */
-public class FileWriterWithEncoding extends Writer {
-    // Cannot extend ProxyWriter, as requires writer to be
-    // known when super() is called
+public class FileWriterWithEncoding extends ProxyWriter {
 
     /**
      * Initializes the wrapped file writer. Ensure that a cleanup occurs if the writer creation fails.
@@ -85,9 +83,6 @@ public class FileWriterWithEncoding extends Writer {
         }
     }
 
-    /** The writer to decorate. */
-    private final Writer out;
-
     /**
      * Constructs a FileWriterWithEncoding with a file encoding.
      *
@@ -110,7 +105,7 @@ public class FileWriterWithEncoding extends Writer {
      * @throws IOException in case of an I/O error.
      */
     public FileWriterWithEncoding(final File file, final Charset encoding, final boolean append) throws IOException {
-        this.out = initWriter(file, encoding, append);
+        super(initWriter(file, encoding, append));
     }
 
     /**
@@ -135,7 +130,7 @@ public class FileWriterWithEncoding extends Writer {
      * @throws IOException in case of an I/O error.
      */
     public FileWriterWithEncoding(final File file, final CharsetEncoder charsetEncoder, final boolean append) throws IOException {
-        this.out = initWriter(file, charsetEncoder, append);
+        super(initWriter(file, charsetEncoder, append));
     }
 
     /**
@@ -160,7 +155,7 @@ public class FileWriterWithEncoding extends Writer {
      * @throws IOException in case of an I/O error.
      */
     public FileWriterWithEncoding(final File file, final String charsetName, final boolean append) throws IOException {
-        this.out = initWriter(file, charsetName, append);
+        super(initWriter(file, charsetName, append));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


garydgregory:
How is this simpler when there is _more_ code?

wodencafe:
@garydgregory Now the FileWriterWithEncoding extends ProxyWriter, and the actual Proxy class for writing is stored in ProxyWriter rather than inside FileWriterWithEncoding itself. We can probably also get rid of the write methods in FileWriterWithEncoding, and just let them delegate to the superclass ProxyWriter. This will reduce the amount of code significantly, but also the JavaDoc for those methods will be inherited from ProxyWriter.

Also starting with Java 11, there is the static Writer.nullWriter(), which can be used instead of the anonymous class instance NULL_WRITER (whenever Apache Commons IO upgrades minimum version to Java 11).

src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
```diff
@@ -46,9 +46,53 @@
  *
  * @since 1.4
  */
-public class FileWriterWithEncoding extends Writer {
-    // Cannot extend ProxyWriter, as requires writer to be
-    // known when super() is called
+public class FileWriterWithEncoding extends ProxyWriter {
+
+    private static final Writer NULL_WRITER = new Writer() {
```
kinow, line 51: Not sure if it's handling `null`s, or being used when some value is `null`. But maybe `NOOP_WRITER` would be a better name?

src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
```diff
@@ -160,7 +209,11 @@ public FileWriterWithEncoding(final File file, final String charsetName) throws
      * @throws IOException in case of an I/O error.
      */
     public FileWriterWithEncoding(final File file, final String charsetName, final boolean append) throws IOException {
-        this.out = initWriter(file, charsetName, append);
+        super(NULL_WRITER);
+
+        super.out = initWriter(file, charsetName, append);
```
kinow, line 214: Ah, I think now I understand this pull request.

@wodencafe I think we pass the `NULL_WRITER` to the parent `ProxyWriter` as a way to give it a valid object. But in reality, that `NULL_WRITER` is then replaced here when we call `super.out = ....`.

While I guess that would work, but not sure if modifying setting `super.out` twice (once after the `super(NULL_WRITER)`, then here :point_up: ) would be the best approach.

src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
kinow: Just saw your comment about the `Writer.nullWriter()`, I guess that's where you got the name from :+1: no need to change it.

kinow:
@wodencafe I was curious, and tried the following too before jumping to other PRs:

```diff
diff --git a/src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java b/src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
index 3d7cfb27..7fbeee23 100644
--- a/src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
+++ b/src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
@@ -46,10 +46,7 @@
  *
  * @since 1.4
  */
-public class FileWriterWithEncoding extends Writer {
-    // Cannot extend ProxyWriter, as requires writer to be
-    // known when super() is called
-
+public class FileWriterWithEncoding extends ProxyWriter {
     /**
      * Initializes the wrapped file writer. Ensure that a cleanup occurs if the writer creation fails.
      *
@@ -85,9 +82,6 @@ private static Writer initWriter(final File file, final Object encoding, final b
         }
     }
 
-    /** The writer to decorate. */
-    private final Writer out;
-
     /**
      * Constructs a FileWriterWithEncoding with a file encoding.
      *
@@ -110,7 +104,7 @@ public FileWriterWithEncoding(final File file, final Charset charset) throws IOE
      * @throws IOException in case of an I/O error.
      */
     public FileWriterWithEncoding(final File file, final Charset encoding, final boolean append) throws IOException {
-        this.out = initWriter(file, encoding, append);
+        super(initWriter(file, encoding, append));
     }
 
     /**
@@ -135,7 +129,7 @@ public FileWriterWithEncoding(final File file, final CharsetEncoder charsetEncod
      * @throws IOException in case of an I/O error.
      */
     public FileWriterWithEncoding(final File file, final CharsetEncoder charsetEncoder, final boolean append) throws IOException {
-        this.out = initWriter(file, charsetEncoder, append);
+        super(initWriter(file, charsetEncoder, append));
     }
 
     /**
@@ -160,7 +154,7 @@ public FileWriterWithEncoding(final File file, final String charsetName) throws
      * @throws IOException in case of an I/O error.
      */
     public FileWriterWithEncoding(final File file, final String charsetName, final boolean append) throws IOException {
-        this.out = initWriter(file, charsetName, append);
+        super(initWriter(file, charsetName, append));
     }
 
     /**
```

I think it achieves something similar to what you are suggesting here, but without the extra internal class? I haven't tested it, and haven't worked on IO lately, so other committers would have to take a look at it too. But WDYT?

Bruno

wodencafe:
Hi @kinow , thanks for reviewing the code. Actually I think that bottom approach is simpler, I'll give it a try, and update the PR accordingly.

wodencafe:
Hello again @kinow , just wanted to let you know that the changes have been made.

src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
```diff
@@ -245,7 +240,7 @@ public FileWriterWithEncoding(final String fileName, final String charsetName, f
      */
```
kinow, line 240: @wodencafe do we need to override the methods below? Could we just use the parent methods directly?

kinow:
@wodencafe we would also need a JIRA for this, could you create one? I think we will be able to make `FilterWriterWithEncoding` extend `ProxyWriter`. But in case we find a roadblock somewhere and we are not able to make `FileWriterWithEncoding` extend `ProxyWriter`, we should be able to at least improve the existing comment in the code explaining better why that's not possible (just in case it's indeed no possible :+1: )

Thanks!

wodencafe:
@kinow No problem, here is the associated JIRA ticket: https://issues.apache.org/jira/browse/IO-756

src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
wodencafe: Good point, we should just let the super class handle it.

garydgregory:
FYI: I should be able to review this over the weekend.

data443-cboyd:
Thank you @garydgregory and @kinow for the assistance with this pull request.

wodencafe:
Thank you @garydgregory and @kinow for the assistance with this pull request.

src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
```diff
@@ -59,7 +57,7 @@
      * @return the initialized writer
      * @throws IOException if an error occurs
      */
-    private static Writer initWriter(final File file, final Object encoding, final boolean append) throws IOException {
+    private static final Writer initWriter(final File file, final Object encoding, final boolean append) throws IOException {
```
kinow, line 60: This one is not really necessary @wodencafe ? `initWriter` is already private, so making it `final` won't change much for internal/external devs?

kinow (changes requested):
@wodencafe I had a look today with the IDE, debugging `FileWriterWithEncodingTest` and I think the updated code & constructor are working OK. One comment about the `final` being redundant. Thanks for creating the JIRA issue.

One last thing that makes it easier/quicker to review and merge is having a single commit in this pull request (not always needed, but for small changes it's useful). Would you be able to take a look at my comment for `final`, and also squash your commits, please?

Thank you!!!
Bruno

src/main/java/org/apache/commons/io/output/FileWriterWithEncoding.java
wodencafe: I usually do this out of habit, but after doing some research it seems that there is no functional difference for adding the `final` modifier to a private static nested class. It has been removed.

wodencafe:
@kinow The commits have been squashed and and updated as per your request. Thank you for the help!

kinow (approved):
Change looks good to me. Need to confirm if there's any binary compatibility issue, but other than that, +1. Thanks @wodencafe !

PR #299, opened 2021-11-05
IOUtils clean up javadoc

wodencafe:
This PR removes some redundant methods from IOUtils, specifically those which use String and StringBuffer, as the same methods that accept CharSequence as a parameter exist, and it is the interface that String and StringBuffer both implement.
There are some other minor code and documentation tweaks.

EDIT: This PR now just updates the javadoc of IOUtils to be more accurate and tidier.

### DIFF
--- a/src/main/java/org/apache/commons/io/IOUtils.java
+++ b/src/main/java/org/apache/commons/io/IOUtils.java
@@ -25,6 +25,7 @@ import java.io.CharArrayWriter;
 import java.io.Closeable;
 import java.io.EOFException;
 import java.io.File;
+import java.io.Flushable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -76,14 +77,14 @@ import org.apache.commons.io.output.UnsynchronizedByteArrayOutputStream;
  * The byte-to-char methods and char-to-byte methods involve a conversion step.
  * Two methods are provided in each case, one that uses the platform default
  * encoding and the other which allows you to specify an encoding. You are
- * encouraged to always specify an encoding because relying on the platform
- * default can lead to unexpected results, for example when moving from
+ * encouraged to always specify an encoding, because relying on the platform
+ * default can lead to unexpected results - for example, when moving from
  * development to production.
  * </p>
  * <p>
  * All the methods in this class that read a stream are buffered internally.
- * This means that there is no cause to use a {@code BufferedInputStream}
- * or {@code BufferedReader}. The default buffer size of 4K has been shown
+ * This means that there is no cause to use a {@link BufferedInputStream}
+ * or {@link BufferedReader}. The default buffer size of 8K has been shown
  * to be efficient in tests.
  * </p>
  * <p>
@@ -103,10 +104,10 @@ import org.apache.commons.io.output.UnsynchronizedByteArrayOutputStream;
  * This may improve performance for applications that need to do a lot of copying.
  * </p>
  * <p>
- * Wherever possible, the methods in this class do <em>not</em> flush or close
- * the stream. This is to avoid making non-portable assumptions about the
- * streams' origin and further use. Thus the caller is still responsible for
- * closing streams after use.
+ * Wherever possible, the methods in this class do <em>not</em> {@link Flushable#flush()}
+ * or {@link Closeable#close()} the stream. This is to avoid making non-portable
+ * assumptions about the streams' origin and further use. Thus the <i>caller</i> is still
+ * responsible for closing streams after use, unless otherwise specified by the method.
  * </p>
  * <p>
  * Origin of code: Excalibur.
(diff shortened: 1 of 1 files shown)


kinow (commented):
This would need a JIRA issue, and since public methods are being removed, I think we will need to bump it to the next major release (or just deprecate instead of removing methods). Thanks!

wodencafe:
Hi @kinow , thank you for taking the time to review this pull request. I have a question for you about the deprecated methods. Since CharSequence is the interface that is implemented by the types that the (removed in the PR) methods accepted as first parameter (String, StringBuffer), is it still necessary to retain those overloaded methods, since the existing methods that accept CharSequence are using the same logic?

kinow:
> Hi @kinow , thank you for taking the time to review this pull request. I have a question for you about the deprecated / removed methods. Since CharSequence is the interface that is implemented by the types that the (removed in the PR) methods accepted as first parameter (String, StringBuffer), is it still necessary to retain those overloaded methods, since the existing methods that accept CharSequence are using the same logic?

Hi @wodencafe the issue is backward compatibility.

If we remove public methods (or modify the code in a way that breaks binary backward compatibility [^1][^2] or sometimes behavioral backward compatibility [^3][^4]) we need to think about the users of the API first.

We could have users of Commons IO that are using those methods in their code. So if we remove these methods, users would expect a [major](https://semver.org/) version release so they are aware that there are API changes. Releasing major versions can be inconvenient to users too, as even those that were not affected will have to either review the changes or simply upgrade their code. For that reason, we accumulate changes that require major releases by deprecating unwanted code whenever possible.

I think your change makes sense, if users can rely on the parent type for String/StringBuffer. That would be less code to maintain I guess. We just may need to delay reviewing and merging it until we are ready for a major release :+1: 

Hope that makes sense.

Thanks!
Bruno

[^1]: https://www.oracle.com/java/technologies/compatibility.html#binary
[^2]: https://en.wikipedia.org/wiki/Binary-code_compatibility
[^3]: https://www.oracle.com/java/technologies/compatibility.html#behavioral
[^4]: https://docs.microsoft.com/en-us/dotnet/core/compatibility/categories#behavioral-change

garydgregory:
-1 this breaks binary compatibility. There is no need to do that at this time.

We can consider dropping all deprecations in 3.0, and there are no plans for that yet.

wodencafe:
Ah thank you for providing links and explaining @kinow and @garydgregory , I will put the methods back and update the pull request accordingly.

kinow:
Brilliant @wodencafe ! I haven't looked at CI, but it appears to be failing. We need to confirm if it is something with this PR or if it is failing on master too.

garydgregory:
Binary compatibility is still broken. See the builds referred to from this PR.

garydgregory:
Let's close this PR unless there are changes that won't break BC.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -280,6 +280,8 @@ public static BufferedOutputStream buffer(final OutputStream outputStream, final
      * @since 2.5
      */
     public static BufferedReader buffer(final Reader reader) {
+        // reject null early on rather than waiting for IO operation to fail
+        Objects.requireNonNull(reader, "reader");
```
garydgregory, line 284: This comment is wrong and the check is not needed: The BufferedReader constructor code path already checks for null input and throws an NPE.

data443-cboyd:
> Let's close this PR unless there are changes that won't break BC.

Hi @garydgregory and @kinow , I have an updated version of this branch locally that only changes the Javadocs to appropriately match the contract / functionality of the IOUtils methods, and also some misc. formatting / fixes to the existing Javadoc itself. I should be able to push the changes tonight / tomorrow.

Thank you for the review and feedback.

wodencafe:
> Let's close this PR unless there are changes that won't break BC.

Hi @garydgregory and @kinow , I have an updated version of this branch locally that only changes the Javadocs to appropriately match the contract / functionality of the IOUtils methods, and also some misc. formatting / fixes to the existing Javadoc itself. I should be able to push the changes tonight / tomorrow.

Thank you for the review and feedback.

wodencafe:
Hello again @kinow and @garydgregory , I have updated this Pull Request to tidy up the javadoc in IOUtils.

garydgregory:
Lots of changes here, TY. I'll try to go through them all this weekend.

garydgregory:
HI @wodencafe 
It's quite hard to review such a giant changeset that claims to only be about Javadoc but also changes the code in `contentEqualsIgnoreEOL`, so let's drop code changes, then I can just review Javadoc. TY.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -271,109 +276,115 @@ public static BufferedOutputStream buffer(final OutputStream outputStream, final
...
+     * @return a new {@code byte[]} array of the given size.
      * @since 2.9.0
      */
     public static byte[] byteArray(final int size) {
         return new byte[size];
     }
 
     /**
-     * Returns a new char array of size {@link #DEFAULT_BUFFER_SIZE}.
+     * Returns a new {@code char[]} array of size {@link #DEFAULT_BUFFER_SIZE}.
```
garydgregory, line 360: This is no good IMO: "a a new char array" is clear and correct. "a new {@code char} array" would also be correct. BUT "a "new {@code char[]} array" reads like an array of array. So let's pass on this.

garydgregory:
Closing, no feedback.

PR #300, opened 2021-11-05
IOUtils chain together IOException for Closeables, add overload

wodencafe:
This pull request chains together IOExceptions thrown from closing multiple Closeable instances, which allows the method to attempt to close all of the Closeable instances rather than halting at the first IOException. An overload is added to allow closing multiple Closeable instances and passing the possible resulting IOException to an IOConsumer<IOException>.


garydgregory:
-1: This makes no sense to me as the exception are not causaly unrelated and you are forcing them to be. This might be a use case for IOExceptionList.

wodencafe:
Hi @garydgregory , good suggestion, IOExceptionList seems like the perfect fit for this. I have revised the code to use this class.

garydgregory:
I ended up reworking the underlying so that now the method looks like this
```
    public static void close(final Closeable... closeables) throws IOException {
        IOConsumer.forEach(closeables, IOUtils::close);
    }
```
See also `forEachIndex` which provides slightly different information for compatibility with other call sites.

garydgregory:
Hi @wodencafe Thank you for your update. Now I see one new method, without any tests, and not used from anywhere.

- Should the [] be changed to a ... and moved to the last parameter position? I would think so.
- Can the method be used within the code base?
- There are no tests.

wodencafe:
Hello @garydgregory , my apologies, I got swamped this weekend and haven't updated this PR with tests. I also agree that the [] should be changed to ... varargs and swapped to be the last parameter, so I will implement this change, and see if there are other places in Commons IO that can take advantage of this new functionality. I will add tests for this today.

Thank you for reviewing the changes, and the contributions and feedback.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -414,6 +433,48 @@ public static void close(final Closeable closeable, final IOConsumer<IOException
...
+    public static <T extends Closeable> void close(final IOConsumer<IOException> consumer, final Stream<T> closeables) throws IOException {
+        if (closeables != null) {
+            try {
+                close(closeables);
+            } catch (final IOException e) {
+                if (consumer != null) {
+                    consumer.accept(e);
+                }
+            } finally {
+                closeables.close();
```
garydgregory, line 454: This seems wrong to me since this method did not create the stream. The caller should manage the stream IMO, it does not matter that a stream also happens to be closeable.

src/main/java/org/apache/commons/io/function/IOConsumer.java
```diff
@@ -85,13 +150,25 @@
...
+    /**
+     * Returns this {@code IOConsumer} wrapped inside of a {@link Consumer}
+     * that throws {@link UncheckedIOException} for any {@link IOException}s
+     * that are thrown by this {@code IOConsumer}.
+     *
+     * @return a {@code Consumer} that wraps this {@code IOConsumer}.
+     * @since 2.12.0
+     */
+    default Consumer<T> asConsumer() {
+        return wrap(this);
```
garydgregory, line 162: I do not think we need an extra method that just delegates to an existing one. What is the added value? Am I missing something here?

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -394,6 +394,25 @@ public static void close(final Closeable... closeables) throws IOException {
...
+     * @param closeables The resource(s) to close, may be null.
+     * @throws IOExceptionList if an I/O error occurs.
+     * @since 2.12.0
+     */
+    public static <T extends Closeable> void close(final Stream<T> closeables) throws IOExceptionList {
+        if (closeables != null) {
+            try {
+                IOConsumer.forEachIndexed(closeables, IOUtils::close);
+            } finally {
+                closeables.close();
```
garydgregory, line 411: This method should not release a resource it did not acquire IMO.

garydgregory (changes requested):

src/main/java/org/apache/commons/io/function/IOConsumer.java
data443-cboyd: Hi @garydgregory , my thought process at the time was that the user may want to  override this method with their own implementation of IOConsumer, but after considering your comments, it would be just as easy for the user who requires a custom implementation to provide that themselves externally, and otherwise rely on the static method for wrapping. I will remove this method.

wodencafe: Hi @garydgregory , my thought process at the time was that the user may want to override this method with their own implementation of IOConsumer, but after considering your comments, it would be just as easy for the user who requires a custom implementation to provide that themselves externally, and otherwise rely on the static method for wrapping. I will remove this method.

src/main/java/org/apache/commons/io/IOUtils.java
wodencafe: Agreed, closing the stream here is making an assumption that this should be a terminal operation on the stream, which is a limitation on its usefulness. The caller can close the stream on their own when they are finished with it. I will update the code.

src/main/java/org/apache/commons/io/IOUtils.java
wodencafe: Agreed, closing the stream here is making an assumption that this should be a terminal operation on the stream, which is a limitation on its usefulness. The caller can close the stream on their own when they are finished with it. I will update the code.

wodencafe:
Hi @garydgregory , I have updated the pull request with the changes you suggested.
Thank you!

src/main/java/org/apache/commons/io/IOUtils.java
jochenw: Please, do. And also, add a single sentence to the Javadocs, that mentions the fact. (Always good, if the caller is made aware of his/her ressponsibilities.)

PR #301, opened 2021-11-11
IOUtils use TaggedInputStream instead of ProxyInputStream

wodencafe:
This pull request is to use `TaggedInputStream` as a superclass where possible, in place of `ProxyInputStream`, to allow access to the methods `isCauseOf` and `throwIfCauseOf`.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/AutoCloseInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/AutoCloseInputStream.java
@@ -36,7 +36,7 @@ import java.io.InputStream;
  *
  * @since 1.4
  */
-public class AutoCloseInputStream extends ProxyInputStream {
+public class AutoCloseInputStream extends TaggedInputStream {
 
     /**
      * Creates an automatically closing proxy for the given input stream.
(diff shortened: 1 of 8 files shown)


src/main/java/org/apache/commons/io/input/AutoCloseInputStream.java
```diff
@@ -76,7 +76,7 @@ protected void afterRead(final int n) throws IOException {
      */
     @Override
     public void close() throws IOException {
-        in.close();
+        super.close();
```
kinow, line 79: Right, had a look at the source code, and I think:

- before: `in.close()` (removed above) could throw any `IOException` directly
- after: `super.close()` will call `ProxyInputStream#close()` which will then call `TaggedInputStream#handleIOException` (instead of `ProxyInputStream#handleIOException`)

So all good here :+1:

kinow (commented):
@wodencafe thanks for the PR (another good one!). I think I undestood the change and motivation, but will need more time to look for test coverage, backward compatibility (I think it should be OK to use a class that's parent of `ProxyInputSteam`, probably binary compatible), or possible bugs/regressions. But from a quick look, couldn't find anything out of place.

But we will need a JIRA for this one @wodencafe , could you create/link one here, please?

Thanks!
Bruno

garydgregory:
Closing, no reply since November of last year.

PR #303, opened 2021-11-13
BoundedInputStream extend ProxyInputStream

wodencafe:
Hello, this change is to have BoundedInputStream extend ProxyInputStream, to take advantage of the parent class ProxyInputStream for managing the delegate. This change is along the lines of https://github.com/apache/commons-io/pull/296


garydgregory:
I don't understand why you created another PR in the same style as #301, again without tests, without waiting to see how we resolve #301... let's take this one step at a time, otherwise we are all spending time spinning our wheels.

garydgregory:
Closing, no reply since November of last year in #301.

PR #310, opened 2021-12-07
(doc) [IO-484] Fix incorrect FilenameUtils documentation for null bytes

Marcono1234:
During the development of [IO-484](https://issues.apache.org/jira/browse/IO-484) the behavior was first to remove null bytes (afe78a030b57e3f74825e134d6615ef4878778cc) but then later the implementation was changed to throw an exception instead (5d072ef89fbc2532f621a5a0b4d6791cb926a997).
However, not the complete documentation was adjusted.

This pull request corrects the documentation by mentioning that an `IllegalArgumentException` will be thrown for null bytes.

However, there are cases where the null bytes check is missing, or might be undesired. I have added TODO comments at these locations. Please let me know if you want any additional changes there.

### DIFF
--- a/src/main/java/org/apache/commons/io/FilenameUtils.java
+++ b/src/main/java/org/apache/commons/io/FilenameUtils.java
@@ -194,7 +194,8 @@ public class FilenameUtils {
      *
      * @param basePath  the base path to attach to, always treated as a path
      * @param fullFileNameToAdd  the fileName (or path) to attach to the base
-     * @return the concatenated path, or null if invalid.  Null bytes inside string will be removed
+     * @return the concatenated path, or null if invalid
+     * @throws IllegalArgumentException if the result path contains null bytes
      */
     public static String concat(final String basePath, final String fullFileNameToAdd) {
         final int prefix = getPrefixLength(fullFileNameToAdd);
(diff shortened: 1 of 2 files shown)


garydgregory:
Hi @Marcono1234 
Thank you for your patience. The new docs are not quite right because we are not checking for null _bytes_, but we are checking for `\u0000` which is bigger than a `byte`, it's a `char`.

garydgregory:
@Marcono1234 ping

Marcono1234:
You are right, I have used the term "null character" now, which is also used by Wikipedia and in some of the documentation comments of the JDK classes.
Is that what you had in mind?

Also, what should I do regarding the TODO comments I added, should I just leave them in?

garydgregory:
@Marcono1234 
Thank you for the update. Yes, much better now. I think we can take out the TODO comments for now and make this a Javadoc only PR. We can revisit error check later as this will require more tests to either validate the current behavior or change it. TY! 😀

Marcono1234:
Ok, I have removed the comments now. Here are the issues the TODO comments were about, in case you or someone else wants to have a look at them in the future:

- `doGetFullPath(String, boolean)`: In some cases does not call (indirectly) `requireNonNullChars`
- `equals(String, String, boolean, IOCase)`: Normalization can throw `IllegalArgumentException` when file names contain null characters. Currently not documented, but not clear what the desired behavior is. Should exception be caught and `false` be returned? If exception should not be caught, then exception has to be documented, also for callers of this method. And should `fileName2` then always be validated as well? Currently if normalization is requested but `fileName1` is `null`, `fileName2` is not checked for null characters.
- `getExtension(String)`: Does not call `requireNonNullChars`

PR #322, opened 2022-01-30
Avoid use deprecated read basic file attributes method.

arturobernalg:


garydgregory:
Why? You'd never do this with a vararg method, especially once the deprecated method is deleted in the next major version. Plus, if we do this now, then we'd want to change it again in 3.0 IMO.

arturobernalg:
> Why? You'd never do this with a vararg method, especially once the deprecated method is deleted in the next major version. Plus, if we do this now, then we'd want to change it again in 3.0 IMO.

sorry @garydgregory, don't fallow you. What we should never do?
TY

garydgregory:
The class `PathUtils` defines `readBasicFileAttributes(Path)` and `readBasicFileAttributes(Path. LinkOption...)`, the former being deprecated. When `readBasicFileAttributes(Path)` is removed in 3.0, sources will compile automatically against `readBasicFileAttributes(Path. LinkOption...)`.

PR #325, opened 2022-02-05
IOCase.isCaseSensitive(IOCase) result is backward

code-ferry:
When the argument is IOCase.SENSITIVE, I think the static method isCaseSensitive should return true.

I did the following test with the code.

```
public void testIsCaseSensitive() {
  // return true. I think it is right.
  boolean b1 = IOCase.SENSITIVE.isCaseSensitive();
  System.out.println(b1);

  // but invoke static method and return false.
  boolean b2 = IOCase.isCaseSensitive(IOCase.SENSITIVE);
  System.out.println(b2);
}
```

### DIFF
--- a/src/main/java/org/apache/commons/io/filefilter/RegexFileFilter.java
+++ b/src/main/java/org/apache/commons/io/filefilter/RegexFileFilter.java
@@ -16,6 +16,8 @@
  */
 package org.apache.commons.io.filefilter;
 
+import org.apache.commons.io.IOCase;
+
 import java.io.File;
 import java.io.Serializable;
 import java.nio.file.FileVisitResult;
@@ -23,8 +25,6 @@ import java.nio.file.Path;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.function.Function;
 import java.util.regex.Pattern;
-
-import org.apache.commons.io.IOCase;
 
 /**
  * Filters files using supplied regular expression(s).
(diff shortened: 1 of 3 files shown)


garydgregory:
Hello @code-ferry 
This PR needs a test that fails when the main side of the changes are not applied. Otherwise, if the main code changes back in the future, no one will know it breaks behavior.  You will need to test all values.

XenoAmess:
> Hello @code-ferry This PR needs a test that fails when the main side of the changes are not applied. Otherwise, if the main code changes back in the future, no one will know it breaks behavior. You will need to test all values.

IMO the test should not be added now, but should be added when the original codes written.
If the codes is added into the lib with test, I don't think mistake in this level should happened.
(however we cannot change what happened already; adding the tests now is a better choice than doing nothing.)

src/main/java/org/apache/commons/io/filefilter/RegexFileFilter.java
```diff
@@ -16,6 +16,8 @@
  */
 package org.apache.commons.io.filefilter;
 
+import org.apache.commons.io.IOCase;
```
XenoAmess, line 19: according to minimal change, this import's position change is unnecessary to this issue, suggest revert these 4 lines changes.

src/main/java/org/apache/commons/io/filefilter/RegexFileFilter.java
```diff
@@ -24,8 +26,6 @@
 import java.util.function.Function;
 import java.util.regex.Pattern;
 
-import org.apache.commons.io.IOCase;
```
XenoAmess, line 27: and this too.

code-ferry:
> Hello @code-ferry This PR needs a test that fails when the main side of the changes are not applied. Otherwise, if the main code changes back in the future, no one will know it breaks behavior. You will need to test all values.

The chanage in method toFlags has the test in class RegexFileFilterTest.
I add a test of static method isCaseSensitive in class IOCaseTest.

PR #327, opened 2022-02-15
Fix incollect value in field currentBufferIndex when creating new buffer first time.

code-ferry:
The currentBufferIndex means the index of the current buffer in the class AbstractByteArrayOutputStream. 
when creating new buffer first time and the write size , currentBufferIndex should not change.

Luckily, this mistable does not affect any public methods and the result. After my modification, all the unit tests also pass.

```
    @Test
    public void testCurrentBufferIndex() throws IOException, IllegalAccessException, InstantiationException, NoSuchFieldException {
        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
        outputStream.write(new byte[]{0x31, 0x32});

        // Get Field currentBufferIndex
        Field field = ByteArrayOutputStream.class.getSuperclass().getDeclaredField("currentBufferIndex");
        field.setAccessible(true);
        int currentBufferIndex = (int) field.get(outputStream);

        // print 1. I think the right value is 0.
        //
        // Only 2 bytes are written, less than the default value 1024.
        // The size of array buffers is only 1, and currentBufferIndex should be 0.
        System.out.println(currentBufferIndex);
    }
```

### DIFF
--- a/src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
+++ b/src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
@@ -357,10 +357,10 @@ public class ByteArrayOutputStreamTest {
     @ValueSource(ints = {1024, 64})
     public void testPrivateFieldOfCurrentBufferIndex(int size) throws NoSuchFieldException, IllegalAccessException {
         // Use the default size and custom size.
-        ByteArrayOutputStream os = new ByteArrayOutputStream(size);
+        final ByteArrayOutputStream os = new ByteArrayOutputStream(size);
 
         // Get private field object.
-        Field field = AbstractByteArrayOutputStream.class.getDeclaredField("currentBufferIndex");
+        final Field field = AbstractByteArrayOutputStream.class.getDeclaredField("currentBufferIndex");
         field.setAccessible(true);
 
         // Data for writing
--- a/src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
+++ b/src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
@@ -26,6 +26,8 @@ import static org.junit.jupiter.api.Assertions.fail;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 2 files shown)


garydgregory:
Hi @code-ferry 
Thank you for your PR. This change needs a test that fails without the change to the main tree; otherwise, a future change can regress this change.

code-ferry:
@garydgregory 
I had privided a test that fails without the change to the main tree. 
But some checks were not successful because the unit test of FileUtilsWaitForTest.testWaitForInterrupted may fail randomly. It's not my problem.

src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
```diff
@@ -350,6 +353,31 @@ public void testWriteZero(final String baosName, final BAOSFactory<?> baosFactor
...
     }
 
+    @ParameterizedTest(name = "[{index}] {0}")
+    @ValueSource(ints = {1024, 64})
+    public void testPrivateFieldOfCurrentBufferIndex(int size) throws NoSuchFieldException, IllegalAccessException {
+        // Use the default size and custom size.
+        final ByteArrayOutputStream os = new ByteArrayOutputStream(size);
+
+        // Get private field object.
+        final Field field = AbstractByteArrayOutputStream.class.getDeclaredField("currentBufferIndex");
```
garydgregory, line 363: @code-ferry 
Thank you for your update .
You'll want to use Mockito for this kind of testing, it is already a dependency.

garydgregory: Actually it would better to also test the contents of the stream maybe by converting it to a string. So a black box test instead of a white box test. This would insure that the actual behavior is correct, not just the internal mechanism.

code-ferry: @garydgregory
In general, unit tests should be white box tests. But I puzzled with this PR. Change the position of  the variable currentBufferIndex have no effect on the results of public methods, such as write and reset, so I used black box tests to illustrate the problem.
In other words, this PR is not necessary. I will close this issue.

PR #339, opened 2022-03-27
Add missing javadoc for exceptions thrown for invalid arguments

dvmarcilio:
Hello! I added some missing javadoc for `NullPointerException` and `IllegalArgumentException` in some Util classes.
I added test assertions only when tests already existed for a method.

For `PathUtils.filter(PathFilter, Path... )` the doc was actually inconsistent: mentioning IAE instead of NPE.

Thanks for taking the time to review.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -1181,7 +1181,8 @@ public class FileUtils {
      *
      * @param file The file to delete.
      * @return the given file.
-     * @throws IOException if the file cannot be deleted.
+     * @throws NullPointerException     if the parameter is {@code null}
+     * @throws IOException              if the file cannot be deleted.
      * @see File#delete()
      * @since 2.9.0
      */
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -3119,8 +3121,9 @@ private static int toMaxDepth(final boolean recursive) {
      * </p>
      *
      * @param file the File to touch.
-     * @throws IOException if an I/O problem occurs.
-     * @throws IOException if setting the last-modified time failed.
+     * @throws NullPointerException if the parameter is {@code null}
+     * @throws IOException          if an I/O problem occurs.
+     * @throws IOException          if setting the last-modified time failed.
```
garydgregory, line 3126: Listing each exception type once is enough IMO.

dvmarcilio: I force pushed a new version merging the two descriptions for `IOException` in the same line. Does that work?

PR #347, opened 2022-04-15
Add @SuppressWarnings("unchecked").

arturobernalg:


garydgregory:
Hi @arturobernalg 
Thank you for the PR. I'm not 100% sure how we should handle these warnings because these can indeed cause a CCE. So maybve we should at least document this fact.

PR #349, opened 2022-04-23
Configure Maven Fork Test parameters

CAIWan1998:
Maven will run all tests in a single forked VM by default. This can be problematic if there are a lot of tests or some very memory-hungry ones. We can fork more test VM by setting `<fork>1.5C</fork>`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.


garydgregory:
I don't see a need for this, all builds are already green on Windows, macOS, and Linux. We want consistency and 1 VM gives us that.

kinow (commented):
I agree with @garydgregory . I'd merge this one if our builds took a long time to run the tests and thia change showed some good perf improvement. Which I am not sure if it is a problem now.

PR #350, opened 2022-04-25
Add FileWatcher feature

saiple:
Hello everyone.
I want to suggest to add some features to creating file system watchers very easy.

You only need to write one line a code:
FileWatcher.createWatcher(path, watchEvent -> System.out.println(watchEvent.kind().name() + " " + watchEvent.context()));
Put here path, which should be watched, watch properties (optional), and supplier, if this event would triggered.

Be free to review and comment. Thanks.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileWatcher.java
+++ b/src/main/java/org/apache/commons/io/FileWatcher.java
@@ -1,0 +1,90 @@
+package org.apache.commons.io;
+
+import org.apache.commons.io.watcher.WatcherException;
+import org.apache.commons.io.watcher.WatcherProperties;
+
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.WatchEvent;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+public class FileWatcher {
+
+    private static List<WatchThread> WATCH_THREADS = new ArrayList<>();
+
+    private FileWatcher() {
+    }
+
+    public static void createWatcher(Path path,
+                                         Consumer<WatchEvent<?>> eventConsumer) {
+        createWatcher(path, new WatcherProperties(), eventConsumer);
+    }
+
+    public static void createWatcher(Path path,
+                                         WatcherProperties watcherProperties,
+                                         Consumer<WatchEvent<?>> eventConsumer) {
+        WatchThread watchThread = new WatchThread(path, watcherProperties, eventConsumer);;
+        watchThread.start();
+    }
+
+    public static List<WatchThread> getWatchThreads() {
+        return new ArrayList<>(WATCH_THREADS);
+    }
+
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/commons/io/FileWatcher.java
```diff
@@ -0,0 +1,90 @@
+package org.apache.commons.io;
```
kinow, line 1: Missing apache license is causing most builds to fail due to RAT test failure.

src/main/java/org/apache/commons/io/watcher/WatcherProperties.java
```diff
@@ -0,0 +1,31 @@
+package org.apache.commons.io.watcher;
```
kinow, line 1: License header.

src/main/java/org/apache/commons/io/watcher/WatcherException.java
```diff
@@ -0,0 +1,7 @@
+package org.apache.commons.io.watcher;
```
kinow, line 1: License header.

kinow (commented):
I think adding the license headers will fix the RAT failure in GH Actions. It will also need a JIRA for this new feature. Thanks @saiple !

garydgregory:
Hm, Commons VFS already does stuff like this, for example https://commons.apache.org/proper/commons-vfs/commons-vfs2/apidocs/org/apache/commons/vfs2/impl/DefaultFileMonitor.html

src/main/java/org/apache/commons/io/watcher/WatcherProperties.java
saiple: added

src/main/java/org/apache/commons/io/watcher/WatcherException.java
saiple: added

saiple:
> Hm, Commons VFS already does stuff like this, for example https://commons.apache.org/proper/commons-vfs/commons-vfs2/apidocs/org/apache/commons/vfs2/impl/DefaultFileMonitor.html

Hi @garydgregory ! Sorry for long delay. Yea, I know that Commons VFS this feature, but I wanted creating watchers by calling one line of code, so most of methods of this lib (Commons IO) have this ideology. In VFS it need to write at less 5 lines of code for reaching it, it makes code less cleaner. I use this lib in my private projects, and I wanted to use this method natively from lib.

garydgregory:
Hello @saiple,
-1.
I am am sorry but this code seems flawed and not in the right home. It seems too easy to break an app on exit or at time really. 

It's hard to tell of course since this PR has zero unit tests so use cases are imaginary at this point. It seems that little or no thought was given to concurrency issues, for example, the use of a concurrent unsafe list, use of non-daemon threads, threads updating a global variable.

IMO, the "lines of code" argument is not relevant since you can abstract these in a method in your app. I am not a fan of duplicating this kind of functionality across IO and VFS. In my mind, IO is a low level library and VFS is a higher level library where I see this type of functionality residing. I am curious to learn what others think.

What is also problematic IMO is the use of a global variable to track threads as well as the other issues mentioned. So -1.

PR #351, opened 2022-04-26
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -17,6 +17,9 @@ name: Java CI
 
 on: [push, pull_request]
 
+permissions:
+  contents: read
+
 jobs:
   build:
 


kinow (approved):

kinow:
Thanks!

garydgregory:
@kinow are you doing this for all Commons components?

kinow:
> @kinow are you doing this for all Commons components?

Sounds like a good idea I think. I'll start with Commons Imaging/Text/IO/etc, the ones I normally change. Then slowly go over others (feel free to modify any of the other components if you'd like :+1: )

garydgregory:
Is there a CVE reference for this?

kinow:
Not sure if this generated a CVE. It is an OpenSSF recommendation, however: https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

The closest CVE I could find was this one, @garydgregory (and I know that GH action, I'd seen it in use in another OSS project :worried: ): https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-32724

garydgregory:
Thanks for the pointers @kinow 🔍👍

aherbert:
I am trying to find out if this is a security issue on our repos.

The GH actions run using the permissions granted to the  `GITHUB_TOKEN`. See [Permissions for the GITHUB_TOKEN](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token).

The default permissions are set at the enterprise, organization, or repository level. The question is how do we find out if Apache has set this to `permissive` or `restricted` across all the Commons repos? It requires admin permissions on the Github repo so may have to be done by INFRA. If it is set as `restricted` then this is a non-issue as the defaults are read on contents and metadata and nothing else.

This is only an issue if the default access is set to `permissive` across our repos.

Note that it does not do any harm to set it for all the workflows as this is the mechanism used to change the permissions from the default. Explicitly setting it in the workflow yaml makes it clear that the intention is a read-only build of the repo contents.

PR #358, opened 2022-05-19
Add test for AbstractByteArrayOutputStream.resetImpl and .toByteArrayImpl

lacinoire:
Hey 😊
I want to contribute the following test:

Test that `baout.toString()` is equal to `""` when `reset` is called. 
This tests the methods [`AbstractByteArrayOutputStream.toByteArrayImpl`](https://github.com/apache/commons-io/blob/4aab769c3279ea63b7bcdaa163c50761bc540f8c/src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java#L188) and [`AbstractByteArrayOutputStream.resetImpl`](https://github.com/apache/commons-io/blob/4aab769c3279ea63b7bcdaa163c50761bc540f8c/src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java#L148). 
This test is based on the test [`copy_inputStreamToWriter`](https://github.com/apache/commons-io/blob/4aab769c3279ea63b7bcdaa163c50761bc540f8c/src/test/java/org/apache/commons/io/CopyUtilsTest.java#L83).

Curious to hear what you think!

(I wrote this test as part of a research study at TU Delft. [Find out more](https://github.com/lacinoire/lacinoire/blob/main/README.md))

### DIFF
--- a/src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
+++ b/src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
@@ -26,7 +26,6 @@ import static org.junit.jupiter.api.Assertions.fail;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.OutputStream;
 import java.io.Writer;
 import java.io.OutputStreamWriter;
 import java.nio.charset.StandardCharsets;
@@ -37,7 +36,6 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.io.function.IOFunction;
 import org.apache.commons.io.input.ClosedInputStream;
 import org.apache.commons.io.test.TestUtils;
-import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
@@ -360,8 +358,8 @@ public class ByteArrayOutputStreamTest {
 
     @Test
     public void testToByteArrayImplAndResetImpl() throws Exception {
-        final int file_size = (1024 * 4) + 1;
-        final byte[] inData = TestUtils.generateTestData(file_size);
+        final int fileSize = 4097;
+        final byte[] inData = TestUtils.generateTestData(fileSize);
         try (InputStream in = new ByteArrayInputStream(inData)) {
             try (ByteArrayOutputStream baout = new ByteArrayOutputStream()) {
                 try (Writer writer = new OutputStreamWriter(baout, StandardCharsets.US_ASCII)) {
@@ -369,7 +367,7 @@ public class ByteArrayOutputStreamTest {
                     writer.flush();
                 }
                 baout.reset();
-                Assertions.assertEquals("", baout.toString());
+                assertEquals("", baout.toString());
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
```diff
@@ -350,6 +360,22 @@ public void testWriteZero(final String baosName, final BAOSFactory<?> baosFactor
         }
     }
 
+    private static final int FILE_SIZE = (1024 * 4) + 1;
+    private final byte[] inData = TestUtils.generateTestData(FILE_SIZE);
```
garydgregory, line 364: These data do not need to be tracked for all tests. Move them inside the one method that use them.

src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
```diff
@@ -350,6 +360,22 @@ public void testWriteZero(final String baosName, final BAOSFactory<?> baosFactor
...
+
+    @Test
+    public void testToByteArrayImplAndResetImpl() throws Exception {
+        InputStream in = new ByteArrayInputStream(inData);
```
garydgregory, line 368: Use try with resources

garydgregory (changes requested):
Thank you for your PR, see comments.

src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
```diff
@@ -350,6 +358,22 @@ public void testWriteZero(final String baosName, final BAOSFactory<?> baosFactor
         }
     }
 
+    @Test
+    public void testToByteArrayImplAndResetImpl() throws Exception {
+        final int file_size = (1024 * 4) + 1;
```
garydgregory, line 363: - No underscore in local variable names. 
- No need for spelling out 4097 in an expression.

src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
```diff
@@ -350,6 +358,22 @@ public void testWriteZero(final String baosName, final BAOSFactory<?> baosFactor
...
+        final byte[] inData = TestUtils.generateTestData(file_size);
+        try (InputStream in = new ByteArrayInputStream(inData)) {
+            try (ByteArrayOutputStream baout = new ByteArrayOutputStream()) {
+                try (Writer writer = new OutputStreamWriter(baout, StandardCharsets.US_ASCII)) {
+                    CopyUtils.copy(in, writer);
+                    writer.flush();
+                }
+                baout.reset();
+                Assertions.assertEquals("", baout.toString());
```
garydgregory, line 372: No need for Assertions prefix static in a test method. Use the other methods in this file as a style guide.

garydgregory (changes requested):
@lacinoire 
Thank you for your update. Please see my scattered comments. Run a local default build (Run: `mvn`) before you push your PR as the build fails (see checkstyle error in output).

lacinoire:
Thank you for the second round of comments 🙂
I addressed them and also made sure that `mvn checkstyle:check`  and `mvn verify clean` pass.
Hope the CI does to 😄

garydgregory:
That's not enough, you want to run "mvn" which will run the default goal which runs more checks.

lacinoire:
> That's not enough, you want to run "mvn" which will run the default goal which runs more checks.

Also passes ✅ 

Curious if the coverage.yml run shows any changes.

lacinoire:
Hey @garydgregory,
looked into those two tests in detail and indeed, as far as I can see they should also cover most the instructions I was targeting. Also they should covere whole new branches which i clearly not the case, trusting covdecov here.
I will go back and check why our own coverage calculation is so off in this case.

Thank you for the in-depth feedback and help with this PR! :pray: 🙂

PR #374, opened 2022-08-24
Better docs in IOUtils and IOUtils.byteArray(int size)

ArdenL-Liu:
IOUtils.byteArray(int size) add the verification to assure that the size is legal(size > 0), the illegal(size <=0) should throw IllegalArgumentException.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -3537,9 +3537,7 @@ public class FileUtils {
 
     /**
      * Instances should NOT be constructed in standard programming.
-     * @deprecated Will be private in 3.0.
      */
-    @Deprecated
     public FileUtils() { //NOSONAR
 
     }
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -3537,6 +3537,7 @@ public class FileUtils {
 
     /**
      * Instances should NOT be constructed in standard programming.
+     * @deprecated Will be private in 3.0.
      */
     @Deprecated
     public FileUtils() { //NOSONAR
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -3538,6 +3538,7 @@ public class FileUtils {
     /**
      * Instances should NOT be constructed in standard programming.
      */
+    @Deprecated
     public FileUtils() { //NOSONAR
 
     }
--- a/src/main/java/org/apache/commons/io/IOUtils.java
+++ b/src/main/java/org/apache/commons/io/IOUtils.java
@@ -346,9 +346,15 @@ public class IOUtils {
      *
      * @param size array size.
      * @return a new byte array of the given size.
+     *
+     * @exception  IllegalArgumentException  If {@code size <= 0}
+     *
      * @since 2.9.0
      */
     public static byte[] byteArray(final int size) {
+        if (size <= 0) {
+            throw new IllegalArgumentException("byte size <= 0");
+        }
         return new byte[size];
     }
 
@@ -479,7 +485,7 @@ public class IOUtils {
      * @param closeable the objects to close, may be null or already closed
      * @since 2.0
      *
-     * @see Throwable#addSuppressed(java.lang.Throwable)
+     * @see Throwable#addSuppressed(Throwable)
      */
     public static void closeQuietly(final Closeable closeable) {
         closeQuietly(closeable, null);
@@ -529,7 +535,7 @@ public class IOUtils {
      * @param closeables the objects to close, may be null or already closed
      * @see #closeQuietly(Closeable)
      * @since 2.5
-     * @see Throwable#addSuppressed(java.lang.Throwable)
+     * @see Throwable#addSuppressed(Throwable)
      */
     public static void closeQuietly(final Closeable... closeables) {
         if (closeables != null) {
@@ -583,7 +589,7 @@ public class IOUtils {
      * </p>
      *
      * @param input the InputStream to close, may be null or already closed
-     * @see Throwable#addSuppressed(java.lang.Throwable)
+     * @see Throwable#addSuppressed(Throwable)
      */
     public static void closeQuietly(final InputStream input) {
         closeQ(input);
@@ -633,7 +639,7 @@ public class IOUtils {
      * </p>
      *
      * @param output the OutputStream to close, may be null or already closed
-     * @see Throwable#addSuppressed(java.lang.Throwable)
+     * @see Throwable#addSuppressed(Throwable)
      */
     public static void closeQuietly(final OutputStream output) {
         closeQ(output);
@@ -666,7 +672,7 @@ public class IOUtils {
      * </p>
      *
      * @param reader the Reader to close, may be null or already closed
-     * @see Throwable#addSuppressed(java.lang.Throwable)
+     * @see Throwable#addSuppressed(Throwable)
      */
     public static void closeQuietly(final Reader reader) {
         closeQ(reader);
@@ -699,7 +705,7 @@ public class IOUtils {
      *
      * @param selector the Selector to close, may be null or already closed
      * @since 2.2
-     * @see Throwable#addSuppressed(java.lang.Throwable)
+     * @see Throwable#addSuppressed(Throwable)
      */
     public static void closeQuietly(final Selector selector) {
         closeQ(selector);
@@ -732,7 +738,7 @@ public class IOUtils {
      *
      * @param serverSocket the ServerSocket to close, may be null or already closed
      * @since 2.2
-     * @see Throwable#addSuppressed(java.lang.Throwable)
+     * @see Throwable#addSuppressed(Throwable)
      */
     public static void closeQuietly(final ServerSocket serverSocket) {
         closeQ(serverSocket);
@@ -765,7 +771,7 @@ public class IOUtils {
      *
      * @param socket the Socket to close, may be null or already closed
      * @since 2.0
-     * @see Throwable#addSuppressed(java.lang.Throwable)
+     * @see Throwable#addSuppressed(Throwable)
      */
     public static void closeQuietly(final Socket socket) {
         closeQ(socket);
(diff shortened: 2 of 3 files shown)


src/test/java/org/apache/commons/io/IOUtilsTest.java
```diff
@@ -1716,4 +1716,18 @@ public void testWriteLittleString() throws IOException {
...
+    public void testByteArrayWithIllegalSize() {
+        try {
+            int size = -1;
+            byte[] bytes = IOUtils.byteArray(size);
+
+            size = 0;
+            bytes = IOUtils.byteArray(size);
+        }catch (Exception e) {
+            assertEquals(e.getClass().getName(), IllegalArgumentException.class.getName());
+        }
```
kinow, line 1729: I think instead it could use JUnit's `@Test(expected=IllegalArgumentException.class)`, or `assertThrows`.

kinow: And I think when the first exception happens, the second is part of the test is never executed? Probably needs to be broken down into two separate tests, maybe with `@ParameterizedTest` too.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -346,9 +346,15 @@ public static byte[] byteArray() {
      *
      * @param size array size.
      * @return a new byte array of the given size.
+     *
+     * @exception  IllegalArgumentException  If {@code size <= 0}
```
kinow, line 350: `@throws` instead of `@exception`

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -3724,7 +3724,9 @@ public static Writer writer(final Appendable appendable) {
 
     /**
      * Instances should NOT be constructed in standard programming.
+     * @deprecated Will be private in 3.0.
      */
+    @Deprecated
```
kinow, line 3729: This is doing more than what's described in the PR title. Maybe worth moving to a separate PR to make it easier to track in the changelog.

kinow (changes requested):

src/main/java/org/apache/commons/io/IOUtils.java
ArdenL-Liu: Thx, I will rollback it

src/main/java/org/apache/commons/io/IOUtils.java
ArdenL-Liu: Thx, update it with @throws

src/test/java/org/apache/commons/io/IOUtilsTest.java
ArdenL-Liu: ok, I changed IOUtilsTest.testByteArrayWithIllegalSize() with @ParameterizedTest.
you can review the code.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -3537,9 +3537,7 @@ public static void writeStringToFile(final File file, final String data, final S
 
     /**
      * Instances should NOT be constructed in standard programming.
-     * @deprecated Will be private in 3.0.
      */
-    @Deprecated
```
pjfanning, line 3542: Should this `@Deprecated` be removed? There was a comment to remove a different `@Deprecated` but this one should be kept.

src/test/java/org/apache/commons/io/IOUtilsTest.java
garydgregory: Why would we trade a more vague exception for the more precise and quite specific one already thrown by the VM? Maybe this issue can be simplified with better Javadoc?

src/main/java/org/apache/commons/io/FileUtils.java
ArdenL-Liu: rollback it

src/test/java/org/apache/commons/io/IOUtilsTest.java
ArdenL-Liu: Hi garydgregory,
    Thank you for your good suggestion, I updated the doc and changed the unittest, you can review it again.

src/main/java/org/apache/commons/io/FileUtils.java
garydgregory: -1: Don't remove this tag,

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -341,6 +341,7 @@ public static byte[] byteArray() {
 
     /**
      * Returns a new byte array of the given size.
+     * Throws java.lang.NegativeArraySizeException if the size is negative.
```
garydgregory, line 344: This is in the wrong place, use a throws tag.

src/test/java/org/apache/commons/io/IOUtilsTest.java
```diff
@@ -1716,4 +1718,12 @@ public void testWriteLittleString() throws IOException {
         }
     }
 
+    @Test
+    public void testByteArrayWithNegativeSize() {
+        int size = -1;
+        assertThrows(NegativeArraySizeException.class,() -> {
```
garydgregory, line 1724: You can just say `assertThrows(NegativeArraySizeException.class,() -> IOUtils.byteArray(-1));` but I am not even sure we should have this test since it does not improve our code coverage, it just test what the JVM does.

src/test/java/org/apache/commons/io/IOUtilsTest.java
```diff
@@ -79,6 +79,8 @@
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
```
garydgregory, line 82: -1: Unused

garydgregory (changes requested):
Hello @ArdenL-Liu 
Thank you for the PR, please see my comments.

src/test/java/org/apache/commons/io/IOUtilsTest.java
ArdenL-Liu: Hi garydgregory,
    from your comments, I checked the code and fixed again, you can review it again.
Thanks.

src/main/java/org/apache/commons/io/IOUtils.java
ArdenL-Liu: Hi garydgregory,
   I fixed it with @throws, you can review it again.

src/main/java/org/apache/commons/io/FileUtils.java
pjfanning: this is not resolved - you have still removed a comment that was in the code before - can you put it back?

src/main/java/org/apache/commons/io/IOUtils.java
pjfanning: Can you remove this deprecation as @kinow requested?

src/main/java/org/apache/commons/io/FileUtils.java
garydgregory: The `@deprecated` tag should NOT be removed.

ArdenL-Liu: Hi All,
     sorry for my error operation, I fixed the doc and add @deprecated tag about FileUtils.

src/test/java/org/apache/commons/io/IOUtilsTest.java
ArdenL-Liu: In order to improve my work efficiency，I will study github workflow in the next week.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -346,6 +346,7 @@ public static byte[] byteArray() {
      *
      * @param size array size.
      * @return a new byte array of the given size.
+     * @throws NegativeArraySizeException if the size is negative.
```
garydgregory, line 349: This PR seems ok to me now except for this addition. We do not document this type of JVM exception anywhere else IIRC. If we were to doc this, then it would make sense to doc all callers as well, then this would touch almost everything which would not help anyone IMO.

src/test/java/org/apache/commons/io/IOUtilsTest.java
```diff
@@ -16,40 +16,20 @@
  */
 package org.apache.commons.io;
 
-import static org.junit.jupiter.api.Assertions.assertArrayEquals;
```
garydgregory, line 19: Remove noise in the PR: There is no need to update these imports.

src/test/java/org/apache/commons/io/IOUtilsTest.java
```diff
@@ -60,25 +40,7 @@
...
-import org.apache.commons.io.output.NullWriter;
-import org.apache.commons.io.output.StringBuilderWriter;
-import org.apache.commons.io.test.TestUtils;
-import org.apache.commons.io.test.ThrowOnCloseReader;
-import org.apache.commons.lang3.StringUtils;
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Disabled;
-import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.io.TempDir;
+import static org.junit.jupiter.api.Assertions.*;
```
garydgregory, line 43: "*" imports are a no go

ArdenL-Liu: fixed

src/main/java/org/apache/commons/io/IOUtils.java
ArdenL-Liu: As yours, this is why I want to add throws NegativeArraySizeException tag to this method,  In fact, in order to help  users to use and understand these methods,  I think we should add it to the doc of  all methods like it.

garydgregory:
FYI: I've handled the Javadoc "java.lang" issue separately throughout most Apache Commons components.

garydgregory:
@ArdenL-Liu
Please **rebase** on git master and see any comments that may be outstanding.

ArdenL-Liu:
Hi garydgregory，
  thank you, I will rebase on git master.

ArdenL-Liu:
Hi garydgregory，
    You're great.  the comments updated in master line.

src/main/java/org/apache/commons/io/IOUtils.java
garydgregory: Wait... this deprecation is fine. It's slightly out of scope for the PR but it's fine.

PR #375, opened 2022-08-25
Fix typos

mernst:


RELEASE-NOTES.txt
```diff
@@ -341,7 +341,7 @@ Commons IO 1.4 requires Java 1.3.
 Historical list of changes: https://commons.apache.org/proper/commons-io/changes-report.html
 
 For complete information on Apache Commons IO, including instructions on how to submit bug reports,
-patches, or suggestions for improvement, see the Apache Apache Commons IO website:
```
kinow, line 344: These "Apache Apache" typos will be re-introduced in the next release.

The `RELEASE-NOTES.txt` file is auto-generated by the Commons Parent maven parent project. The template has "see the Apache ${project.name} website:". In the case of Commons IO, the project name in `pom.xml` is "Apache Commons IO", hence the "Apache Apache".

During the next release a release manager would fire the commands inherited from the Commons Parent module, and re-introduce all these fixes. I think we must either update the project name or update the template (can't tell which one is the right fix).

kinow (commented):
Left a comment about the fixes for `RELEASE-NOTES.txt`. Other than that, it looks really good! Thanks @mernst

RELEASE-NOTES.txt
mernst: Thanks for the explanation.  I have backed out the "Apache Apache" changes, so that the remainder of the pull request can be merged.

PR #377, opened 2022-08-29
[IO-769] FileUtils copyDirectory() should NOT use COPY_ATTRIBUTES 

tresf:
Since commons-io@2.9, the `copyDirectory()` function uses `COPY_ATTRIBUTES` to honor the `preserveFileDate` boolean.

This properly preserves the date, but introduces an inadvertent effect on file permissions.

For example, if copying from `%USERPROFILE%` to `C:\Program Files` the following permissions changes can be observed.

| Version           | Permissions (`icals <file>`)    |
|-------------------|---------------------------------|
| `commons-io@2.8`  | `NT AUTHORITY\SYSTEM:(I)(F)`<br>`BUILTIN\Administrators:(I)(F)`<br>`BUILTIN\Users:(I)(RX)`<br>`APPLICATION PACKAGE AUTHORITY\ALL APPLICATION PACKAGES:(I)(RX)`<br>`APPLICATION PACKAGE AUTHORITY\ALL RESTRICTED APPLICATION PACKAGES:(I)(RX)` |
| `commons-io@2.9` | `NT AUTHORITY\SYSTEM:(F)`<br>`BUILTIN\Administrators:(F)`<br>`WIN10ARM\owner:(F)`<br> |

* This regression is very similar for Unixes.  Prior to 2.9, files copied to a `root:root` (Linux) or `root:admin` (MacOS) owned directory would inherit the target permissions.  Starting with 2.9, the files will retain their original ownership.
* The OLD behavior is a closer match to the OS's default copy behavior (Linux, Mac and Windows will use the permissions of the target directory when copying).

The regression can be traced to to the 2.9 release and all releases thereafter.

Prior to submitting the PR, a conversation was had on the mailing list and stackoverflow.
* Mailing list convo: https://lists.apache.org/thread/schfj3m01ppd3287mxnn04lpp25hgw74
* Stackoverflow convo: https://stackoverflow.com/questions/73428286

| Version           | Status                          |
|-------------------|---------------------------------|
| `commons-io@1.1` (2005) | :white_check_mark: Inherits permissions of destination |
| `commons-io@2.4` (2013) | :white_check_mark: Inherits permissions of destination |
| `commons-io@2.5` (2016) | :white_check_mark: Inherits permissions of destination |
| `commons-io@2.6` (2017) | :white_check_mark: Inherits permissions of destination |
| `commons-io@2.7` (2020) | :white_check_mark: Inherits permissions of destination |
| `commons-io@2.8` (2020) | :white_check_mark: Inherits permissions of destination |
| `commons-io@2.9` (2021) | :no_entry_sign: Does not inherit permissions of destination |
| `commons-io@2.10` (2021) | :no_entry_sign: Does not inherit permissions of destination |
| `commons-io@2.11` (2021) | :no_entry_sign: Does not inherit permissions of destination |

This PR removes the COPY_ATTRIBUTES flag unless a file is being moved (to match the OS default move behavior), and instead uses NIO calls to preserve the timestamp information.

TODO:
* [ ] :speech_balloon: **Discussion:** With this PR, the solution uses `BasicFileAttributeView.setTimes(FileTime, FileTime, FileTime)` to set ALL timestamp information.  This differs slightly from pre-2.9 versions.  According to the JavaDocs, this may actually set file stamps to beginning of epoch, so this may be undesired.  If there's an IOException, it will fallback to the older `File.setLastModified(long)` API, but it could be argued that this should remain the only API for maximum backwards compatibility.
* [ ] :speech_balloon: **Discussion:** Why did the previous `setLastModified(File, File)` need `PathUtils.setLastModifiedTime()` if
       `sourceFile.isFile()` was `true`, but needed `setLastModifiedTime(File, long)` if `sourceFile.isFile()` was `false`.  [This is the code I'm inquiring about](https://github.com/apache/commons-io/blob/f22a4227401855ecbfdf8184bbe37275c3aeb5c3/src/main/java/org/apache/commons/io/FileUtils.java#L2854-L2857) which will be lost with this PR, so I'd like to understand the history of this to ensure we aren't introducing a regression with the new change.
* [ ] :hourglass_flowing_sand: **Action:** Hand-test this change and impacting APIs on various platforms and compare behavior to pre-2.9 versions.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -35,7 +35,14 @@ import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.nio.charset.UnsupportedCharsetException;
-import java.nio.file.*;
+import java.nio.file.CopyOption;
+import java.nio.file.FileVisitOption;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.LinkOption;
+import java.nio.file.NotDirectoryException;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.nio.file.attribute.BasicFileAttributeView;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.nio.file.attribute.FileTime;
@@ -49,7 +56,6 @@ import java.time.chrono.ChronoLocalDate;
 import java.time.chrono.ChronoLocalDateTime;
 import java.time.chrono.ChronoZonedDateTime;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
@@ -766,7 +772,7 @@ public class FileUtils {
     public static void copyFile(final File srcFile, final File destFile, final boolean preserveFileDate) throws IOException {
         // @formatter:off
         copyFile(srcFile, destFile, preserveFileDate, new CopyOption[] {StandardCopyOption.REPLACE_EXISTING});
-        // @formatter:ond
+        // @formatter:on
     }
 
     /**
@@ -1301,7 +1307,7 @@ public class FileUtils {
                 if (srcFile.isDirectory()) {
                     doCopyDirectory(srcFile, dstFile, fileFilter, exclusionList, preserveDirDate, copyOptions);
                 } else {
-                    copyFile(srcFile, dstFile, copyOptions);
+                    copyFile(srcFile, dstFile, preserveDirDate, copyOptions);
                 }
             }
         }
@@ -2377,7 +2383,8 @@ public class FileUtils {
         requireAbsent(destFile, "destFile");
         final boolean rename = srcFile.renameTo(destFile);
         if (!rename) {
-            copyFile(srcFile, destFile, copyOptions);
+            // Don't interfere with file date on move, handled by StandardCopyOption.COPY_ATTRIBUTES
+            copyFile(srcFile, destFile, false, copyOptions);
             if (!srcFile.delete()) {
                 FileUtils.deleteQuietly(destFile);
                 throw new IOException("Failed to delete original file '" + srcFile + "' after copy to '" + destFile + "'");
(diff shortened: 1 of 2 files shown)


garydgregory:
See also https://issues.apache.org/jira/projects/IO/issues/IO-769

garydgregory:
Hi @tresf 
Thank you for taking the time and putting in the effort to create this PR as well as explaining the issue clearly here and on the development mailing list.

I'd like feedback from the community. @kinow ? @chtompki ?

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -2840,37 +2836,30 @@ private static File requireFileIfExists(final File file, final String name) {
...
+            final BasicFileAttributeView destAttrView = Files.getFileAttributeView(targetFile.toPath(), BasicFileAttributeView.class);
+            // null guards are not needed; BasicFileAttributes.setTimes(...) is null safe
+            destAttrView.setTimes(srcAttr.lastModifiedTime(), srcAttr.lastAccessTime(), srcAttr.creationTime());
+        } catch(IOException unused) {
+            // Fallback: Only set modified time to match source file
+            targetFile.setLastModified(sourceFile.lastModified());
+        }
+
+        // TODO: (Help!) Determine historically why setLastModified(File, File) needed PathUtils.setLastModifiedTime() if
+        //  sourceFile.isFile() was true, but needed setLastModifiedTime(File, long) if sourceFile.isFile() was false
```
kinow, line 2862: Thanks for adding the note. No idea why that changed, didn't have time to dig into `git`'s history, sorry.

garydgregory:
Notice that in the (old) Jira tickets above, two are about attributes not being copied, which we now do but this PR proposes to undo. So catch-22. Ideas?

tresf:
> Notice that in the (old) Jira tickets above, two are about attributes not being copied, which we now do but this PR proposes to undo. So catch-22. Ideas?

The API already offers `COPY_ATTRIBUTES` in several APIs as `CopyOptions` parameters.  Note, the current API doesn't allow the removal of the assumed `COPY_ATTRIBUTES` and if you look at the history of the StackOverflow and mailing list posts, this is often the first instinct, e.g:
1. User doesn't get desired copy behavior
2. Community recommends adjusting `CopyOptions`

... however the current behavior doesn't give this control to the user, despite it being the first instinct by the community, quoting:

> **Gary wrote:** Use `org.apache.commons.io.FileUtils.copyFile(File, File, boolean, CopyOption...)` to exercise full control over what you want to do. Under the covers it ends up calling `java.nio.file.Files.copy(Path, Path, CopyOption...)`.

So if this regression is not reverted, I'd at least advocate for an API which would make the above statement correct.  The current behavior silently adds `COPY_ATTRIBUTES`, which removes the ability to control this flag.

kinow:
>Notice that in the (old) Jira tickets above, two are about attributes not being copied, which we now do but this PR proposes to undo. So catch-22. Ideas?

I think we should start by reverting the change from 2.9 with this PR, and prepare a new release. I was only suggesting that someone looks at the other JIRA issues to close if it is fixed by this PR, or leave it open for the next minor or major release, as needed.

>So if this regression is not reverted, I'd at least advocate for an API which would make the above statement correct. The current behavior silently adds COPY_ATTRIBUTES, which removes the ability to control this flag.

That sounds like a compromise. Probably one of the two would then have to be marked as deprecated. I don't have a large-enough Java code base that I am maintaining a the moment to measure the impact, so take my comments with a pinch of salt :slightly_smiling_face: No hard feelings if other options are chosen here :+1:

garydgregory:
Hi @tresf 
Please rebase on git master to see what happens with the new test for [IO-575](https://issues.apache.org/jira/browse/IO-575).

tresf:
> Hi @tresf, Please rebase on git master to see what happens with the new test for [IO-575](https://issues.apache.org/jira/browse/IO-575).

Done, I think you need to kick-start the workflow now. :)

menscikov:
Hello, `FileUtils.copyInputStreamToFile()` has the same issue starting from **2.9** version.
Please fix it also.

Schmidor:
`FileUtils.moveFile(File srcFile, File destFile) ` was also changed and got COPY_ATTRIBUTES in 2.9

garydgregory:
What is the status of this PR?

garydgregory:
> Why did the previous setLastModified(File, File) need PathUtils.setLastModifiedTime() if
> sourceFile.isFile() was true, but needed setLastModifiedTime(File, long) if sourceFile.isFile() was false. [This is the code I'm inquiring about](https://github.com/apache/commons-io/blob/f22a4227401855ecbfdf8184bbe37275c3aeb5c3/src/main/java/org/apache/commons/io/FileUtils.java#L2854-L2857) which will be lost with this PR, so I'd like to understand the history of this to ensure we aren't introducing a regression with the new change.
> 

This might have to do with different APIs [not] being able to set directory timestamps.

garydgregory:
The consensus, reading this PR and the ticket, is to revert the behavior, so I will merge this PR.

PR #391, opened 2022-10-05
[IO-782] SequenceReader should close readers when its close method is called

digiovinazzo:
Pull request to fix https://issues.apache.org/jira/browse/IO-782

### DIFF
--- a/src/main/java/org/apache/commons/io/input/SequenceReader.java
+++ b/src/main/java/org/apache/commons/io/input/SequenceReader.java
@@ -20,9 +20,12 @@ import static org.apache.commons.io.IOUtils.EOF;
 
 import java.io.IOException;
 import java.io.Reader;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Objects;
+import org.apache.commons.io.IOExceptionList;
 
 /**
  * Provides the contents of multiple Readers in sequence.
@@ -33,6 +36,7 @@ public class SequenceReader extends Reader {
 
     private Reader reader;
     private Iterator<? extends Reader> readers;
+    private Iterable<? extends Reader> readersIterable;
 
     /**
      * Constructs a new instance with readers
@@ -40,7 +44,8 @@ public class SequenceReader extends Reader {
      * @param readers the readers to read
      */
     public SequenceReader(final Iterable<? extends Reader> readers) {
-        this.readers = Objects.requireNonNull(readers, "readers").iterator();
+        this.readersIterable = Objects.requireNonNull(readers, "readers");
+        this.readers = readers.iterator();
         this.reader = nextReader();
     }
 
@@ -60,8 +65,26 @@ public class SequenceReader extends Reader {
      */
     @Override
     public void close() throws IOException {
+        if (readersIterable == null) {
+            // already closed
+            return;
+        }
+
+        final List<IOException> ioExceptionList = new ArrayList<>();
+        for (Reader reader : readersIterable) {
+            try {
+                reader.close();
+            } catch (IOException e) {
+                ioExceptionList.add(e);
+            }
+        }
+        this.readersIterable = null;
         this.readers = null;
         this.reader = null;
+
+        if (!ioExceptionList.isEmpty()) {
+            throw new IOExceptionList(ioExceptionList);
+        }
     }
 
     /**
--- a/src/test/java/org/apache/commons/io/input/SequenceReaderTest.java
+++ b/src/test/java/org/apache/commons/io/input/SequenceReaderTest.java
@@ -56,7 +56,7 @@ public class SequenceReaderTest {
     }
 
     @Test
-    public void testClose() throws IOException {
+    public void testAutoClose() throws IOException {
         try (Reader reader = new SequenceReader(new CharSequenceReader("FooBar"))) {
             checkRead(reader, "Foo");
             reader.close();
@@ -65,6 +65,12 @@ public class SequenceReaderTest {
     }
 
     @Test
+    public void testClose() throws IOException {
+        final Reader reader = new SequenceReader(new CharSequenceReader("FooBar"));
+        checkRead(reader, "Foo");
+        reader.close();
+        checkReadEof(reader);
+    }
     public void testMarkSupported() throws Exception {
         try (Reader reader = new SequenceReader()) {
             assertFalse(reader.markSupported());
--- a/src/test/java/org/apache/commons/io/input/SequenceReaderTest.java
+++ b/src/test/java/org/apache/commons/io/input/SequenceReaderTest.java
@@ -16,9 +16,12 @@
  */
 package org.apache.commons.io.input;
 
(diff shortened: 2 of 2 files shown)


src/main/java/org/apache/commons/io/input/SequenceReader.java
```diff
@@ -16,13 +16,13 @@
...
+import org.apache.commons.io.IOExceptionList;
 
 import java.io.IOException;
 import java.io.Reader;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.Objects;
+import java.util.*;
+
+import static org.apache.commons.io.IOUtils.EOF;
```
garydgregory, line 25: Do not reorder the imports please.

garydgregory (changes requested):

digiovinazzo:
I checked for example `BufferedReader` and it does close the `Reader` passed in the constructor. 
Moreover even the old `java.io.SequenceInputStream` closes its `InputStream`s.

garydgregory:
@digiovinazzo 
Thank you for the clarification, this PR seems to make sense then :-)

src/test/java/org/apache/commons/io/input/SequenceReaderTest.java
```diff
@@ -56,14 +60,86 @@ private void checkReadEof(final Reader reader) throws IOException {
...
+            public void close() throws IOException {
+                closeReader1.set(true);
+            }
+        };
+
+        try (SequenceReader sequenceReader = new SequenceReader(reader1, emptyReader)) {
+            assertEquals('A', sequenceReader.read());
+            assertEquals(EOF, sequenceReader.read());
+        } catch (IOException e) {
+            fail("No IOException expected");
```
garydgregory, line 136: Why not just let the exception percolate to fail the test?

digiovinazzo: That's another option, it was just to have a meaningful message on fail

src/main/java/org/apache/commons/io/input/SequenceReader.java
```diff
@@ -33,19 +36,21 @@ public class SequenceReader extends Reader {
...
     public SequenceReader(final Iterable<? extends Reader> readers) {
-        this.readers = Objects.requireNonNull(readers, "readers").iterator();
+        this.readersIterable = Objects.requireNonNull(readers, "readers");
+        this.readers = readers.iterator();
         this.reader = nextReader();
     }
 
     /**
-     * Constructs a new instance with readers
+     * Construct a new instance with readers
```
garydgregory, line 53: Don't change the Javadoc style please.

src/main/java/org/apache/commons/io/input/SequenceReader.java
```diff
@@ -60,8 +65,26 @@ public SequenceReader(final Reader... readers) {
...
+            // already closed
+            return;
+        }
+
+        final List<IOException> ioExceptionList = new ArrayList<>();
+        for (Reader reader : readersIterable) {
+            try {
+                reader.close();
+            } catch (IOException e) {
+                ioExceptionList.add(e);
```
garydgregory, line 78: I'm not sure about the semantics here, which are completely different from `SequenceInputStream`. For good or bad, I think it would be better for the closing logic of this class to match `SequenceInputStream`.

src/main/java/org/apache/commons/io/input/SequenceReader.java
```diff
@@ -33,14 +36,16 @@ public class SequenceReader extends Reader {
 
     private Reader reader;
     private Iterator<? extends Reader> readers;
+    private Iterable<? extends Reader> readersIterable;
```
garydgregory, line 39: No need to change the name, it makes the PR noisier.

src/test/java/org/apache/commons/io/input/SequenceReaderTest.java
```diff
@@ -56,14 +60,86 @@ private void checkReadEof(final Reader reader) throws IOException {
...
+        reader.close();
+        checkReadEof(reader);
+    }
+
+    @Test
+    public void testCloseReaders() {
+        final AtomicBoolean closeEmptyReader = new AtomicBoolean(false);
+        final Reader emptyReader = new Reader() {
+            @Override
+            public int read(char[] cbuf, int off, int len) throws IOException {
```
garydgregory, line 84: Use `final` where you can.

src/test/java/org/apache/commons/io/input/SequenceReaderTest.java
```diff
@@ -56,14 +60,86 @@ private void checkReadEof(final Reader reader) throws IOException {
...
+
+            @Override
+            public void close() throws IOException {
+                closeEmptyReader.set(true);
+            }
+        };
+
+        final AtomicBoolean closeReader1 = new AtomicBoolean(false);
+        final Reader reader1 = new Reader() {
+            private final char[] content = new char[]{'A'};
```
garydgregory, line 100: Use abbreviated notation, IOW `... = {'A'};`

src/test/java/org/apache/commons/io/input/SequenceReaderTest.java
```diff
@@ -56,14 +60,86 @@ private void checkReadEof(final Reader reader) throws IOException {
...
+            private int position = 0;
+
+            @Override
+            public int read(char[] cbuf, int off, int len) throws IOException {
```
garydgregory, line 104: Use `final` where you can.

src/test/java/org/apache/commons/io/input/SequenceReaderTest.java
```diff
@@ -56,14 +60,86 @@ private void checkReadEof(final Reader reader) throws IOException {
...
+                return EOF;
+            }
+
+            @Override
+            public void close() throws IOException {
+                closeEmptyReader.set(true);
+            }
+        };
+
+        final AtomicBoolean closeReader1 = new AtomicBoolean(false);
```
garydgregory, line 98: No need to initialize an `AtomicBoolean` to `false`, that's the default, IOW -> `new AtomicBoolean();`

src/test/java/org/apache/commons/io/input/SequenceReaderTest.java
garydgregory: Simpler is better:
```
        try (SequenceReader sequenceReader = new SequenceReader(reader1, emptyReader)) {
            assertEquals('A', sequenceReader.read());
            assertEquals(EOF, sequenceReader.read());
        } finally {
            assertTrue(closeEmptyReader.get());
            assertTrue(closeReader1.get());
        }
```

garydgregory (changes requested):

garydgregory:
The issue is now resolved in git master by making the implementation closer to what `SequenceInputStream` does. Please verify your use case and close. TY!

garydgregory:
Closing per my previous comment and no feedback in 4 days.

digiovinazzo:
I was offline last week, @garydgregory thanks for taking care of this!

PR #418, opened 2022-12-29
[IO-784] Add support for Appendable to HexDump util

fkjellberg:
This PR adds support to write the output from HexDump utility to an `Appendable` making it possible to use e.g. `System.out` or `StringBuilder` as output targets.

### DIFF
--- a/src/main/java/org/apache/commons/io/HexDump.java
+++ b/src/main/java/org/apache/commons/io/HexDump.java
@@ -18,8 +18,11 @@ package org.apache.commons.io;
 
 import java.io.IOException;
 import java.io.OutputStream;
+import java.io.OutputStreamWriter;
 import java.nio.charset.Charset;
 import java.util.Objects;
+
+import org.apache.commons.io.output.CloseShieldOutputStream;
 
 /**
  * Dumps data in hexadecimal format.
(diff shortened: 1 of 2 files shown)


garydgregory:
Code coverage is negative with this PR, so you are likely missing some code paths in the new code.

fkjellberg:
@garydgregory I added one more test and rebased the branch

src/main/java/org/apache/commons/io/HexDump.java
```diff
@@ -53,7 +56,26 @@ public class HexDump {
...
+     * characters (if any) that those bytes represent printed per each line
+     * of output.
+     *
+     * @param data  the byte array to be dumped
+     * @param out  the Appendable to which the data is to be written
+     *
+     * @throws IOException is thrown if anything goes wrong writing
+     *         the data to appendable
+     */
+
```
garydgregory, line 71: Remove whitespace between Javadoc and code.

garydgregory: Add Javadoc since tag to new public and protected elements.

src/main/java/org/apache/commons/io/HexDump.java
```diff
@@ -118,14 +144,55 @@ public static void dump(final byte[] data, final long offset,
...
+     * @throws ArrayIndexOutOfBoundsException if the index is
+     *         outside the data array's bounds
+     * @throws NullPointerException if the output stream is null
+     */
+    public static void dump(final byte[] data, final long offset,
+                            final OutputStream stream, final int index)
+            throws IOException, ArrayIndexOutOfBoundsException {
+        Objects.requireNonNull(stream, "stream");
+
+        // make explicit the dependency on the default encoding
```
garydgregory, line 189: This comment is not needed: "make explicit the dependency on the default encoding"

src/main/java/org/apache/commons/io/HexDump.java
```diff
@@ -53,7 +56,26 @@ public class HexDump {
...
+     * of output.
+     *
+     * @param data  the byte array to be dumped
+     * @param out  the Appendable to which the data is to be written
+     *
+     * @throws IOException is thrown if anything goes wrong writing
+     *         the data to appendable
+     */
+
+    public static void dump(final byte[] data, final Appendable out)
```
garydgregory, line 72: Rename `out` to `appendable` so it has no chance of being confused with an Output* class.

garydgregory (changes requested):
Please see my comments.

src/main/java/org/apache/commons/io/HexDump.java
fkjellberg: @garydgregory Thanks for reviewing. I've fixed both.

src/main/java/org/apache/commons/io/HexDump.java
fkjellberg: This comment came from the original code. I've removed it now.

src/main/java/org/apache/commons/io/HexDump.java
fkjellberg: Fixed

src/main/java/org/apache/commons/io/HexDump.java
```diff
@@ -118,14 +148,54 @@ public static void dump(final byte[] data, final long offset,
...
+     * @throws ArrayIndexOutOfBoundsException if the index is
+     *         outside the data array's bounds
+     * @throws NullPointerException if the output stream is null
+     */
+    public static void dump(final byte[] data, final long offset,
+                            final OutputStream stream, final int index)
+            throws IOException, ArrayIndexOutOfBoundsException {
+        Objects.requireNonNull(stream, "stream");
+
+        try (OutputStreamWriter out = new OutputStreamWriter(CloseShieldOutputStream.wrap(stream), Charset.defaultCharset())) {
```
garydgregory, line 193: No need for `CloseShieldOutputStream`.

src/main/java/org/apache/commons/io/HexDump.java
```diff
@@ -118,14 +148,54 @@ public static void dump(final byte[] data, final long offset,
...
+            dump(data, offset, out, index, data.length - index);
+            out.flush();
```
garydgregory, line 195: Let the caller decide when to flush and/or close.

garydgregory (changes requested):
Hello @fkjellberg 
Please see my comments.

src/main/java/org/apache/commons/io/HexDump.java
fkjellberg: Commenting both on the `CloseShieldOutputStream` comment above and this comment.

The original implementation flushed the underlying stream. It actually flushed at every row written. Since the `OutputStreamWriter` will be automatically closed within the try-with-resources block and looking at the source code for `StreamEncoder` within the `OutputStreamWriter`, I now notice that the `StreamEncoder` will flush before closing the stream. I will remove the explicit flush in a follow up commit.

The original implementation left the `OutputStream` open after the call. Since I wrap the `OutputStream` in an `OutputStreamWriter` that is closed, it will propagate that close call to the underlying stream as well. I'm using the `CloseShieldOutputStream` to protect it from being closed and preserve the same behavior as before.

I think we should preserve the same behavior as the original code when it comes to keeping the `OutputStream` open after the call.

src/test/java/org/apache/commons/io/HexDumpTest.java
```diff
@@ -253,6 +253,16 @@ public void testDumpOutputStream() throws IOException {
 
         // verify proper behavior with null stream
         assertThrows(NullPointerException.class, () -> HexDump.dump(testArray, 0x10000000, null, 0));
+
```
garydgregory, line 256: How about reusing `ThrowOnCloseInputStream`?

fkjellberg: @garydgregory Thanks for the pointer. I was not aware of this test utility class. Code updated.

garydgregory:
I need to look at this more closely over the weekend: I don't know why the HTTP providers have to be unique and different compared to all the others. Is the use of concepts in this PR backward? In the PR, the "free" code now _also_ "closes" resources and that feels backward to me. I expect the "close" code to also "free" resources are part of closing, not the other way around.
Any thoughts?

PR #427, opened 2023-01-31
Simplified res closing and prevented creating/closing process OutputS…

tbw777:
…tream and ErrorStream.

There is no needed to make local variables as null and opening unused streams to close them later.


garydgregory:
My only concern here is that the method will now fail if _anything_ goes wrong closing streams. This is a change in behavior that we do not test apparently. This could be considered reasonable if other processes try to use the same resources but I am not 100% sure. What do others think?

tbw777:
@garydgregory 
The exception will also be thrown to the top because there is no corresponding catch as in the original version. Or do you mean something else?

garydgregory:
Hm, two behavior might be the same indeed. I'll look again in the AM.

PR #442, opened 2023-03-22
add filterReadeLines & consumerReadeLine

li-keguo:
add
```java
void readLines(final File file, final Charset charset, final Consumer<String> consumer);
List<String> readLines(final File file, final Charset charset, final Predicate<String> filter);
```


garydgregory:
-1: no unit tests, and the build breaks.

li-keguo:
> Hello @li-keguo Why not use Java's Files.lines() APIs instead of these?

HI @garydgregory 
It is a good idea to implement filters using Files.lines.
```java
Files.lines(file.toPath(), charset)
                    .filter(filter)
                    .collect(Collectors.toList());
```
However, this is not consumer-friendly. The stream intermediate process does not trigger execution.

li-keguo:
Help me fix ci

li-keguo:
ok I close

PR #447, opened 2023-04-11
Add blocking read mode to QueueInputStream.

maxxedev:
Add blocking read mode to QueueInputStream.

To more closely mimic InputStream::read behavior, add blocking read mode to QueueInputStream that will wait until a queue element becomes available.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/QueueInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/QueueInputStream.java
@@ -23,6 +23,7 @@ import java.io.InputStream;
 import java.io.InterruptedIOException;
 import java.io.PipedInputStream;
 import java.io.PipedOutputStream;
+import java.io.UncheckedIOException;
 import java.util.Objects;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -86,12 +99,24 @@ public QueueOutputStream newQueueOutputStream() {
...
-     * Reads and returns a single byte.
+     * Reads and returns a single byte. In blocking read mode, the method will wait if necessary until a queue element
+     * becomes available. In non-blocking read mode, the method will return -1 immediately if the queue is empty.
      *
      * @return either the byte read or {@code -1} if the end of the stream has been reached
+     * @throws InterruptedIOException if the thread is interrupted while reading from the queue.
      */
     @Override
-    public int read() {
+    public int read() throws InterruptedIOException {
```
garydgregory, line 109: -1; I'm not a fan of this style because it is the opposite of OO: For each char, we test and execute one branch or another _based on state that never changes_, which tells me this should be implemented either as a subclass or pluggable strategy (with a lambda for example).

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -88,12 +111,32 @@ public QueueOutputStream newQueueOutputStream() {
...
+            super(blockingQueue);
+        }
+
+        @Override
+        public int read() {
+            try {
+                return blockingQueue.take();
+            } catch (final InterruptedException e) {
+                Thread.currentThread().interrupt();
+                final InterruptedIOException ioException = new InterruptedIOException();
```
garydgregory, line 134: @maxxedev 
This seems unusual, why not simply use `new InterruptedIOException(e)` or even simpler `throw new UncheckedIOException(e)`?

src/test/java/org/apache/commons/io/input/QueueInputStreamTest.java
```diff
@@ -17,24 +17,29 @@
 package org.apache.commons.io.input;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.apache.commons.io.input.QueueInputStream.createBlockingQueueInputStream;
```
garydgregory, line 20: @maxxedev 
Only use static imports for JUnit assertions, otherwise, it's too confusing as to who does what IMO.

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -57,24 +59,45 @@
...
-     * Constructs a new instance with given buffer
+     * Constructs a new instance with given queue
      *
      * @param blockingQueue backing queue for the stream
      */
     public QueueInputStream(final BlockingQueue<Integer> blockingQueue) {
         this.blockingQueue = Objects.requireNonNull(blockingQueue, "blockingQueue");
     }
 
+    /**
```
garydgregory, line 80: I'd rather not have these factory methods, if we need something to build, let's use a Builder.

garydgregory: @maxxedev Hm, let me try something and get back to you...

garydgregory:
@maxxedev See git master where I took a different approach. What's next to do is a test for the new "Taking" class. I think it might make sense to have the test in a subclass or a completely separate class for clarity. The refactored test class makes it much harder to understand what we are testing and how (IMO).

src/main/java/org/apache/commons/io/input/QueueInputStream.java
maxxedev: UncheckedIOException constructor requires a IOException.

InterruptedException does not extend IOException
InterruptedIOException extends IOException

maxxedev:
> @maxxedev See git master where I took a different approach. [...]

Thanks. I will look into improving the tests

garydgregory:
> > @maxxedev See git master where I took a different approach. [...]
> 
> Thanks. I will look into improving the tests

You can rebase this PR, or close it and create a new one. Up to you ;-)

maxxedev:
Thinking more about this, there might be a way to keep it simple and still achieve the same result.

I opened a new PR. see [https://github.com/apache/commons-io/pull/452 ](https://github.com/apache/commons-io/pull/452)

garydgregory:
Closing: PR #452 merged

PR #449, opened 2023-04-19
IO-789 Symbolic link file filter accepts all files, not just symbolic links

SwingGuy1024:
This change fixed the filters, but also restructures the code slightly to facilitate Unit Testing on all platforms. (This is a challenge for this particular class.


garydgregory:
Superceded by #450

PR #450, opened 2023-04-19
IO-790 symbolic link file filter test incomplete, needs to handle windows issues

SwingGuy1024:
This should be pulled after IO-789

### DIFF
--- a/src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
+++ b/src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
@@ -17,191 +17,21 @@
 
 package org.apache.commons.io.filefilter;
 
-import java.io.File;
-import java.io.IOException;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
 import java.nio.file.FileVisitResult;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.util.function.BiFunction;
 
 import org.apache.commons.io.file.PathUtils;
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
-
-import static org.junit.jupiter.api.Assertions.*;
 
 /**
  * Tests {@link SymbolicLinkFileFilter}.
  */
 public class SymbolicLinkFileFilterTest {
 
-    public static final String TARGET_SHORT_NAME = "SLFF_Target";
-    public static final String TARGET_EXT = ".txt";
-    public static final String TARGET_NAME = TARGET_SHORT_NAME + TARGET_EXT;
-    public static final String DIRECTORY_NAME = "SLFF_TargetDirectory";
-    public static final String DIRECTORY_LINK_NAME = "SLFF_LinkDirectory";
-    public static final String MISSING = "Missing";
-    private static File testTargetFile;         // hard file
-    private static Path testTargetPath;         // hard file Path
-    private static File parentDirectoryFile;    // System Temp directory
-    private static File testLinkFile;           // symbolic link to hard file
-    private static String linkName;             // Name of link file
-    private static Path testLinkPath;           // symbolic link to hard file Path
-    private static File targetDirFile;          // 
-    private static Path targetDirPath;          // hard directory Path
-    private static Path linkDirPath;            // symbolic link to hardDirectory
-    private static File linkDirFile;
-    private static SymbolicLinkFileFilter filter;
-    private static File missingFile;    // non-existent file
-
-    private static Path createRealSymbolicLink(Path link, Path target) {
-        try {
-            if (Files.exists(link)) {
-                Files.delete(link);
-            }
-            return Files.createSymbolicLink(link, target);
-        } catch (IOException e) {
-            throw new IllegalStateException("Failure to create Symbolic Link", e);
-        }
-    }
-
-    private static Path createMockSymbolicLink(Path lnk, Path tgt) {
-        try {
-            return Files.createFile(lnk);
-        } catch (IOException e) {
-            throw new IllegalStateException("Failure to create Symbolic Link", e);
-        }
-    }
-
-    // Mock filter for testing on Windows.
-    private static SymbolicLinkFileFilter createMockFilter() {
-        return new SymbolicLinkFileFilter() {
-            @Override
-            boolean isSymbolicLink(final Path filePath) {
-                return filePath.toFile().exists() && filePath.toString().contains("Link"); // Mock test
-            }
-        };
-    }
-
-    /**
-     * <p>Unit test setup creates a hard file, a symbolic link to the hard file, a hard directory,
-     * and a symbolic link to that directory. All are created in the temp directory</p>
-     * <p>Unit test teardown deletes all four of these files.</p>
-     *
-     * @throws IOException If it fails to create the temporary files
-     */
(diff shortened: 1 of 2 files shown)


garydgregory:
@SwingGuy1024 
You have 2 similar PRs for this class and this one fails. Which PR do you want us to review? A PR for a fix should contain changes to main and test, where a test fails if the changes to main are not applied.

SwingGuy1024:
Gary,
The PR for IO-790 has changes from both PRs, so that's what you should review. Sorry about that. I shouldn't have created two separate JIRA issues for this.
— Miguel Muñoz
———
4210 Via Arbolada #226Los Angeles, CA 90042
323-225-7285
–

The Sun, with all those planets going around it and dependent on it, can still ripen a vine of grapes like it has nothing else to do in the world.
  — Galileo 

    On Wednesday, April 19, 2023 at 03:07:29 AM PDT, Gary Gregory ***@***.***> wrote:  
 
 


@SwingGuy1024
You have 2 similar PRs for this class and this one fails. Which PR do you want us to review? A PR for a fix should contain changes to main and test, where a test fails if the changes to main are not applied.

garydgregory:
You really should not have your address and phone number in emails IMO.

Gary

src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
```diff
@@ -17,21 +17,198 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.function.BiFunction;
 
 import org.apache.commons.io.file.PathUtils;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
```
garydgregory, line 32: Don't change the order of static imports, keep them first.

src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
```diff
@@ -17,21 +17,198 @@
...
+    private static SymbolicLinkFileFilter filter;
+
+    private static Path createRealSymbolicLink(Path link, Path target) {
+        try {
+            if (Files.exists(link)) {
+                Files.delete(link);
+            }
+            return Files.createSymbolicLink(link, target);
+        } catch (IOException e) {
+            throw new IllegalStateException("Failure to create Symbolic Link", e);
```
garydgregory, line 67: Let the exception percolate up.

src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
```diff
@@ -17,21 +17,198 @@
...
+        }
+    }
+
+    private static Path createMockSymbolicLink(Path lnk, Path tgt) {
+        try {
+            return Files.createFile(lnk);
+        } catch (IOException e) {
+            throw new IllegalStateException("Failure to create Symbolic Link", e);
```
garydgregory, line 75: Let the exception percolate up.

src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
```diff
@@ -17,21 +17,198 @@
...
+     * @throws IOException If it fails to create the temporary files
+     */
+    @BeforeAll
+    static void testSetup() throws IOException {
+        final BiFunction<Path, Path, Path> symbolicLinkCreator;
+
+        // We can't create symbolic links on Windows without admin privileges,
+        // so iff that's our OS, we mock them.
+        final String os = System.getProperty("os.name");
+        if (os.toLowerCase().contains("windows")) {
```
garydgregory, line 103: Reuse `org.apache.commons.lang3.SystemUtils.IS_OS_WINDOWS`

src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
```diff
@@ -17,21 +17,198 @@
...
+            if (Files.exists(link)) {
+                Files.delete(link);
+            }
+            return Files.createSymbolicLink(link, target);
+        } catch (IOException e) {
+            throw new IllegalStateException("Failure to create Symbolic Link", e);
+        }
+    }
+
+    private static Path createMockSymbolicLink(Path lnk, Path tgt) {
```
garydgregory, line 71: No weird abbreviations needed.

src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
```diff
@@ -17,21 +17,198 @@
...
+    private static File targetDirFile;          //
+    private static Path targetDirPath;          // hard directory Path
+    private static Path testLinkDirPath;        // symbolic link to hardDirectory
+    private static File testLinkDirFile;
+    private static File missingFile;            // non-existent file
+    private static SymbolicLinkFileFilter filter;
+
+    private static Path createRealSymbolicLink(Path link, Path target) {
+        try {
+            if (Files.exists(link)) {
```
garydgregory, line 62: @SwingGuy1024 
Why is `Files.deleteIfExists()` not used here?

src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
SwingGuy1024: Done.

src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
SwingGuy1024: Not sure why this is important in a unit test, for an exception we never expect to see. But I've rewritten the code to let the exception percolate up, as requested.
I feel wrapping a checked exception that we never expect to see is a very efficient way to handle them in interface methods that don't declare exceptions like BiFunction. (If we expect the exception to occur in a real-world scenario, I would handle it differently.)

src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
SwingGuy1024: Because I'm new to the Files API. :) I changed it.

src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
SwingGuy1024: Done.

src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
SwingGuy1024: Done. This is an improvement. Thanks.

src/test/java/org/apache/commons/io/filefilter/SymbolicLinkFileFilterTest.java
SwingGuy1024: Done. CheckStyles doesn't have a test for this? (I'm not too familiar with CheckStyles anymore. It's been years.)

PR #453, opened 2023-04-23
Apply nanoseconds precision for timeout duration.

maxxedev:
```
Apply nanoseconds precision for timeout duration.
Also improve javadocs
```


kinow (commented):
Looks good to me. I couldn't see anywhere in the class telling the unit of time for the timeout. Maybe that should go in the javadoc? Thanks!

maxxedev:
The timeout is configurable through the public API with a `Duration` value which encapsulates both time amount and time unit.

This PR changes so that internal usage and conversion of Duration is more precise up to nanoseconds.

PR #457, opened 2023-05-14
Implemented overloaded readFileToByteArray() method 

jmloudis:
Hello team,

Implemented a new overloaded readFileToByteArray() method that creates byte[] based on offset and length parameters. This is a new feature that was request by the following [jira ticket](https://issues.apache.org/jira/projects/IO/issues/IO-548?filter=allissues). 

- The file will be read starting by the arrayOffset parameter and read up until the arrayLength parameter based on the byte[] index.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -2582,6 +2582,29 @@ public class FileUtils {
     }
 
     /**
+     * Reads the contents of a file into a byte[]. Start reading from offSet in file index to specific length.
+     *
+     * @param file the file to read, nust not be {@code null}
+     * @param arrayOffset read from an offset in the file
+     * @param arrayLength read file up to certain length based on index
+     * @return byte[]
+     * @throws IOException if an I/O error occurs
+     * @since 2.12.0
+     */
+    public static byte[] readFileToByteArray(final File file, final int arrayOffset, int arrayLength) throws IOException {
+        Objects.requireNonNull(file, "file");
+        try (InputStream inputStream = openInputStream(file)) {
+            final long fileLength = file.length();
+            // file.length() may return 0 for system-dependent entities, treat 0 as unknown length - see IO-453
+            return fileLength > 0 ? IOUtils.toByteArray(inputStream, arrayOffset, arrayLength) : IOUtils.toByteArray(inputStream);
+        }
+
+
+    }
+
+
+
+    /**
      * Reads the contents of a file into a String using the default encoding for the VM.
      * The file is always closed.
      *
--- a/src/main/java/org/apache/commons/io/IOUtils.java
+++ b/src/main/java/org/apache/commons/io/IOUtils.java
@@ -2664,6 +2664,50 @@ public class IOUtils {
         return data;
     }
 
+
+    /**
+     *
+     * Gets the contents of an {@link InputStream} as a {@code byte[]}.Use this method when you want to specify the
+     * indexes that you want to read from the InputStream.
+     *
+     * @param input
+     * @param offset The offset from the start of the InputStream to start ready - index based
+     * @param length The end of the InputStream to stop reading - index based
+     * @return byte[]
+     * @throws IOException if an I/O error occurs or {@link InputStream} length is smaller than parameter {@code size}.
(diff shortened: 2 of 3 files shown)


src/test/java/org/apache/commons/io/FileUtilsTest.java
```diff
@@ -2446,6 +2447,57 @@ public void testReadFileToByteArray() throws Exception {
...
+        assertEquals(30, data[0]); // return one byte
+    }
+
+    @Test
+    public void testReadFileToByteArrayOffsetAndLengthZero() throws Exception {
+        final File file = new File(tempDirFile, "read.txt");
+        final byte[] arr = new byte[] {10, 20, 30, 40 ,50, 60, 70};
+        Files.write(file.toPath(), arr);
+
+        final byte[] data = FileUtils.readFileToByteArray(file, 0, 0);
```
garydgregory, line 2479: The resulting arrays are off by 1 in length, if I ask to read a length of 0, i should get back a byte array of length 0.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -2581,6 +2581,29 @@ public static byte[] readFileToByteArray(final File file) throws IOException {
         return Files.readAllBytes(file.toPath());
     }
 
+    /**
+     * Reads the contents of a file into a byte[]. Start reading from offSet in file index to specific length.
+     *
+     * @param file the file to read, nust not be {@code null}
+     * @param arrayOffset read from an offset in the file
```
garydgregory, line 2588: Parameters are misnamed: This is the index in the source file.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -2581,6 +2581,29 @@ public static byte[] readFileToByteArray(final File file) throws IOException {
...
+     * @param arrayLength read file up to certain length based on index
```
garydgregory, line 2589: This is how much to read from the file.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2664,6 +2664,50 @@ public static byte[] toByteArray(final InputStream input, final int size) throws
...
+     */
+    public static byte[] toByteArray(final InputStream input, final int offset, final int length) throws IOException {
+
+        if (length < 0) {
+            throw new IllegalArgumentException("Size must be equal or greater than zero: " + length);
+        }
+        if (offset < 0 ) {
+            throw new IllegalArgumentException("Offset must be grater than or equal to zero: " + offset);
+        }
+        if (length < offset) {
```
garydgregory, line 2690: The index and length are unrelated, the length is the length of the desired result. The question is what should happen if there is not enough file data.

garydgregory (changes requested):
-1: Does not match the ticket but the ticket is closed anyway.

jmloudis:
Thank you for the feedback! I will look fixing these issues this weekend.

src/test/java/org/apache/commons/io/FileUtilsTest.java
```diff
@@ -2446,6 +2447,56 @@ public void testReadFileToByteArray() throws Exception {
...
+    public void testReadFileToByteArrayOffsetAndLengthException() {
+
+        final File file = new File(tempDirFile, "read.txt");
+        final byte[] arr = new byte[] {10, 20, 30, 40, 50, 60, 70};
+
+        try {
+            Files.write(file.toPath(), arr);
+            FileUtils.readFileToByteArray(file, 2, 1);
+        } catch (IOException | IllegalArgumentException ex) {
+            assertEquals("Length must be greater than or equal to offset: " + 2, ex.getMessage());
```
garydgregory, line 2494: This makes no sense to me: the position and length should not be related; "2, 1" should mean, starting at 0-based position 2, read one byte. IOW, I am describing the same semantics as 'java.io.InputStream.read(byte[], int, int)'. In this test, the file is 7 bytes in length, so "2, 1", should return the third byte value of "30".

garydgregory (changes requested):
Please see my comments.

garydgregory:
Closing, no reply in 3 weeks.

PR #462, opened 2023-06-26
IO-805 - FilenameUtils.isIPv4Address does not handle illegal integers

CNJingo:
This change catches the numberFormatException in isIpv4Address and then returns false instead of throwing an exception.

### DIFF
--- a/src/main/java/org/apache/commons/io/FilenameUtils.java
+++ b/src/main/java/org/apache/commons/io/FilenameUtils.java
@@ -1074,7 +1074,12 @@ public class FilenameUtils {
         // verify that address subgroups are legal
         for (int i = 1; i <= 4; i++) {
             final String ipSegment = m.group(i);
-            final int iIpSegment = Integer.parseInt(ipSegment);
+            final int iIpSegment;
+            try {
+                iIpSegment = Integer.parseInt(ipSegment);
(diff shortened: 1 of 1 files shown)


kinow (changes requested):
I think this deserves a junit test case to reproduce the issue.

CNJingo:
Hi @kinow, isIpv4Address method is a private method. there is a way can I test it?

kinow:
> Hi @kinow, isIpv4Address method is a private method. there is a way can I test it?

There must be. It must be accessible via a public method somewhere. I think `getPrefixLength` calls` `isIpv6` that calls the `isIpv4`. If that's correct, then a test would have to be written for `getPrefixLength` that triggers the code modified here, otherwise there's no way to make sure that a) it works as expected nor b) that it won't be changed by another dev later.

CNJingo:
@kinow Thanks for the advice. I'll add test code using ```getPrefixLength``` method.

CNJingo:
I'll close this PR. isIPv4Address method already validate parameter using Matcher with regex '^(\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})$'. I think there is no possibility Integer.parseInt() throws numberFormatException.

PR #486, opened 2023-09-28
[IO-810] Wrong Javadoc regarding FileNotFoundException in FileUtils

chkpnt:
Update Javadoc of `FileUtils.readFileToByteArray(...)`, `FileUtils.readFileToString(...)` and `FileUtils.readLines(...)` regarding `FileNotFoundException`.

Since commons-io [2.12](https://github.com/apache/commons-io/commit/dcb09db922e1edbd6dc6cb1531b827451d58aa83#diff-0bd8160a11e12ea1f96476a00e78d0a95b965d07c7fe126e1af498c08511dd70), the methods throw a `NoSuchFileException` rather than a `FileNotFoundException`.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -2570,9 +2570,8 @@ public class FileUtils {
      * @param file the file to read, must not be {@code null}
      * @return the file contents, never {@code null}
      * @throws NullPointerException if file is {@code null}.
-     * @throws FileNotFoundException if the file does not exist, is a directory rather than a regular file, or for some
-     *         other reason cannot be opened for reading.
-     * @throws IOException if an I/O error occurs.
+     * @throws IOException if an I/O error occurs, including when the file does not exist, is a directory rather than a
+     *         regular file, or for some other reason why the file cannot be opened for reading.
      * @since 1.1
      */
     public static byte[] readFileToByteArray(final File file) throws IOException {
@@ -2587,9 +2586,8 @@ public class FileUtils {
      * @param file the file to read, must not be {@code null}
      * @return the file contents, never {@code null}
      * @throws NullPointerException if file is {@code null}.
-     * @throws FileNotFoundException if the file does not exist, is a directory rather than a regular file, or for some
-     *         other reason cannot be opened for reading.
-     * @throws IOException if an I/O error occurs.
+     * @throws IOException if an I/O error occurs, including when the file does not exist, is a directory rather than a
+     *         regular file, or for some other reason why the file cannot be opened for reading.
      * @since 1.3.1
      * @deprecated 2.5 use {@link #readFileToString(File, Charset)} instead (and specify the appropriate encoding)
      */
@@ -2606,9 +2604,8 @@ public class FileUtils {
      * @param charsetName the name of the requested charset, {@code null} means platform default
      * @return the file contents, never {@code null}
      * @throws NullPointerException if file is {@code null}.
-     * @throws FileNotFoundException if the file does not exist, is a directory rather than a regular file, or for some
-     *         other reason cannot be opened for reading.
-     * @throws IOException if an I/O error occurs.
+     * @throws IOException if an I/O error occurs, including when the file does not exist, is a directory rather than a
+     *         regular file, or for some other reason why the file cannot be opened for reading.
      * @since 2.3
      */
     public static String readFileToString(final File file, final Charset charsetName) throws IOException {
@@ -2622,9 +2619,8 @@ public class FileUtils {
      * @param charsetName the name of the requested charset, {@code null} means platform default
      * @return the file contents, never {@code null}
      * @throws NullPointerException if file is {@code null}.
-     * @throws FileNotFoundException if the file does not exist, is a directory rather than a regular file, or for some
-     *         other reason cannot be opened for reading.
-     * @throws IOException if an I/O error occurs.
+     * @throws IOException if an I/O error occurs, including when the file does not exist, is a directory rather than a
+     *         regular file, or for some other reason why the file cannot be opened for reading.
      * @throws java.nio.charset.UnsupportedCharsetException thrown instead of {@link java.io
      * .UnsupportedEncodingException} in version 2.2 if the named charset is unavailable.
      * @since 2.3
@@ -2640,9 +2636,8 @@ public class FileUtils {
      * @param file the file to read, must not be {@code null}
      * @return the list of Strings representing each line in the file, never {@code null}
      * @throws NullPointerException if file is {@code null}.
-     * @throws FileNotFoundException if the file does not exist, is a directory rather than a regular file, or for some
-     *         other reason cannot be opened for reading.
-     * @throws IOException if an I/O error occurs.
+     * @throws IOException if an I/O error occurs, including when the file does not exist, is a directory rather than a
+     *         regular file, or for some other reason why the file cannot be opened for reading.
      * @since 1.3
      * @deprecated 2.5 use {@link #readLines(File, Charset)} instead (and specify the appropriate encoding)
      */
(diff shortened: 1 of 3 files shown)


src/test/java/org/apache/commons/io/FileUtilsTest.java
```diff
@@ -2463,18 +2485,56 @@ public void testReadFileToStringWithEncoding() throws Exception {
...
-        try {
-            final String[] data = {"hello", "/u1234", "", "this is", "some text"};
-            TestUtils.createLineBasedFile(file, data);
+        final String[] data = {"hello", "/u1234", "", "this is", "some text"};
+        TestUtils.createLineBasedFile(file, data);
 
-            final List<String> lines = FileUtils.readLines(file, UTF_8);
-            assertEquals(Arrays.asList(data), lines);
-        } finally {
-            TestUtils.deleteFile(file);
```
chkpnt, line 2476: The file is created in the tempDir, so deleting it explicitly isn't necessary.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -2698,7 +2691,7 @@ private static void requireAbsent(final File file, final String name) throws Fil
      *
      * @param file1 The first file to compare.
      * @param file2 The second file to compare.
-     * @throws IOException if an I/O error occurs.
+     * @throws IOException              if an I/O error occurs.
```
chkpnt, line 2694: Ups... auto-formatting. I'll revert the whitespace changes from methods which are not part of IO-810.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -2618,15 +2615,14 @@ public static String readFileToString(final File file, final Charset charsetName
...
      * @return the file contents, never {@code null}
      * @throws NullPointerException if file is {@code null}.
-     * @throws FileNotFoundException if the file does not exist, is a directory rather than a regular file, or for some
-     *         other reason cannot be opened for reading.
-     * @throws IOException if an I/O error occurs.
+     * @throws IOException if an I/O error occurs, including when the file does not exist, is a directory rather than a
+     *         regular file, or for some other reason why the file cannot be opened for reading.
      * @throws java.nio.charset.UnsupportedCharsetException thrown instead of {@link java.io
-     * .UnsupportedEncodingException} in version 2.2 if the named charset is unavailable.
+     *                                                      .UnsupportedEncodingException} in version 2.2 if the named charset is unavailable.
```
garydgregory, line 2625: Hi @chkpnt 
Thank you for your PR.
Please don't change the formatting, it makes it harder to review the actual changes.

garydgregory (changes requested):
Hi @chkpnt 
Thank you for your PR.
Please don't change the formatting, it makes it harder to review the actual changes.

src/main/java/org/apache/commons/io/FileUtils.java
chkpnt: Tried to apply the boys scouting rule for the sake of consistency... (sometimes the javadocs are aligned, sometime not). But yes, it was an unsolicited change, sorry for that!

(IHMO formatting shouldn't be something handled manually, I can recommend the integration of something like spotless in the build chain. An other issue I had was that my IDE added the new imports with an wildcard. As I didn't saw any other wildcard imports, I thought that might be undesired, so I had to change the import to single imports manually.)

src/test/java/org/apache/commons/io/FileUtilsTest.java
```diff
@@ -2463,18 +2487,56 @@ public void testReadFileToStringWithEncoding() throws Exception {
...
-        }
+        final List<String> lines = FileUtils.readLines(file, UTF_8);
+        assertEquals(Arrays.asList(data), lines);
+    }
+
+    @Test
+    public void testReadLines_Errors() {
+        assertThrows(NullPointerException.class, () -> FileUtils.readLines(null));
+        assertThrows(NoSuchFileException.class, () -> FileUtils.readLines(new File("non-exsistent")));
```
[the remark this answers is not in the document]
chkpnt: My thoughts were the following:
- the Javadoc is specifying the API to the users of the library
- the tests can be more detailed, so that the common-io-developers know which concrete exception is thrown.
- if the implementation changes, so that another exception is thrown, the developer get noticed and can consider about the consequences (e.g. writing a simple hint to the changelog).

But of course, if you see the tests as the specification, then yes.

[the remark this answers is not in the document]
garydgregory: > My thoughts were the following:
> 
>     * the Javadoc is specifying the API to the users of the library
> 
>     * the tests can be more detailed, so that the common-io-developers know which concrete exception is thrown.
> 
>     * if the implementation changes, so that another exception is thrown, the developer get noticed and can consider about the consequences (e.g. writing a simple hint to the changelog).
> 
> 
> But of course, if you see the tests as the specification, then yes.

If we give ourselves more flexibility by documenting the more general exception, I do indeed feel that we should afford ourselves the same in the tests.

Using tests as specification is a bit strong but I do want the flexibility for now.

src/test/java/org/apache/commons/io/FileUtilsTest.java
chkpnt: NP, your repo, your decision ;-)

Do you want me to add the deletion in the `*_AccessDeniedExceptionOnPosixFileSystem`, too? I took my cue from the `testWriteString*` tests, where the cleanup relies on the tempDir.

src/test/java/org/apache/commons/io/FileUtilsTest.java
```diff
@@ -2463,18 +2487,56 @@ public void testReadFileToStringWithEncoding() throws Exception {
...
+
+        Exception exception = assertThrows(IOException.class, () -> FileUtils.readLines(tempDirFile));
+        assertEquals("Is a directory", exception.getMessage());
```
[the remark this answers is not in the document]
chkpnt: The message isn't coming from common-io, that's the reason why I added this assert. There is no more concrete exception class in this case, and the message would be the feature to detect unwanted changes. But yes, it's going beyond the Javadoc-specification, so I'm removing it.

src/test/java/org/apache/commons/io/FileUtilsTest.java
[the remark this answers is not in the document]
chkpnt: NP, both views are valid.

src/test/java/org/apache/commons/io/FileUtilsTest.java
```diff
@@ -2463,6 +2482,24 @@ public void testReadFileToStringWithEncoding() throws Exception {
...
+    public void testReadFileToString_Errors() {
+        assertThrows(NullPointerException.class, () -> FileUtils.readFileToString(null));
+        assertThrows(IOException.class, () -> FileUtils.readFileToString(new File("non-exsistent")));
+        assertThrows(IOException.class, () -> FileUtils.readFileToString(tempDirFile));
+        assertThrows(IOException.class, () -> FileUtils.readFileToString(tempDirFile, "unsupported-charset"));
+    }
+
+    @Test
+    @EnabledIf("isPosixFilePermissionsSupported")
+    public void testReadFileToString_IOExceptionOnPosixFileSystem() throws Exception {
```
chkpnt, line 2495: Just stumbled over

```
        if (!SystemUtils.IS_OS_WINDOWS) {
            final Set<PosixFilePermission> parentPerms = Files.getPosixFilePermissions(target.getParentFile().toPath());
            final Set<PosixFilePermission> targetPerms = Files.getPosixFilePermissions(target.toPath());
            assertEquals(parentPerms, targetPerms);
        }
```

in line 862. So do you prefer such an condition in `testReadFileToByteArray_Errors()` instead of this conditional test? (But I'm not experienced enough in regards of filesystems: Is `!SystemUtils.IS_OS_WINDOWS` enough to rely on for the usage of `PosixFilePermissions`? `isPosixFilePermissionsSupported()` might be safer?)

garydgregory: Hi @chkpnt 
I prefer fewer changes in this PR because it is supposed to be about Javadoc, not code. The use of `@EnabledIf("isPosixFilePermissionsSupported")` is fine.

chkpnt:
Upsi.... one of the last changes were wrong. :-( I'll fix the build.

PR #490, opened 2023-10-01
Add simple test for StreamIterator

labkey-adam:
Note: This test will fail until `StreamIterator.iterator()` is fixed to return the wrapper class


garydgregory:
@labkey-adam 
- Thank you for your PR.
- In the future, please run `mvn` locally to pick up failures from all build checks.

PR #478, opened 2023-09-07
`FileUtils.deleteDirectory` could throw `UncheckedIOException`

jglick:
At some point between 2.11.0 and 2.13.0, I think as of 323d376b4a934a5a6ebdc552dc923db9e267e569, there seems to have been an incompatible change due to use of `UncheckedIOException`. I found the following from a piece of code calling `FileUtils.deleteDirectory` and catching and logging `IOException`, which failed to catch the runtime exception:

```
java.nio.file.NoSuchFileException: /path/to/file
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
	at java.base/sun.nio.fs.UnixFileAttributeViews$Posix.readAttributes(UnixFileAttributeViews.java:234)
	at java.base/sun.nio.fs.UnixFileAttributeViews$Posix.readAttributes(UnixFileAttributeViews.java:147)
	at java.base/sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:149)
	at java.base/sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
	at java.base/java.nio.file.Files.readAttributes(Files.java:1764)
	at org.apache.commons.io.function.Uncheck.apply(Uncheck.java:162)
Caused: java.io.UncheckedIOException
	at org.apache.commons.io.function.Uncheck.wrap(Uncheck.java:242)
	at org.apache.commons.io.function.Uncheck.apply(Uncheck.java:164)
	at org.apache.commons.io.file.PathUtils.readAttributes(PathUtils.java:1259)
	at org.apache.commons.io.file.PathUtils.readPosixFileAttributes(PathUtils.java:1349)
	at org.apache.commons.io.file.PathUtils.deleteFile(PathUtils.java:582)
	at org.apache.commons.io.file.PathUtils.delete(PathUtils.java:476)
	at org.apache.commons.io.FileUtils.forceDelete(FileUtils.java:1337)
	at org.apache.commons.io.function.IOStream.lambda$forAll$11(IOStream.java:340)
	at java.base/java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
[...]

PR #491, opened 2023-10-06
[IO-814] Don't throw UncheckedIOException

elharo:
Fortunately the API here offered a backwards compatible fix. Since it was already declared to return null for some exceptions, this makes it return null for IOException too. This reverts the behavior breaking change without adding any API changes to anyone's code.

Also some javadoc cleanups while I was in here.

### DIFF
--- a/src/main/java/org/apache/commons/io/file/PathUtils.java
+++ b/src/main/java/org/apache/commons/io/file/PathUtils.java
@@ -1252,8 +1252,8 @@ public final class PathUtils {
      * @param <A> The {@link BasicFileAttributes} type
      * @param path The Path to test.
      * @param type the {@link Class} of the file attributes required to read.
-     * @param options options indicating how to handle symbolic links
-     * @return the file attributes or null if the attributes can't be read
+     * @param options options indicating how to handle symbolic links.
+     * @return the file attributes or null if the attributes can't be read.
      * @see Files#readAttributes(Path, Class, LinkOption...)
      * @since 2.12.0
      */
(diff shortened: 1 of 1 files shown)


garydgregory (changes requested):
Please don't change the style: Sentences start with a capital letter and end with a period.

elharo:
Except they aren't sentences. They're sentence fragments, and this case is explicitly called out in the [Oracle style guidelines](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#tag): "When writing the comments themselves, in general, start with a phrase and follow it with sentences if they are needed. When writing a phrase, do not capitalize and do not end with a period".

If the project wants to diverge from that, it's worth calling it out explicitly in a project specific style guide.

garydgregory:
> Except they aren't sentences. They're sentence fragments, and this case is explicitly called out in the [Oracle style guidelines](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#tag): "When writing the comments themselves, in general, start with a phrase and follow it with sentences if they are needed. When writing a phrase, do not capitalize and do not end with a period".
> 
> If the project wants to diverge from that, it's worth calling it out explicitly in a project specific style guide.

We do, under a header called "Respect The Original Style" on https://commons.apache.org/patches.html

I suppose it could go into more detail of course ;-)

jglick:
Would a variant of the test from #478 be useful here as a regression test? Like

```java
PathUtils.deleteFile(tempDir.resolve("nonexistent").resolve("file-does-not-exist.bin"), StandardDeleteOption.OVERRIDE_READ_ONLY));
```

which if I understand correctly should now return normally?

PR #493, opened 2023-10-08
[IO-414] Don't write a BOM on every (or any) line

elharo:
Something very weird happened with the previous iteration of this PR. This PR should work.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -71,7 +71,10 @@ The <action> type attribute can be add,update,fix,remove.
         StreamIterator fails to close its internal Stream.
       </action>
       <action dev="ggregory" type="fix" issue="IO-814" due-to="Elliotte Rusty Harold, Gary Gregory">
-        Don't throw UncheckedIOException #491. 
+        Don't throw UncheckedIOException #491.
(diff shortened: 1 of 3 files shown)


garydgregory:
Hi @elharo 
You can run `mvn` from a command shell to catch these build errors. `mvn` runs our default goal which includes all the checks.

elharo:
mvn test doesn't catch checkstyle errors. I have to look for those specifically.

On the previous iteration, the CI passed on my PR. The code was somehow modified after I submitted it.

garydgregory:
> mvn test doesn't catch checkstyle errors. I have to look for those specifically.
> 
> On the previous iteration, the CI passed on my PR. The code was somehow modified after I submitted it.

No you don't, as noted above, run `mvn`, by itself, this causes the default Maven goal to run, which includes all of the checks.

garydgregory:
Side note: In general, don't edit `changes.xml` in PRs, this usually causes conflicts if another PR also edits that file or if, invariably, that file has been edited in the course of plain old development.

As an alternative, we could use Log4j's change log mechanism which avoids any and all conflicts.

PR #495, opened 2023-10-08
Replace a number of deprecated methods

elharo:
Most;y by adding explicit charsets, except in the cases where the omission of a charset was what was being tested

### DIFF
--- a/src/test/java/org/apache/commons/io/FileUtilsTest.java
+++ b/src/test/java/org/apache/commons/io/FileUtilsTest.java
@@ -2829,25 +2829,25 @@ public class FileUtilsTest extends AbstractTempDirTest {
     @Test
     public void testWrite_WithAppendOptionFalse_ShouldDeletePreviousFileLines() throws Exception {
         final File file = TestUtils.newFile(tempDirFile, "lines.txt");
-        FileUtils.writeStringToFile(file, "This line was there before you...");
+        FileUtils.writeStringToFile(file, "This line was there before you...", StandardCharsets.UTF_8);
 
-        FileUtils.write(file, "this is brand new data", false);
+        FileUtils.write(file, "this is brand new data", StandardCharsets.UTF_8, false);
 
         final String expected = "this is brand new data";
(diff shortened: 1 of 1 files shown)


garydgregory:
-1: We have to test ALL the code we provide. Also, notice this PR _decreases_ the code coverage percentage:
```
@@             Coverage Diff              @@
##             master     #495      +/-   ##
============================================
- Coverage     84.97%   84.92%   -0.05%     
+ Complexity     3372     3370       -2     
============================================
  Files           227      227              
  Lines          8080     8080              
  Branches        953      953              
============================================
- Hits           6866     6862       -4     
- Misses          960      964       +4     
  Partials        254      254             
```

elharo:
I think in all cases here each method still tests what it tested before. Only setup and verification code has been rewritten to use non-deprecated methods and to be more platform independent. There were a couple places I left unchanged precisely because they test deprecated method. If I missed one of those, point to it and I'll fix it.

Code coverage numbers are not accurate to three significant figures.

garydgregory:
We need to keep the code as is to show it work on Java 18 and up where the default encoding is UTF-8. We should consider undeprecating these methods on Java 18 and up.

elharo:
That would be nice. Unfortunately, "An implementation may override the default charset with the system property file.encoding on the command line. If the value is COMPAT, the default charset is derived from the native.encoding system property, which typically depends upon the locale and charset of the underlying operating system." so we're going to need to specify charsets for the next 25 years. :-(

We might end up setting up separate JDK 21 CIs, one with UTF-8 and one with something weird like Cp936, to flush out the issues.

garydgregory:
> We might end up setting up separate JDK 21 CIs, one with UTF-8 and one with something weird like Cp936, to flush out the issues.


That sounds good (and interesting) but this PR should be closed.

elharo:
I need to take another look at this, but I'm not convinced the existing tests are correct or testing what they think they're testing.

garydgregory:
Closing: See above.

PR #497, opened 2023-10-12
[IO-427] Add TrailerInputStream

ax-lothas:


src/main/java/org/apache/commons/io/input/TrailerInputStream.java
```diff
@@ -0,0 +1,180 @@
...
+ * 
+ * <p>
+ * Useful, e.g., for handling checksums: payload is followed by a fixed size hash, so while
+ * streaming the payload the trailer finally contains the expected hash (this example needs
+ * extra caution to revert actions when the final checksum match fails).
+ * </p>
+ */
+public final class TrailerInputStream extends InputStream {
+
+    private final  InputStream source;
```
garydgregory, line 43: This class should extend `java.io.FilterInputStream` or  `org.apache.commons.io.input.ProxyInputStream`.

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
```diff
@@ -0,0 +1,180 @@
...
+     * 
+     * <p>
+     * Reads exactly {@code trailerLength} bytes from {@code source}.
+     * </p>
+     * 
+     * @param source underlying stream from which is read.
+     * @param trailerLength the length of the trailer which is hold back (must be &gt;= 0).
+     * @throws IOException initializing the trailer buffer failed.
+     */
+    public TrailerInputStream(final InputStream source, final int trailerLength)
```
garydgregory, line 61: Use the Builder pattern to avoid constructor creep. For example `org.apache.commons.io.input.BOMInputStream.Builder`.

garydgregory (changes requested):
Some note. I'll comment in the ticket on a separate issue.

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
ax-lothas: Extending `FilterInputStream` or `ProxyInputStream` would save overriding
* `close`, and
* `available`

but would require to override
* `mark`,
* `reset`, and
* `markSupported`.

The current implementation is incompatible with `mark`/`reset` as it doesn't track which bytes are already read and wich ones are new.

So, I don't see the benefit.

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
ax-lothas: What should be the default value of `trailerLength`? `0` is rather useless, so both arguments have to be provided anyway.

elharo: Personally I prefer constructors, especially when as here there's exactly one constructor.

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
```diff
@@ -0,0 +1,180 @@
...
+            System.arraycopy(b, off + readIntoBuffer - underflow, this.trailer, 0, underflow);
+        }
+        // IOUtils.read reads as many bytes as possible, so reading 0 bytes means EOF.
+        // Then, we have to mark this.
+        return read == 0 && len != 0 ? IOUtils.EOF : read;
+    }
+
+    @Override
+    public int available() throws IOException {
+        return this.source.available();
```
elharo, line 149: shouldn't this subtract the bytes in the trailer?

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
```diff
@@ -0,0 +1,180 @@
...
+        } finally {
+            super.close();
+        }
+    }
+
+    public int getTrailerLength() {
+        return this.trailer.length;
+    }
+
+    public byte[] copyTrailer() {
```
elharo, line 165: I think one method to getTrailer() should suffice and would be noticeably simpler. I don't expect trailers to be all that long.

elharo (changes requested):

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
ax-lothas: The trailer is filled in the constructor. So every single byte read after the constructor finished causes exactly one byte to return to the caller of the `read` methods. Think about it as shift register.

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
ax-lothas: Done.

garydgregory:
If this class is indeed generally useful, can it be reused from our Tailer classes?

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
```diff
@@ -0,0 +1,193 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import org.apache.commons.io.IOUtils;
+
+/**
+ * Reads the underlying input stream while holding back the trailer.
+ * 
+ * <p>
+ * "Normal" read calls read the underlying stream except the last few bytes (the trailer). The
+ * trailer is updated with each read call. The trailer can be gotten by one of the copyTrailer
```
elharo, line 25: getTrailer

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
```diff
@@ -0,0 +1,193 @@
...
+ * overloads.
+ * </p>
+ * 
+ * <p>
+ * It is safe to fetch the trailer at any time but the trailer will change with each read call
```
elharo, line 30: That seems wonky. I'd almost prefer iot to throw IllegalStateException unitl it has the real trailer

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
```diff
@@ -0,0 +1,193 @@
...
+
+    // The current implementation is incompatible with mark/reset as it doesn't track which bytes are
+    // already read and which ones are new. This tracking would be necessary to not overwrite the
+    // trailer with earlier bytes in the source stream. Remember that the trailer is not meant to
+    // contain the last read bytes but the last bytes in the stream (which differs when using reset
+    // to jump to an earlier position of the source stream).
+
+    private final  InputStream source;
+    /**
+     * Invariant: After every method call which exited without exception, the trailer has to be
```
elharo, line 69: has to be --> is

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * Constructs the TrailerInputStream and initializes the trailer buffer.
+     * 
+     * <p>
+     * Reads exactly {@code trailerLength} bytes from {@code source}.
+     * </p>
+     * 
+     * @param source underlying stream from which is read.
+     * @param trailerLength the length of the trailer which is hold back (must be &gt;= 0).
```
elharo, line 82: held back

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
elharo: IMHO builders are vastly overused. They're not native to Java's design, and constructors are more natural. There's a place for builders, but they shouldn't be the default.

src/main/java/org/apache/commons/io/input/TrailerInputStream.java
```diff
@@ -0,0 +1,193 @@
...
+        } finally {
+            super.close();
+        }
+    }
+
+    public int getTrailerLength() {
+        return this.trailer.length;
+    }
+
+    public byte[] copyTrailer() {
```
elharo, line 189: copyTrailer --> getTrailer

elharo (changes requested):
Getting the trailer before the stream is fully read should probably throw an illegal state exception.

garydgregory:
> Whilst the class seems to work well, I'm not convinced that there is a sufficient need for it.
> 
> Every extra class adds to the maintenance load, so there needs to be a demonstrated to make it worth the extra effort.

Right, I agree, and as I mentioned earlier: Could this be reused as is within IO itself (Tailer class) and from Tika which provides a similar class?

PR #500, opened 2023-10-14
adds new method to query the host OS for the temp directory path

david-patchfox:
adds new method `FileUtils.getTempDirectoryPathConsistent(boolean)` to give developers the ability to ensure the system temp path is reported in a consistent manner across all OS's. 

At issue is the fact that the underlying call to `System.getProperty("java.io.tmpdir")` does not return results in a consistent format. Depending on the host OS, the system file separator may or may not be included at the tail of the result. If the developer is unaware of this behavior their code may break in unexpected ways when ported to an OS that behaves differently than the one they wrote the code on. See these bug reports for details and background: 

https://bugs.java.com/bugdatabase/view_bug?bug_id=4391434
https://bugs.java.com/bugdatabase/view_bug?bug_id=JDK-8318067

PR does not change behavior of existing `FileUtils.getTempDirectoryPath()` but does add a docstring comment alerting developers of the issue and directing them to `FileUtils.getTempDirectoryPathConsistent(boolean)` if need to ensure the system temp directory is reported in a consistent manner across all host OS's. 

New method is tested by way of `FileUtilsTest.testGetTempDirectoryPathConsistentNoTrailingSeparator()` and `FileUtilesTest.testGetTempDirectoryPathConsistentWithTrailingSeparator()`.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -1444,7 +1444,7 @@ public class FileUtils {
 
     /**
      * Returns a {@link File} representing the system temporary directory.
-     *
+     * 
      * @return the system temporary directory.
      * @since 2.0
      */
@@ -1453,8 +1453,38 @@ public class FileUtils {
     }
 
     /**
-     * Returns the path to the system temporary directory.
+     * Returns the path to the system temporary directory with the option to include or disinclude the trailing 
+     * file separator. 
      *
+     * @param includeTrailingSeparator switch indicating whether or not you want the returned path to include a 
+     * trailing file separator. 
+     * @return the path to the system temporary directory with or without the trailing file separator as per 
+     * caller supplied argument.
+     * @since 2.14.1
+     */
+    public static String getTempDirectoryPathConsistent(boolean includeTrailingSeparator) {
+        String tempDir = System.getProperty("java.io.tmpdir");
+        String tempDirLastChar = tempDir.substring(tempDir.length() - 1);
+        String systemFileSeparator = System.getProperty("file.separator");
+        if (includeTrailingSeparator) {
+            tempDir = tempDirLastChar.equals(systemFileSeparator) 
+                            ? tempDir 
+                            : tempDir + systemFileSeparator;
+        } else {
+            tempDir = tempDirLastChar.equals(systemFileSeparator) 
+                            ? tempDir.substring(0, tempDir.length() - 1) 
+                            : tempDir;
+        }
+
+        return tempDir;
+    }
+
+    /**
+     * Returns the path to the system temporary directory.
+     * 
+     * @apiNote this method relies on Java system property 'java.io.tmpdir' which may or may not return a path with 
+     * a trailing file separator, depending on the OS Java is running in. If you need the temp directory reported in a
+     * consistent manner across all OS environments, use {@link #getTempDirectoryPathConsistent(boolean)}.
      * @return the path to the system temporary directory.
      * @since 2.0
      */
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -1453,7 +1453,7 @@ public class FileUtils {
     }
 
     /**
-     * Returns the path to the system temporary directory with the option to include or disinclude the trailing 
+     * Returns the path to the system temporary directory with the option to include or exclude the trailing 
      * file separator. 
      *
      * @param includeTrailingSeparator switch indicating whether or not you want the returned path to include a 
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -1444,7 +1444,7 @@ public class FileUtils {
 
     /**
      * Returns a {@link File} representing the system temporary directory.
-     * 
+     *
      * @return the system temporary directory.
      * @since 2.0
      */
@@ -1483,7 +1483,7 @@ public class FileUtils {
      * Returns the path to the system temporary directory.
      * 
      * @apiNote this method relies on Java system property 'java.io.tmpdir' which may or may not return a path with 
-     * a trailing file separator, depending on the OS Java is running in. If you need the temp directory reported in a
+     * a trailing file separator, depending on the OS environment. If you need the temp directory reported in a
      * consistent manner across all OS environments, use {@link #getTempDirectoryPathConsistent(boolean)}.
      * @return the path to the system temporary directory.
      * @since 2.0
--- a/src/test/java/org/apache/commons/io/FileUtilsTest.java
+++ b/src/test/java/org/apache/commons/io/FileUtilsTest.java
@@ -1587,7 +1587,7 @@ public class FileUtilsTest extends AbstractTempDirTest {
     public void testGetTempDirectory() {
         final File tempDirectory = new File(FileUtils.getTempDirectoryPath());
         assertEquals(tempDirectory, FileUtils.getTempDirectory());
-    }  
+    } 
 
     @Test
     public void testGetTempDirectoryPath() {
--- a/src/test/java/org/apache/commons/io/FileUtilsTest.java
+++ b/src/test/java/org/apache/commons/io/FileUtilsTest.java
@@ -1598,18 +1598,38 @@ public class FileUtilsTest extends AbstractTempDirTest {
     public void testGetTempDirectoryPathConsistentNoTrailingSeparator() {
         String tempDirectoryPath = FileUtils.getTempDirectoryPathConsistent(false);
         assertTrue(tempDirectoryPath.startsWith(System.getProperty("java.io.tmpdir")));
+
         String systemFileSeparator = System.getProperty("file.separator");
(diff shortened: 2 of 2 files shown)


garydgregory:
Path methods handle separators, so one should use those APIs these days...

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -1452,9 +1452,39 @@ public static File getTempDirectory() {
...
+     * @param includeTrailingSeparator switch indicating whether or not you want the returned path to include a 
+     * trailing file separator. 
+     * @return the path to the system temporary directory with or without the trailing file separator as per 
+     * caller supplied argument.
+     * @since 2.14.1
+     */
+    public static String getTempDirectoryPathConsistent(boolean includeTrailingSeparator) {
+        String tempDir = System.getProperty("java.io.tmpdir");
+        String tempDirLastChar = tempDir.substring(tempDir.length() - 1);
+        String systemFileSeparator = System.getProperty("file.separator");
```
robtimus, line 1468: Just a thought, but is it possible to use `tempDir.charAt(tempDir.length() - 1)` and `File.separatorChar`?

david-patchfox: what do you mean? you want  ln1468 changed to use `File.separatorChar` ? what would be the difference?

robtimus: It would use `char` instead of `String` for the comparison, which would save some memory. Perhaps it's premature optimization, but whenever I can use primitives instead of objects I go for it.

garydgregory: > what do you mean? you want ln1468 changed to use `File.separatorChar` ? what would be the difference?

It's better to use an API than a magic string.

david-patchfox: I agree - that feels like a premature optimization :smile:

src/test/java/org/apache/commons/io/FileUtilsTest.java
```diff
@@ -1594,6 +1594,22 @@ public void testGetTempDirectoryPath() {
         assertEquals(System.getProperty("java.io.tmpdir"), FileUtils.getTempDirectoryPath());
     }
 
+    @Test
+    public void testGetTempDirectoryPathConsistentNoTrailingSeparator() {
+        String tempDirectoryPath = FileUtils.getTempDirectoryPathConsistent(false);
```
garydgregory, line 1599: This test is not valid: It does not assert the actual path is correct. For example, if the new API returns a "C:\Some\Odd\Dir" that is not the Java temp dir, then the test would pass.

garydgregory:
I still don't see why this is needed. Just call: `org.apache.commons.io.file.PathUtils.getTempDirectory()`.

src/test/java/org/apache/commons/io/FileUtilsTest.java
david-patchfox: eh ok. I think that's overkill but I can change that. wait one and I'll push an update to the test

david-patchfox: done and done. I've added assertions to both tests that ensure the root path is the same as is reported by property `java.io.tmpdir`

david-patchfox:
> I still don't see why this is needed. Just call: `org.apache.commons.io.file.PathUtils.getTempDirectory()`.

Bottom line - y'all expose a public method to return this as a String and  there's presently nothing in the API doc warning developers of the inconsistent nature of how the underlying java system property will report the temp directory in that circumstance. This does that and provides an alternative method for folks - like us - don't want to diddle with Path objects because doing it the Path way creates bloat code that doesn't need to be there.

david-patchfox:
> I still don't see why this is needed. Just call: `org.apache.commons.io.file.PathUtils.getTempDirectory()`.

actually one more thing as I'm getting the impression you still aren't fathoming why someone would do something differently than you think they should  - here is an example: 

the Path API doesn't permit you to split in the same way a String does. The system tmp folder is named differently depending on what OS the JVM is running on and is therefore of variable path length (macos in particular). If you are doing ETL work with archived input you need to know where certain relative paths are indexed against system tmp it takes several extra statements of code to do that the Path way vs splitting a path string by whatever `java.io.tmpdir` reports as then using the pre-determined relative indexing the ETL packaging scheme is using. Java is already a highly verbose language and not everyone enjoys making a factory factory for the session helper manager to generation a session manager helper to do what ostensibly are basic tasks. 
 
 In other words - sometimes some people have a good reason to want to use Strings and not Path and that's ok  :smiley:

src/test/java/org/apache/commons/io/FileUtilsTest.java
```diff
@@ -1594,6 +1594,24 @@ public void testGetTempDirectoryPath() {
...
+        assertTrue(tempDirectoryPath.startsWith(System.getProperty("java.io.tmpdir")));
+        String systemFileSeparator = System.getProperty("file.separator");
+        String lastCharOfTempDirectoryPath = tempDirectoryPath.substring(tempDirectoryPath.length() - 1);
+        assertNotEquals(lastCharOfTempDirectoryPath, systemFileSeparator);
```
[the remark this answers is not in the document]
david-patchfox: huh? the method being tested is pulling the system temp directory in exactly the same way as the existing `FileUtils.getTempDirectoryPath()` method and that's by calling `System.getProperty("java.io.tmpdir")` which gets populated during phase1init by way of SystemProps.java ([link](https://github.com/openjdk/jdk/blob/e25a49a993f270c33f7929e629fb3075a11fdec9/src/java.base/share/classes/jdk/internal/util/SystemProps.java#L113C72-L113C72)) . 

what's the reasonable concern here?

src/test/java/org/apache/commons/io/FileUtilsTest.java
```diff
@@ -1594,6 +1594,24 @@ public void testGetTempDirectoryPath() {
...
+    }
+
+    @Test
+    public void testGetTempDirectoryPathConsistentWithTrailingSeparator() {
+        String tempDirectoryPath = FileUtils.getTempDirectoryPathConsistent(true);
+        assertTrue(tempDirectoryPath.startsWith(System.getProperty("java.io.tmpdir")));
```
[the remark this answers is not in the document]
david-patchfox: same as above - what's the reasonable concern here?

garydgregory:
> I agree that it would be helpful if existing method Javadoc noted that the return string may or may not include a trailing separator.
> 
> However, adding more code and test cases (which have to be maintained) seems completely unnecessary to me. Indeed counter-productive, as it encourages string manipulation rather than proper use of the appropriate routines.
> 
> I think we should just update the Javadoc and suggest coders use the proper methods.

I agree wholeheartedly with @sebbASF, I am -1 to this PR as it stands.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -1452,9 +1452,39 @@ public static File getTempDirectory() {
         return new File(getTempDirectoryPath());
     }
 
+    /**
+     * Returns the path to the system temporary directory with the option to include or disinclude the trailing
```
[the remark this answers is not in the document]
david-patchfox: done

david-patchfox:
> I agree that it would be helpful if existing method Javadoc noted that the return string may or may not include a trailing separator.
> 
> However, adding more code and test cases (which have to be maintained) seems completely unnecessary to me. Indeed counter-productive, as it encourages string manipulation rather than proper use of the appropriate routines.
> 
> I think we should just update the Javadoc and suggest coders use the proper methods.

Respectfully - y'all seem fixated on telling people what "proper" is and as tool makers I think that's wrong headed. Not everyone is going to limit themselves in how they use what you've made to the limit of your imagination of how things can or should be done. It's a big world and none of us - not y'all and not me - can possibly enumerate every permutation of every circumstance in which a tool needs to be employed. In fact - I would argue exploring the delta between what was intended and what's possible is the fun and delight-  and dare I say the point - of coding. 

That being said - this is your tool and not mine. We had a problem and fixed it on our end and I put in a PR to be a mensch to the community by sharing that additional logic. Take it or leave it. I'm happy to continue putting in PR changes as per request but I have to say I'm more than a little surprised at the level of passionate pushback y'all have given me to the idea that someone might (gasp) be using public methods in your API that have been there for years in the manner in which the documentation says is intended. 

:beers:  friends

david-patchfox:
> > I agree that it would be helpful if existing method Javadoc noted that the return string may or may not include a trailing separator.
> > However, adding more code and test cases (which have to be maintained) seems completely unnecessary to me. Indeed counter-productive, as it encourages string manipulation rather than proper use of the appropriate routines.
> > I think we should just update the Javadoc and suggest coders use the proper methods.
> 
> I agree wholeheartedly with @sebbASF, I am -1 to this PR as it stand

> > I agree that it would be helpful if existing method Javadoc noted that the return string may or may not include a trailing separator.
> > However, adding more code and test cases (which have to be maintained) seems completely unnecessary to me. Indeed counter-productive, as it encourages string manipulation rather than proper use of the appropriate routines.
> > I think we should just update the Javadoc and suggest coders use the proper methods.
> 
> I agree wholeheartedly with @sebbASF, I am -1 to this PR as it stands.

ok cool. :beers:

PR #509, opened 2023-11-06
Add test for FileChannels.contentEquals.

smarkwal:
The test in this PR uses `FileChannels.contentEquals` to compare the content of two files. Both files have the same size of 1027 bytes. While the content of file 1 starts with "ABC", the content of file 2 starts with "XYZ". Both files then contain additional identical 1024 bytes (1024 x character "x"). The two files should be considered having different content.

If `FileChannels.contentEquals` is invoked with a buffer size of 1024 bytes, it will return `true` (content is equal) instead of the expected `false` (content is different).

If calling the method with a buffer size of 1023 bytes (just 1 byte less), then the method will return the correct result `false`.

### DIFF
--- a/src/test/java/org/apache/commons/io/channels/FileChannelsTest.java
+++ b/src/test/java/org/apache/commons/io/channels/FileChannelsTest.java
@@ -1,3 +1,19 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


garydgregory (changes requested):
Hello @smarkwal 
Thank you for the test. 
Would you please add the Apache license header to the new file?

smarkwal:
Hi @garydgregory 

I have added the license header and also applied a code style that seems to be used in this project. I was not able to find an `.editorconfig` file or similar. But I found [this CodeStyle wiki page](https://cwiki.apache.org/confluence/display/commons/CodeStyle) and looked at some other source files for clues.

garydgregory:
Hello @smarkwal,

Thank you for your update. Please test git master or a build from the Apache Maven snapshot repository. 

I did not merge this PR since it would fail the build; instead, I expanded the test, and committed a fix at the same time. 

You are credited in `changes.xml`.

smarkwal:
Hi @garydgregory,

Thank you very much for the quick fix and improving the test. I have been able to use 2.15.1-SNAPSHOT without problems in my project. I also did some more extensive testing and the new implementation has worked perfectly. I will close this PR.

Best,
Ste

garydgregory:
👍

PR #525, opened 2023-12-17
[IO-781] Make CharSequenceInputStream.available() more correct in the face of multibyte encodings

elharo:
Per contract, available() is not required to return the exact or maximum number of bytes available without blocking. It is required to return a number less than or equal to that. Always returning 0 is an acceptable implementation (though here we can do better than that). 

@garydgregory 

FYI locally the build fails on JDK 8 with an apparently unrelated error:

[ERROR] /Users/elharo/commons-io/src/test/java/org/apache/commons/io/file/FilesUncheckTest.java:[99,25] cannot access org.apache.commons.io.file.FilesUncheck
  bad class file: /Users/elharo/commons-io/target/classes/org/apache/commons/io/file/FilesUncheck.class
    undeclared type variable: A
    Please remove or make sure it appears in the correct subdirectory of the classpath.

Let's see what the CI does with this.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/CharSequenceInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/CharSequenceInputStream.java
@@ -210,18 +210,14 @@ public class CharSequenceInputStream extends InputStream {
     }
 
     /**
-     * Return an estimate of the number of bytes remaining in the byte stream.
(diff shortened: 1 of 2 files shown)


garydgregory:
@elharo TY!

PR #527, opened 2023-12-19
[IO-808] better handling when source is unexpectedly deleted during move

elharo:
@garydgregory I think this is roughly what's required to address IO-808. Inlining the various validations makes the flow much more obvious and avoids double null checks.

There's no way I know to lock an file so it can't be deleted after we check that it exists and before we move it, but given that we can distinguish between source files that don't exist and ones that are deleted later. The key is to do the exists check after the file type check. Not obvious in the PR, but this is the reverse of how the checks were performed previously.

Now if the source file is a directory an IllegalArgumentException is still thrown. However the new feature is that in all cases where the method fails because the source file doesn't exist an IOException is thrown, no matter whether the source file was deleted while the method was executing.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -292,7 +292,7 @@ public class FileUtils {
      * @return the checksum specified, updated with the content of the file
      * @throws NullPointerException if the given {@link File} is {@code null}.
      * @throws NullPointerException if the given {@link Checksum} is {@code null}.
-     * @throws IllegalArgumentException if the given {@link File} does not exist or is not a file.
+     * @throws IllegalArgumentException if the given {@link File} is not a file.
      * @throws IOException if an IO error occurs reading the file.
      * @since 1.3
      */
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -2393,9 +2393,17 @@ public class FileUtils {
      * @since 2.9.0
      */
     public static void moveFile(final File srcFile, final File destFile, final CopyOption... copyOptions) throws IOException {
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -2393,9 +2393,17 @@ public static void moveFile(final File srcFile, final File destFile) throws IOEx
      * @since 2.9.0
      */
     public static void moveFile(final File srcFile, final File destFile, final CopyOption... copyOptions) throws IOException {
-        validateMoveParameters(srcFile, destFile);
-        requireFile(srcFile, "srcFile");
-        requireAbsent(destFile, "destFile");
+        Objects.requireNonNull(srcFile, "source");
+        Objects.requireNonNull(destFile, "destination");
+        if (!srcFile.isFile()) {
+            if (srcFile.exists()) {
```
elharo, line 2399: Alternatively this logic could move into the private requireFile method, which would then distinguish between source files that are directories and files that don't exist, and throw a different exception in each case.

garydgregory: I think I'm OK with that if that means the code would be reused from different APIs and those would therefore get consistent behavior which we can more simply document.

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -2393,9 +2393,17 @@ public static void moveFile(final File srcFile, final File destFile) throws IOEx
      * @since 2.9.0
```
elharo, line 2393: Either way this needs a @throws IllegalArgumentException here for the case where the srcFile exists but is a directory, unless we want that to be an IOException too. And the method's doc comment should be clearer that directories are not supported. Unless maybe we want to delegate that case to moveDirectory and not throw any exception at all?

PR #528, opened 2023-12-19
[IO-826] Add runtime exception support to broken streams

markslater:
While implementing this, I came across a few minor and cosmetic issues.  I've addressed those in a couple of subsequent commits with the intention of making each easier to review.

The commits are:
1. The implementation of the feature:
    1. Because of runtime type erasure, I couldn't overload the `public BrokenX(final Supplier<IOException> exceptionSupplier)` constructors with `public BrokenX(final Supplier<RuntimeException> exceptionSupplier)`.  I've added factory methods instead, e.g . `public static BrokenReader brokenReader(final Supplier<RuntimeException> exceptionSupplier)`.
    2. All the tests have acquired duplicates that vary only in the type of exception they throw.  This *could* be achieved with parameterized tests, but it seemed more readable to write them out explicitly.
3. I removed the `synchronized` modifier from a couple of methods, because it appeared to have been added in error.
4. I consistified the JavaDoc and testing across all four BrokenX classes.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/BrokenInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/BrokenInputStream.java
@@ -142,7 +142,7 @@ public class BrokenInputStream extends InputStream {
      * @throws IOException as configured.
      */
     @Override
-    public synchronized void reset() throws IOException {
+    public void reset() throws IOException {
         exceptionThrower.doThrow();
     }
 
--- a/src/main/java/org/apache/commons/io/input/BrokenInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/BrokenInputStream.java
@@ -21,7 +21,7 @@ import java.io.InputStream;
 import java.util.function.Supplier;
 
 /**
- * Always throws an {@link IOException} from all the {@link InputStream} methods where the exception is declared.
+ * Always throws an {@link IOException} or a {@link RuntimeException} from all the {@link InputStream} methods where {@link IOException} is declared.
  * <p>
  * This class is mostly useful for testing error handling.
  * </p>
@@ -38,9 +38,26 @@ public class BrokenInputStream extends InputStream {
     public static final BrokenInputStream INSTANCE = new BrokenInputStream();
 
     /**
-     * A supplier for the exception that is thrown by all methods of this class.
+     * Constructs a new stream that always throws a {@link RuntimeException}.
+     *
+     * @param exceptionSupplier a supplier for the exception to be thrown.
+     * @return a new stream that always throws a {@link RuntimeException}.
      */
-    private final Supplier<IOException> exceptionSupplier;
+    public static BrokenInputStream brokenInputStream(final Supplier<RuntimeException> exceptionSupplier) {
+        return new BrokenInputStream(() -> {
+            throw exceptionSupplier.get();
+        });
+    }
+
+    @FunctionalInterface
+    private interface ExceptionThrower {
+        void doThrow() throws IOException;
+    }
+
+    /**
+     * A function that throws the exception that is thrown by all methods of this class.
+     */
+    private final ExceptionThrower exceptionThrower;
 
     /**
      * Constructs a new stream that always throws an {@link IOException}.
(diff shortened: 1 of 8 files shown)


garydgregory:
Hi @markslater 

Thank you for your PR! 

I really like the idea of providing a runtime exception to the "broken" streams, readers, and writers, but I think the implementation should be different. Please see the update for `BrokenInputStream` in git master. There is a new constructor and another with an updated generic parameter. This feels like a simpler take on this feature. If you want to follow this template, feel free to rebase and update this PR (or create a new one) with changes for the other classes, otherwise, I'll get to it later (within a week or so). You are credited in `changes.xml`. Note: The code and tests also work with `Error`, its subclasses, and any exception (since this lets us model all manner of "broken" streams).

markslater:
Thanks for the feedback, @garydgregory!  I'll submit a new PR later today.

One question: Are you OK with it being possible to throw checked exceptions other than `IOException`?  For example, I can write `new BrokenInputStream(new InterruptedException()).read();` to get `BrokenInputStream` to throw `InterruptedException`.  It seems to come down to balancing:
- Accepting `Throwable` in the constructors accommodates `RuntimeExceptions` without doubling the number of constructors, and avoiding the type erasure problem with the `Supplier`-based constructor.
- Additionally allows errors to be thrown (never needed to do this myself, but maybe useful for some people).
- Means checked exceptions can be thrown that wouldn't be permitted by the compiler under normal circumstances.
- Possibly it's a little harder for users of the class to infer the types of exception that can be passed to the constructors to use the class 'correctly'.

Let me know what you think - ultimately it's down to your judgement and preferences.

elharo (changes requested):
I strongly disagree with this. IOExceptions are checked for good reason. They should not be converted into runtime exceptions.

markslater:
@elharo could you expand a little?  I agree with your statement that IOExceptions should not be converted into runtime exceptions, but could you clarify the implications for the proposed change?

garydgregory:
@markslater 
I think the version in git master is fine as it is, it sounds like you did not look.

@elharo 
Keep in mind that this is a "broken" stream and IMO should emulate breaking this in all possible manners. The current version even supports throwing an Error.

markslater:
@garydgregory thanks for confirming.  I did look; I just wanted to double check with you.

markslater:
Agreed.  Looks like we're on the same page!

I'm happy to use your implementation of `BrokenInputStream` as a template for implementing the other classes: if nothing else, aligning with your approach as lead contributor is valuable for the consistency of the project.

Having said that, I'm open to taking a different approach if the discussion @elharo initiated leads us to a different conclusion.

src/main/java/org/apache/commons/io/input/BrokenInputStream.java
```diff
@@ -21,26 +21,43 @@
...
-     * The singleton instance.
+     * A singleton instance using a default IOException.
      *
      * @since 2.12.0
      */
     public static final BrokenInputStream INSTANCE = new BrokenInputStream();
 
     /**
-     * A supplier for the exception that is thrown by all methods of this class.
+     * Constructs a new stream that always throws a {@link RuntimeException}.
```
elharo, line 41: For testing error handling, it's useful to have a fake like this that does something a real stream might do, e.g. throwing an IOException on read.

It's not so useful to have a mock that's doing something no real stream should be doing like throwing a NullPointerException on read.

markslater: The case that motivated me to raise this was testing code that is passed an arbitrary implementation of `Reader`.  

`Reader`s that throw `NullPointerException` on read might go against the intention of checked exceptions, but can exist, both intentionally and inadvertently.  Some methods on `Reader` also throw `IllegalArgumentException` and `UnsupportedOperationException`, though I concede that those are more the result of client errors than the operation of the `Reader` itself.

src/main/java/org/apache/commons/io/input/BrokenInputStream.java
```diff
@@ -21,26 +21,43 @@
...
      */
-    private final Supplier<IOException> exceptionSupplier;
+    public static BrokenInputStream brokenInputStream(final Supplier<RuntimeException> exceptionSupplier) {
+        return new BrokenInputStream(() -> {
+            throw exceptionSupplier.get();
+        });
+    }
+
+    @FunctionalInterface
+    private interface ExceptionThrower {
```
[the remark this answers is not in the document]
markslater: Noted.  That line doesn't exist in your reimplementation on master.

src/main/java/org/apache/commons/io/input/BrokenInputStream.java
markslater: Actually, I need to correct myself.  `java.io.Reader` itself never throws `UnsupportedOperationException`.  The only runtime exceptions its methods throw are `IllegalArgumentException` and `NullPointerException`, during validation of arguments.

markslater:
Should I close this in favour of #528?  I'm guessing if it's closed, no one will be able to comment anymore.  Are we OK moving any remaining discussion to #528?

garydgregory:
OK with me to close.

markslater:
Closed in favour of #530

PR #529, opened 2023-12-20
[IO-808] Rationalize and unify checking for existence of files and directories

elharo:
The existing code is a bit of a hodge-podge. This PR moves most current IllegalArgumentExceptions for non-existent files and directories to FileNotFoundExceptions instead. A few turned into UncheckedIOExceptions to maintain API compatibility. The general principle is:

* If a file or directory is not found, throw a FileNotFoundException (or UncheckedIOException if necessary for backwards compatibility)
* if a file is found where a directory is expected or vice versa, throw an IllegalArgumentException

IllegalArgumentExceptions are still thrown in the case where a file object exists but is a file instead of the expected directory or a directory instead of the expected file. I've come to understand that this is wrong too, but this is existing behavior that can be improved in a subsequent PR.

In addition to correctly reporting missing files, this PR removes a number of duplicate checks and methods, and trues to clarify the names of private methods, e.g. by distinguishing whether we're checking for a file, directory, or file object. 

Some further changes that should happen here would break the API and require a major version update. 

@garydgregory

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -292,13 +292,12 @@ public class FileUtils {
      * @return the checksum specified, updated with the content of the file
      * @throws NullPointerException if the given {@link File} is {@code null}.
      * @throws NullPointerException if the given {@link Checksum} is {@code null}.
-     * @throws IllegalArgumentException if the given {@link File} does not exist or is not a file.
+     * @throws IllegalArgumentException if the given {@link File} is not a file.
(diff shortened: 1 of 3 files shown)


garydgregory:
@elharo 
Please run 'mvn' locally before you push to avoid all these failures.

elharo:
It doesn't hurt to add more detailed exception documentation. If there are any public methods where FileNotFoundException is thrown and neither FileNotFoundException nor IOException is documented, then that should be fixed. FileNotFoundException is a subclass of IOException so it's not absolutely necessary to document it separately though it can be.

PR #530, opened 2023-12-20
[IO-826] Reimplemented add runtime exception support to broken streams

markslater:
As discussed in #528.

One point to note: Is there a reason to keep the deprecated constructors such as `BrokenInputStream(final IOException exception)`?  Given that the new constructors such as `BrokenInputStream(final Throwable exception)` accept a superclass of `IOException`, it seems the old constructor could be removed without a breaking change.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/BrokenInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/BrokenInputStream.java
@@ -120,7 +120,7 @@ public class BrokenInputStream extends InputStream {
      * @throws IOException as configured.
      */
     @Override
-    public synchronized void reset() throws IOException {
+    public void reset() throws IOException {
         throw rethrow();
     }
 
--- a/src/main/java/org/apache/commons/io/input/BrokenInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/BrokenInputStream.java
@@ -33,7 +33,7 @@ import org.apache.commons.io.function.Erase;
 public class BrokenInputStream extends InputStream {
 
     /**
-     * The singleton instance.
+     * A singleton instance using a default IOException.
      *
      * @since 2.12.0
      */
@@ -85,7 +85,7 @@ public class BrokenInputStream extends InputStream {
     /**
      * Throws the configured exception.
      *
-     * @return nothing
+     * @return nothing.
      * @throws IOException as configured.
      */
     @Override
@@ -106,7 +106,7 @@ public class BrokenInputStream extends InputStream {
     /**
      * Throws the configured exception.
      *
-     * @return nothing
+     * @return nothing.
      * @throws IOException as configured.
      */
     @Override
@@ -136,8 +136,8 @@ public class BrokenInputStream extends InputStream {
     /**
      * Throws the configured exception.
      *
-     * @param n ignored
-     * @return nothing
+     * @param n ignored.
+     * @return nothing.
      * @throws IOException as configured.
      */
     @Override
--- a/src/main/java/org/apache/commons/io/input/BrokenInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/BrokenInputStream.java
@@ -33,7 +33,7 @@ import org.apache.commons.io.function.Erase;
 public class BrokenInputStream extends InputStream {
 
     /**
-     * A singleton instance using a default IOException.
+     * The singleton instance using a default IOException.
      *
      * @since 2.12.0
      */
--- a/src/main/java/org/apache/commons/io/input/BrokenInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/BrokenInputStream.java
@@ -86,7 +86,7 @@ public class BrokenInputStream extends InputStream {
      * Throws the configured exception.
      *
      * @return nothing.
-     * @throws IOException as configured.
+     * @throws IOException always throws the exception configured in the constructor.
      */
     @Override
     public int available() throws IOException {
@@ -96,7 +96,7 @@ public class BrokenInputStream extends InputStream {
     /**
      * Throws the configured exception.
      *
-     * @throws IOException as configured.
+     * @throws IOException always throws the exception configured in the constructor.
      */
     @Override
     public void close() throws IOException {
@@ -107,7 +107,7 @@ public class BrokenInputStream extends InputStream {
      * Throws the configured exception.
      *
      * @return nothing.
-     * @throws IOException as configured.
+     * @throws IOException always throws the exception configured in the constructor.
      */
     @Override
     public int read() throws IOException {
@@ -117,7 +117,7 @@ public class BrokenInputStream extends InputStream {
     /**
      * Throws the configured exception.
      *
-     * @throws IOException as configured.
+     * @throws IOException always throws the exception configured in the constructor.
      */
     @Override
     public void reset() throws IOException {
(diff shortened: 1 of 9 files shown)


garydgregory:
While this may not break source compatibility, it would break binary compatibility per https://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html#jls-13.4.12 (at least how I read it and how japicmp sees it).

markslater:
Yes, you're absolutely right.  I hadn't thought of that.

src/main/java/org/apache/commons/io/input/BrokenInputStream.java
```diff
@@ -117,10 +117,10 @@ public int read() throws IOException {
     /**
      * Throws the configured exception.
      *
-     * @throws IOException always thrown
+     * @throws IOException as configured.
      */
     @Override
-    public synchronized void reset() throws IOException {
+    public void reset() throws IOException {
```
garydgregory, line 123: Keep `synchronized`, there is no reason to break the superclass' specification here.

src/main/java/org/apache/commons/io/input/BrokenInputStream.java
```diff
@@ -136,9 +136,9 @@ private RuntimeException rethrow() {
     /**
      * Throws the configured exception.
      *
-     * @param n ignored
-     * @return nothing
-     * @throws IOException always thrown
+     * @param n ignored.
+     * @return nothing.
+     * @throws IOException as configured.
```
garydgregory, line 141: This is confusing compared to the previous Javadoc IMO because it is in fact the case that _an_ exception is _always_ thrown. You could say instead "Always throws the exception supplied from a constructor."

src/test/java/org/apache/commons/io/input/BrokenInputStreamTest.java
```diff
@@ -34,7 +35,7 @@
  */
 public class BrokenInputStreamTest {
 
-    static final class CustomeException extends Exception {
+    static final class CustomException extends Exception {
```
garydgregory, line 38: Good catch. There is no need to duplicate exception classes, just extract this class into its own file and reuse in all tests.

src/test/java/org/apache/commons/io/input/BrokenReaderTest.java
```diff
@@ -20,70 +20,125 @@
...
-    @BeforeEach
-    public void setUp() {
-        exception = new IOException("test exception");
-        brokenReader = new BrokenReader(exception);
     }
 
-    @Test
-    public void testClose() {
-        assertEquals(exception, assertThrows(IOException.class, () -> brokenReader.close()));
+    static Stream<Class<? extends Throwable>> parameters() {
```
garydgregory, line 44: Let's not duplicate this factory method all over the place. Extract this method in a new class `TestFixtures' or some such name.

garydgregory (changes requested):
@markslater 
TY for your update. 
See my few comments.

src/main/java/org/apache/commons/io/input/BrokenInputStream.java
markslater: Done

src/main/java/org/apache/commons/io/input/BrokenInputStream.java
```diff
@@ -117,10 +117,10 @@ public int read() throws IOException {
     /**
      * Throws the configured exception.
      *
-     * @throws IOException always thrown
+     * @throws IOException always throws the exception configured in the constructor.
```
garydgregory, line 120: Since there is more than one constructor, I think we should say "in a constructor", not in "in the...".

src/test/java/org/apache/commons/io/input/BrokenInputStreamTest.java
markslater: Done.

src/main/java/org/apache/commons/io/input/BrokenInputStream.java
markslater: Could you clarify?  I can't see any mention of `synchronized` [in the javadoc](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/io/InputStream.html#reset()).

src/main/java/org/apache/commons/io/input/BrokenInputStream.java
markslater: As you like - I read this as 'in the constructor [you callled]', but I don't have a strong feeling either way.

markslater: How about "throws the exception configured on construction"?

garydgregory: > How about "always throws the exception configured on construction"?

I'm not going to go round and round on such a small point. I'll likely review post merge and resolve whatever is left.

src/main/java/org/apache/commons/io/input/BrokenInputStream.java
garydgregory: Uh? Javadoc doesn't document `synchronized` keywords?

markslater: I didn't explain myself very well.  I meant to say I don't see any mention of locking in the javadoc.

I dug a bit deeper and it seems the synchronization of the `reset` method varies by JDK.  Here's a summary of which JDKs mark it as synchronized, sampled from the ones I had to hand:
|   | java.io.InputStream#reset | java.io.Reader#reset |
| - | ---------------------------------- | ---------------------------- |
| openjdk-7 | [:white_check_mark:](https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/io/InputStream.java#L346) | [:x:](https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/io/Reader.java#L248) | 
| corretto-1.8.0_392 | [:white_check_mark:](https://github.com/corretto/corretto-8/blob/8.392.08.1/jdk/src/share/classes/java/io/InputStream.java#L347) | [:x:](https://github.com/corretto/corretto-8/blob/8.392.08.1/jdk/src/share/classes/java/io/Reader.java#L248) |
| temurin-11.0.21 | :white_check_mark: | :x: |
| corretto-20.0.2.10 | [:x:](https://github.com/corretto/corretto-20/blob/20.0.2.10.1/src/java.base/share/classes/java/io/InputStream.java#L737) | [:x:](https://github.com/corretto/corretto-20/blob/20.0.2.10.1/src/java.base/share/classes/java/io/Reader.java#L398) |
| coretto-21.0.1.12 | [:x:](https://github.com/corretto/corretto-21/blob/21.0.1.12.1/src/java.base/share/classes/java/io/InputStream.java#L740) | [:x:](https://github.com/corretto/corretto-21/blob/21.0.1.12.1/src/java.base/share/classes/java/io/Reader.java#L398) |

AFAIK `java.io.Reader#reset` has never been synchronized, so that one is easy.  For `java.io.InputStream#reset`, I guess it should be synchronized to line up with Java 8, since that's the library's targeted version?

garydgregory: Well, this PR is about a new feature and should not flip-flop an unrelated implementation detail IMO. Especially when the setting seems to change between vendor.

garydgregory: FWIW, I see synchronized on InputStream Temurin 17 but not Temurin 21.

garydgregory:
@markslater PR merged, TY!

PR #533, opened 2023-12-21
[IO-808] remove redundant checks; improve Javadoc

elharo:
There were a lot of code paths that checked the same condition— file is not null, file exists, file is/is not a directory — multiple times. I've cleaned this up as best I can. 

@garydgregory

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -1332,10 +1332,15 @@ public class FileUtils {
      */
     public static void forceDelete(final File file) throws IOException {
         Objects.requireNonNull(file, "file");
-        final Counters.PathCounters deleteCounters = PathUtils.delete(
-                file.toPath(), PathUtils.EMPTY_LINK_OPTION_ARRAY,
-                StandardDeleteOption.OVERRIDE_READ_ONLY);
 
+        final Counters.PathCounters deleteCounters;
+        try {
+            deleteCounters = PathUtils.delete(
+                    file.toPath(), PathUtils.EMPTY_LINK_OPTION_ARRAY,
+                    StandardDeleteOption.OVERRIDE_READ_ONLY);
+        } catch (final IOException ex) {
+            throw new IOException("Cannot delete file: " + file, ex);
+        }
         if (deleteCounters.getFileCounter().get() < 1 && deleteCounters.getDirectoryCounter().get() < 1) {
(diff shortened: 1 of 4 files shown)


garydgregory:
@elharo see build failures.

elharo:
Error:  Failures: 
Error:    FileUtilsDeleteDirectoryLinuxTest.testThrowsOnCannotDeleteFile:91 expected: <true> but was: <false>

elharo:
Failing test only runs on Linux. A little more info:

Failures: 
Error:    FileUtilsDeleteDirectoryLinuxTest.testThrowsOnCannotDeleteFile:92 AccessDeniedException #0: /tmp/junit148728722527736013/nested/restricted ==> expected: <true> but was: <false>

elharo:
strange that the failing method didn't actually change in this PR

src/main/java/org/apache/commons/io/FileUtils.java
```diff
@@ -377,8 +378,12 @@ public static boolean contentEquals(final File file1, final File file2) throws I
             return true;
         }
 
-        checkFileExists(file1, "file1");
-        checkFileExists(file2, "file2");
+        if (!file1.isFile()) {
+            throw new IllegalArgumentException("Parameter '" + "file1" + "' is not a file: " + file1);
```
garydgregory, line 382: Hello @elharo 
This must be a typo: `"Parameter '" + "file1" + "' is not a file: "`
There is now the same (weird) pattern x3
```
        if (!file2.isFile()) {
            throw new IllegalArgumentException("Parameter '" + "file2" + "' is not a file: " + file2);
        }
```
This was the point of the `requireX()` method, so this should be fixed and refactored. But then, this looks the same as the existing `requireFile()` method... what am I missing here?

elharo: checkFileExists also includes a check that the file exists and throws a FileNotFoundException if it does not. However, at this point in the method it is already known that the files exist, so we don't want to check that again.

garydgregory: Hello @elharo 
The weird string construction is still bogus IMO and the pattern in triplicate should be refactored into a new private method.

elharo: OK. I see what you're saying now. I'll combine the strings. I did this elsewhere. It's just an artifact of Eclipse inlining.

PR #538, opened 2023-12-26
limit public API to enhance evolvability in future

elharo:
and small Javadoc improvements

@garydgregory


garydgregory:
Note: This PR is probably OK since Apache HBase does not appear to make use of these methods and I intend to create a PR to reuse this class in HBase as HBase already uses Commons IO.

PR #541, opened 2023-12-28
[IO-829] don't decode and reencode characters in a potentially different charset

elharo:
This fixes getReader but I think getInputStream and getByteArray are still issues. @garydgregory


garydgregory:
Is it possible to create a unit test here to show the issue?

garydgregory:
https://issues.apache.org/jira/browse/IO-829

PR #543, opened 2023-12-29
Test that char sequences can be read in multiple encodings

elharo:
The bug occurs when the default character set is different than the specified character set. Since we don't know the default character set in which the test runs a priori, test two incompatible character sets against each other. They can't both be the default character set.

### DIFF
--- a/src/test/java/org/apache/commons/io/build/CharSequenceOriginTest.java
+++ b/src/test/java/org/apache/commons/io/build/CharSequenceOriginTest.java
@@ -16,10 +16,12 @@
  */
 package org.apache.commons.io.build;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/io/build/CharSequenceOriginTest.java
```diff
@@ -84,4 +84,13 @@ public void testGetWriter() {
         assertThrows(UnsupportedOperationException.class, super::testGetWriter);
     }
 
+    @Test
+    public void testGetReader_UTF16() throws IOException {
```
garydgregory, line 88: Doesn't this test the same behavior as 'testGetReaderIgnoreCharset()?

elharo: Yes. I saw your comment asking for a test before I saw you had added one.

garydgregory:
Hi @elharo 
OK to close then?

PR #544, opened 2023-12-29
show how http URI origins don't work

elharo:
java.nio.file.FileSystemNotFoundException: Provider "https" not installed
	at java.nio.file.Paths.get(Paths.java:147)
	at org.apache.commons.io.build.AbstractOrigin$URIOrigin.getPath(AbstractOrigin.java:402)
	at org.apache.commons.io.build.AbstractOrigin.getInputStream(AbstractOrigin.java:540)
	at org.apache.commons.io.build.URIOriginTest.testReadFromURL(URIOriginTest.java:47)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at java.util.ArrayList.forEach(ArrayList.java:1257)

### DIFF
--- a/src/test/java/org/apache/commons/io/build/URIOriginTest.java
+++ b/src/test/java/org/apache/commons/io/build/URIOriginTest.java
@@ -16,11 +16,17 @@
  */
 package org.apache.commons.io.build;
 
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+import java.io.IOException;
+import java.io.InputStream;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.nio.file.Paths;
 
 import org.apache.commons.io.build.AbstractOrigin.URIOrigin;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
 
 /**
  * Tests {@link URIOrigin}.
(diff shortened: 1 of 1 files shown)


elharo:
That is true, but that is an implementation detail. There's nothing in this class or API that indicates it should depend on which file system providers are installed. And this is about URIs, not file systems. I would be equally surprised if CharSequenceOrigin depended on file system providers. 

URIOrigin should be able to provide at l;east streams and readers for any URL the JDK can handle. 

Overall though as I explore this, I'm coming to the conclusion that Origin is attempting to provide an sbstraction  for something that doesn't exist. That is, there is no true common type for files, byte arrays, strings, URLs, etc. Hence all the UnsupportedOPerationExceptions and surprising failures like this one.

garydgregory:
I suppose it's semi-surprising that Java versions 11 and up don't support HTTP since they do provide an HTTP client. At least a read-only file system should be "simple" without getting into the whole WebDAV thing.

elharo:
That's not quite accurate. Java versions 1 and up do support HTTP. They just don't support it as a file system since it really isn't a file system. Neither are byte arrays or char sequences, of course.

garydgregory:
@elharo 
If you rebase on git master, this should now work due to https://github.com/apache/commons-io/pull/630

garydgregory:
Closing, this works now.

See `org.apache.commons.io.build.URIOriginTest.testGetInputStream(String)`:
```java
    @ParameterizedTest
    @ValueSource(strings = {
            "http://example.com",
            "https://example.com"
    })
    void testGetInputStream(final String uri) throws Exception {
        final AbstractOrigin.URIOrigin origin = new AbstractOrigin.URIOrigin(new URI(uri));
        try (final InputStream in = origin.getInputStream()) {
            assertNotEquals(-1, in.read());
        }
    }
```

PR #547, opened 2024-01-01
[IO-807] Characterization test for broken symlinks when copying directories

elharo:
This establishes the current behavior and makes sure we don't accidentally change it.

@garydgregory

### DIFF
--- a/src/test/java/org/apache/commons/io/FileUtilsTest.java
+++ b/src/test/java/org/apache/commons/io/FileUtilsTest.java
@@ -835,13 +835,8 @@ public class FileUtilsTest extends AbstractTempDirTest {
 
         // Now copy sourceDirectory, including the broken link, to another directory
         final File destination = new File(tempDirFile, "destination");
-        try {
-            FileUtils.copyDirectory(sourceDirectory, destination);
-            fail("ignored broken link");
-        } catch (FileNotFoundException ex) {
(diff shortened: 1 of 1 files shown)


src/test/java/org/apache/commons/io/FileUtilsTest.java
```diff
@@ -808,6 +811,39 @@ public void testToURLs3() {
...
+        // Delete the file file to break the symlink
+        assumeTrue(targetFile.delete());
+        assumeFalse(Files.exists(linkPath));
+        assumeTrue(Files.exists(linkPath, LinkOption.NOFOLLOW_LINKS));
+
+        // Now copy sourceDirectory, including the broken link, to another directory
+        final File destination = new File(tempDirFile, "destination");
+        try {
+            FileUtils.copyDirectory(sourceDirectory, destination);
+            fail("ignored broken link");
```
garydgregory, line 840: Please use `assertThrows()`.

garydgregory (changes requested):
@elharo 
Thank you for your PR. Please see my comment in the test.

src/test/java/org/apache/commons/io/FileUtilsTest.java
elharo: done

PR #550, opened 2024-01-01
XmlStreamReader can't parse XML document with multi-line prolog

slachiewicz:
I have a test case in maven-shared-utils that started to fail after the upgrade from 2.13 to 2.14 (also 2.15.1)

```
<?xml\nversion='1.0'\nencoding\n=\n'US-ASCII'\n?>\n<element encoding='attribute value'/>"
```
returned encoding is UTF-8 where I expect to get US-ASCII

 I don't know if it is an allowed format.

https://github.com/apache/maven-shared-utils/pull/170

### DIFF
--- a/src/test/java/org/apache/commons/io/input/XmlStreamReaderTest.java
+++ b/src/test/java/org/apache/commons/io/input/XmlStreamReaderTest.java
@@ -60,6 +60,8 @@ public class XmlStreamReaderTest {
     private static final String UTF_32LE = "UTF-32LE";
     private static final String UTF_32BE = "UTF-32BE";
     private static final String UTF_8 = StandardCharsets.UTF_8.name();
+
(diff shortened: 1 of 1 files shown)


garydgregory:
TY @slachiewicz !
Fixed in the git `master` branch and `2.16.0-SNAPSHOT` [builds](https://repository.apache.org/content/repositories/snapshots/commons-io/commons-io/2.16.0-SNAPSHOT). You are credited in `changes.xml`. I did not apply this patch since it would have failed the build. The test changes are in git master though. Please test your use case and report back 😄

garydgregory:
Closing, see https://github.com/apache/commons-io/pull/550#issuecomment-1874076416

slachiewicz:
Thx, I can confirm that my issue has been fixed.

PR #551, opened 2024-01-04
Amend getFullPathNoEndSeparator description for trailing slash home path behaviour

Philzen:
The description says it will remove the trailing slash from a home path (relative path starting with tilde).

A simple

```java
        @Test void returnsHomeWithTrailingSlash_forHomeWithTrailingSlash() {
            assertThat(FileHelper.getParent("~/")).isEqualTo("~/");
        }
```

proves otherwise.

This fixes the description accordingly.


elharo (approved):
In this case I think the existing behavior and the changed docs are correct.

Philzen:
Any objections merging this?

elharo (approved):

PR #552, opened 2024-01-04
Amend description to reflect empty string result for windows drive

Philzen:
Not sure if this is is a bug or a feature – what i can confirm is that in v2.4 it worked as it said in the description:

```java
@Test void returnsHomeWithTrailingSlash_forHomeWithTrailingSlash() {
    assertThat(FilenameUtils.getFullPathNoEndSeparator("C:")).isEqualTo("C:");
}
```

After upgrading to 2.15.1, i realized my test suite failed, the behaviour is now:
```java
@Test void returnsHomeWithTrailingSlash_forHomeWithTrailingSlash() {
    assertThat(FilenameUtils.getFullPathNoEndSeparator("C:")).isEqualTo("");
}
```

This PR proposes to reflect this change in the javadoc.

### DIFF
--- a/src/main/java/org/apache/commons/io/FilenameUtils.java
+++ b/src/main/java/org/apache/commons/io/FilenameUtils.java
@@ -664,7 +664,7 @@ public class FilenameUtils {
      * a.txt        --&gt; ""
      * a/b/c        --&gt; a/b
      * a/b/c/       --&gt; a/b/c
-     * C:           --&gt; C:
+     * C:           --&gt; ""
      * C:\          --&gt; C:\
      * ~            --&gt; ~
      * ~/           --&gt; ~


elharo:
You've pointed to existing bad API design. getFullPathNoEndSeparator("C:") should return "" if the argument is a Unix path and "C:" if that path is a Windows path. Of course, it's just a string, and we have no way of knowing whether that's a Windows path or a Posix path. In this case, best guess is Windows since "C:" is very common in Windows and very uncommon on Unix, but it's still a guess. Possibly we should deprecate this method and rethink the API here.

Philzen:
> Possibly we should deprecate this method and rethink the API here.

@elharo Agreed. For the time being, any objection merging this PR to bring the JavaDoc inline with the method's actual behavior?

govind-gupta-tfs:
@elharo  pls see the above message

PR #553, opened 2024-01-04
Amend description to reflect correct named user result with trailing slash

Philzen:
Similar to #551, the trailing slash is actually preserved in the result here.

Confirmed to have this behaviour in 2.4 and latest 2.15.1

### DIFF
--- a/src/main/java/org/apache/commons/io/FilenameUtils.java
+++ b/src/main/java/org/apache/commons/io/FilenameUtils.java
@@ -669,7 +669,7 @@ public class FilenameUtils {
      * ~            --&gt; ~
      * ~/           --&gt; ~
      * ~user        --&gt; ~user
-     * ~user/       --&gt; ~user
+     * ~user/       --&gt; ~user/
      * </pre>
      * <p>
      * The output will be the same irrespective of the machine that the code is running on.


src/main/java/org/apache/commons/io/FilenameUtils.java
```diff
@@ -669,7 +669,7 @@ public static String getFullPath(final String fileName) {
      * ~            --&gt; ~
      * ~/           --&gt; ~
      * ~user        --&gt; ~user
-     * ~user/       --&gt; ~user
+     * ~user/       --&gt; ~user/
```
garydgregory, line 672: This causes the Javadoc to contradict itself with the above's `excluding the final directory separator`.
The question is whether Windows's `"C:\"` unchanged output is a special case or a bug.

elharo: The existing javadoc and API is very unclear. I do not understand "Gets the full path from a full fileName, which is the prefix + path, and also excluding the final directory separator."

The argument is not a filename. There is no such thing as a "full path" in normal parlance. Bug filed. https://issues.apache.org/jira/browse/IO-836

I also agree that if this method turns ~user/ into ~user/ that's a bug. It should produce ~user.

C:\ should also (80% certainty) turn into C:, not C:\

I further expect there are additional bugs in this method involving Unix shell conventions and platform dependencies, but I'd have to dig into the rest of the class to understand exactly what those are.

What a tasy can of worms you've opened. :-)

elharo (changes requested):

garydgregory:
Note that we should not treat ~ in any special way since it is a Unix-like shell utility and not a file system functionality.

Philzen:
@elharo @garydgregory i totally see the points you've raised and agree it may be worth tackling the current behaviour in a bugfix.

However, until that happens, i believe it will be worthwhile to sync the JavaDoc with the current behavior to minimize bad surprises when consuming the library.

garydgregory:
Hello @Philzen 
It looks to me like git master already contains the Javadoc change. What about all the callers of the API?

PR #562, opened 2024-01-23
avoid deprecated methods and other minor cleanups in FileUtilsTest

elharo:
@garydgregory

### DIFF
--- a/src/test/java/org/apache/commons/io/FileUtilsTest.java
+++ b/src/test/java/org/apache/commons/io/FileUtilsTest.java
@@ -269,9 +269,8 @@ public class FileUtilsTest extends AbstractTempDirTest {
 
     private void iterateFilesAndDirs(final File dir, final IOFileFilter fileFilter,
         final IOFileFilter dirFilter, final Collection<File> expectedFilesAndDirs) {
-        final Iterator<File> iterator;
+        final Iterator<File> iterator = FileUtils.iterateFilesAndDirs(dir, fileFilter, dirFilter);
         int filesCount = 0;
-        iterator = FileUtils.iterateFilesAndDirs(dir, fileFilter, dirFilter);
         try {
             final List<File> actualFiles = new ArrayList<>();
             while (iterator.hasNext()) {
@@ -288,7 +287,7 @@ public class FileUtilsTest extends AbstractTempDirTest {
         }
     }
 
-    void openOutputStream_noParent(final boolean createFile) throws Exception {
+    private void openOutputStream_noParent(final boolean createFile) throws Exception {
         final File file = new File("test.txt");
         assertNull(file.getParentFile());
         try {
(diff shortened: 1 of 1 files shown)


garydgregory:
-1 as is, deprecated code must still be tested.

I guess we could have separate methods or classes that only test deprecated code but that seems like busy work.

It is OK to change 'main' code to not use deprecated elements, as long as test coverage does not go down, in which case a new test should be written to cover that change, for example if a deprecated method ends up not called at all from anywhere.

elharo:
None of the deprecation changes affect the tests. They are changes to the setup code for the tests.

elharo:
There's one failure on Java 21 only. This test wasn't touched in this PR. It's probably just flaky, but it's also possible it's broekn at HEAD:

 Error:  Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.178 s <<< FAILURE! -- in org.apache.commons.io.FileUtilsWaitForTest
Error:  org.apache.commons.io.FileUtilsWaitForTest.testWaitForInterrupted -- Time elapsed: 0.022 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:63)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:36)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:31)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:183)
	at org.apache.commons.io.FileUtilsWaitForTest.testWaitForInterrupted(FileUtilsWaitForTest.java:60)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)

elharo:
As far as code coverage I see

[2024-01-23T00:44:13.800Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}

garydgregory:
Hi @elharo 
>None of the deprecation changes affect the tests. 

But they do since some of the changes are to `@Test` methods. 

I really don't want to fiddle with this type of change. If we want to create special tests that only test deprecated code (maybe using the `testDeprecated` method prefix), that's fine, then we can only test non-deprecated code in "normal" tests. Otherwise, I don't think we need to change `@Test` methods.

elharo:
Yes, they are in @Test methods but no, they are not tested by these methods. All the ones I have changed here do is write some data into a file that is then overwritten before anything is asserted. The purpose is to make sure that data is overwritten by the subsequent calls when append=False, not to test that the deprecated method is behaving according to spec.

garydgregory:
@elharo I do not see any negative impact on the JaCoCo code coverage report for the `FileUtils` class or the whole code base. Merging.

PR #563, opened 2024-01-23
[IO-842] deprecate writeLines methods without a charset

elharo:
Also corrects Javadoc about which exception is thrown when the charset is not recognized and adds test for same. Note that the exception thrown is not changed here. I brought the documentation into sync with the actual pre-existing behavior. @garydgregory


garydgregory:
https://issues.apache.org/jira/browse/IO-842

garydgregory:
Let's wait to see what happens with https://issues.apache.org/jira/browse/IO-842 
I'll post to the dev mailing list to get a broader audience.

PR #566, opened 2024-01-25
Clarify and correct EndianUtils and SwappedDataInputStream API doc

elharo:
Most of the methods in this class are badly named. Except for the swap methods, they all just read and write in little endian order. E.g. consider readSwappedInteger

```
    public static int readSwappedInteger(final InputStream input) throws IOException {
        final int value1 = read(input);
        final int value2 = read(input);
        final int value3 = read(input);
        final int value4 = read(input);
        return ((value1 & 0xff) << 0) + ((value2 & 0xff) << 8) + ((value3 & 0xff) << 16) + ((value4 & 0xff) << 24);
    }
```

This always reads four bytes from an input stream and always considers the first byte read to be the least significant. It would have been better named readLittleEndianInteger.

I'm tempted to deprecate the whole thing and replace it with something better named, but for now at least let's correct the Java doc. 

@garydgregory

### DIFF
--- a/src/main/java/org/apache/commons/io/EndianUtils.java
+++ b/src/main/java/org/apache/commons/io/EndianUtils.java
@@ -27,11 +27,16 @@ import java.io.OutputStream;
  * Helps with different endian systems.
  * <p>
  * Different computer architectures adopt different conventions for
- * byte ordering. In so-called "Little Endian" architectures (eg Intel),
+ * byte ordering. In so-called "Little Endian" architectures (e.g. X86),
  * the low-order byte is stored in memory at the lowest address, and
  * subsequent bytes at higher addresses. For "Big Endian" architectures
- * (eg Motorola), the situation is reversed.
- * This class helps you solve this incompatibility.
+ * (e.g. Motorola 680X0), the situation is reversed.
+ * Most methods and classes throughout Java &mdash; e.g. DataInputStream and
+ * Double.longBitsToDouble() &mdash; assume data is laid out
+ * in big endian order with the most significant byte first.
+ * The methods in this class all read data in little endian order,
+ * generally by reversing the bytes in the input and then using the
+ * regular Java methods to convert the swapped bytes to a primitive type.
  * </p>
  * <p>
  * Provenance: Excalibur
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/io/EndianUtils.java
```diff
@@ -78,19 +84,20 @@ public static double readSwappedDouble(final InputStream input) throws IOExcepti
     }
 
     /**
-     * Reads a "float" value from a byte array at a given offset. The value is
-     * converted to the opposed endian system while reading.
+     * Reads a little endian "float" value from a byte array at a given offset.
```
garydgregory, line 87: I think that using `{@code _type_}` would be better here.

elharo: Maybe. What's really confusing here is we need to distinguish two different "floats":

1. The Java float primitive in memory in the VM.
2. A 4 byte chunk read from an input stream or an array. 

I'm not sure I've got this distinction clear enough yet. Let me play with it a bit more.

elharo:
PTAL

src/main/java/org/apache/commons/io/EndianUtils.java
```diff
@@ -261,7 +275,8 @@ public static float swapFloat(final float value) {
     }
 
     /**
-     * Converts an "int" value between endian systems.
+     * Converts an {@code int} value between endian systems.
```
garydgregory, line 278: We should be clear that the toggle is between big and little-endian. We'll then avoid the shadow of even thinking about middle-endian madness.

elharo: Good point. Curious, has any system ever actually used middle-endianess or is it just a theoretical possibility? Anyway, let me update this.

garydgregory:
According to Wikipedia, PDP has something close also referred to PDP-endian!

elharo (commented):
Also corrected Javadoc for SwappedDataInputStream, which had all these problems and some more copy paste errors besides.

src/main/java/org/apache/commons/io/EndianUtils.java
```diff
@@ -182,13 +194,14 @@ public static short readSwappedShort(final InputStream input) throws IOException
     }
 
     /**
-     * Reads an unsigned integer (32-bit) value from a byte array at a given
-     * offset. The value is converted to the opposed endian system while
-     * reading.
+     * Reads a little endian unsigned integer (32-bit) value from a byte array at a given
```
garydgregory, line 197: Hi @elharo 

This is slightly confusing to me: The method returns a `long` (64-bit) but we are reading an 32-bit integer? Should we clarify this point? Something like "Read a foo as a bar".

elharo: Since Java's ints are signed you can't fit the unsigned int values from ~2.2 billion  to ~4 billion into an int, so you have to return a long. The long will always be <= 2^32, but might be greater than Integer.MAX_INT. Not sure how much of that to say here.

garydgregory: Makes sense, the comment already mentions "unsigned". TY for the clarification.

PR #569, opened 2024-01-27
Modify Builder.get error condition

pingpingy1:
The `Builder.get` method claims to throw an `UnsupportedOperationException` if its origin is invalid. However, it actually throws a `NullPointerException` if the origin is `null`.
This commit changes the workflow to take care of cases where neither `RandomAccessFile` nor origin is set.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/RandomAccessFileInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/RandomAccessFileInputStream.java
@@ -73,12 +73,17 @@ public class RandomAccessFileInputStream extends InputStream {
         @SuppressWarnings("resource") // Caller closes depending on settings
         @Override
         public RandomAccessFileInputStream get() throws IOException {
+            if (randomAccessFile == null && getOrigin() == null) {
+                throw new UnsupportedOperationException("Neither RandomAccessFile nor origin is set.");
+            }
+            if (randomAccessFile != null && getOrigin() != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/commons/io/input/RandomAccessFileInputStream.java
```diff
@@ -73,12 +73,17 @@ public static class Builder extends AbstractStreamBuilder<RandomAccessFileInputS
         @SuppressWarnings("resource") // Caller closes depending on settings
         @Override
         public RandomAccessFileInputStream get() throws IOException {
+            if (randomAccessFile == null && getOrigin() == null) {
```
elharo, line 76: I wonder if in this case NullPointerException is correct and we should change the doc to match the current behavior instead?

elharo (commented):
This is a really good example of where we shouldn't be using builders. Builders introduce additional error conditions that wouldn't happen with constructors. I'd like to deprecate/remove all of this and use vanilla constructors.

garydgregory:
In this case, I agree that updating the Java document is better because this hierarchy uses UnsupportedOperationException to say "this method is not implemented and can't support the method call. It is not meant for argument or state validation failures.

pingpingy1:
> @pingpingy1 I don't think this PR is needed anymore, please see git master or snapshot builds from https://repository.apache.org/content/repositories/snapshots/

OK, I'll close this PR.

PR #571, opened 2024-01-29
[IO-845] test links to targets outside the source directory

elharo:
Another characterization test, but this one is for behavior I expect we'll want to keep. Specifically it tests that when a link inside the source directory points to a file outside the source directory, the newly copied link still points to that same file. 

I do think we should separate this case from the case where the target of the link is inside the source directory. In that case, I think we'll want to repoint the link to aim at the new copy of the target rather than the original target. That change is not made in this PR. 

Note that this might lock in the breakage that Google reported with their test resource files. However, the more I think about that the more I'm convinced that resolving all aliases to actual files is surprising behavior. Consider a case where a single file is linked to from three different places. Copying the directory would change the graph and replace one file with four independent copies. 

@garydgregory @eamonnmcmanus

### DIFF
--- a/src/test/java/org/apache/commons/io/FileUtilsTest.java
+++ b/src/test/java/org/apache/commons/io/FileUtilsTest.java
@@ -778,7 +778,7 @@ public class FileUtilsTest extends AbstractTempDirTest {
         final String actual = FileUtils.readFileToString(copiedLink, "UTF8");
         assertEquals("HELLO WORLD", actual);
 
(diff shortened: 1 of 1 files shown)


elharo:
Failure looks flaky and unrelated:

Error: Action 'https://api.github.com/repos/actions/cache/tarball/13aacd865c20de90d75de3b17ebe84f7a17d57d2' download has timed out. Error: The request was canceled due to the configured HttpClient.Timeout of 100 seconds elapsing.

PR #573, opened 2024-01-31
Focus Javadoc on current version rather than past versions

elharo:
The versions removed are all over 10 years old. In multiple cases the existing wording was not just out of date but incorrect. In multiple cases I found copy pasta in methods that didn't exist until after version 2.2. @garydgregory


garydgregory:
@elharo 
Please run `mvn` (by itself) before pushing changes to catch these types of build errors.

elharo:
Infra failure on Java 21

Warning: Failed to download action 'https://api.github.com/repos/actions/cache/tarball/13aacd865c20de90d75de3b17ebe84f7a17d57d2'. Error: nodename nor servname provided, or not known (api.github.com:443)
Warning: Back off 11.456 seconds before retry.

PR #575, opened 2024-01-31
More specific catch clause in FileUtils.decodeUrl()

elharo:
These are the only two exception that signal a badly encoded percent octet. A runtime exception from the byte buffer would indicate a bug in this code. @garydgregory

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -1184,7 +1184,7 @@ public class FileUtils {
                             i += 3;
                         } while (i < n && url.charAt(i) == '%');
                         continue;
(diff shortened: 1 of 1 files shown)


garydgregory:
I am slightly worried this will break existing applications for conditions that we do not cover in our unit tests.

elharo:
I think anything that breaks would reveal a bug in our code or, less likely, the JDK. There are no null pointer exceptions here because all the variables are primitives or previously dereferenced before the try block. The only methods that throw as a result of bad encoding are  parseInt, substring, and charAt. Those are well documented and understood. If ByteBuffer throws, that's definitely a bug in our code we shouldn't ignore.

garydgregory:
K, sounds good.

PR #581, opened 2024-02-07
[IO-808] more precise Javadoc about when an IllegalArgumentException is thrown

elharo:
Clarify the circumstances under which an IllegalArgumentException is currently thrown. I think these are all the cases where [IO-808](https://issues.apache.org/jira/projects/IO/issues/IO-808?filter=allopenissues) proposes using IOExceptions instead. This is an intermediate step that merely documents what the code currently does. Some of these should be changed. @garydgregory @sebbASF

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -2245,7 +2245,7 @@ public class FileUtils {
      * @param fileFilter Optional file filter, may be null.
      * @return The files in the directory, never {@code null}.
      * @throws NullPointerException if directory is {@code null}.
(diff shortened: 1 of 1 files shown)


garydgregory:
Hello @elharo 
Please link to Jira URLs for convenience.

garydgregory:
@elharo 
There are lots of build errors where it looks like you mixed up `@link` for `@code`:
```
...
Error:  /home/runner/work/commons-io/commons-io/src/main/java/org/apache/commons/io/FileUtils.java:2442: error: reference not found
Error:       * @throws IllegalArgumentException if {@link srcFile} is a directory
Error:                                                    ^
...
```
Run `mvn` by itself before you push to avoid build errors.

PR #582, opened 2024-02-08
[IO-808] non writeable destinations are I/O errors

elharo:
No API changes. No existing tests failed. I added a new test for this. @garydgregory

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -852,7 +852,6 @@ public class FileUtils {
         createParentDirectories(destFile);
         if (destFile.exists()) {
             checkFileExists(destFile, "destFile");
-            requireCanWrite(destFile, "destFile");
         }
 
         final Path srcPath = srcFile.toPath();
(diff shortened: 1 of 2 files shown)


garydgregory:
A question is whether we should even check for writable attributes before we write. There comments from time to time that our APIs are (obviously IMO) non-atomic, and these checks make the code fail faster but also risk of being out of sync with the underlying file system.

elharo:
This PR seems unlikely to require source changes by users. These methods can already throw IOExceptions and those need to be handled. The only potential problem, and it's not a likely one, is if someone has a separate IllegalArgumentException catch block with special logic to handle read only files. In that case they'd need to move that logic into a block that catches IOException. However, I think more users are likely to be helped by combining IOException into a single exception type.

As to simply removing the writeable check completely, that's a definite maybe. It would simplify the code, and simplify the happy path where there's no problem with read-only files. Public API remains the same. It's a very good point that this risks being out of sync with the fiel system by the time we write. Should I go ahead and remove the requireCanWrite method?

garydgregory:
> Should I go ahead and remove the requireCanWrite method?

I think so.

elharo:
> > Should I go ahead and remove the requireCanWrite method?
> 
> I think so, yes.

done

PR #586, opened 2024-02-09
[IO-808] Report problems in the file system as IOExceptions

elharo:
Problems that can't be determined from the argument alone, that is problems that aren't detected until the file system is read, are IOExceptions, not program bugs represented by IllegalArgumentException. No API changes. @garydgregory

### DIFF
--- a/src/main/java/org/apache/commons/io/FileUtils.java
+++ b/src/main/java/org/apache/commons/io/FileUtils.java
@@ -289,13 +289,13 @@ public class FileUtils {
      * @param name The parameter name to use in the exception message.
      * @throws FileNotFoundException if the file does not exist
      * @throws NullPointerException if the given {@link File} is {@code null}.
-     * @throws IllegalArgumentException if the given {@link File} is not a file.
+     * @throws IOException if the given {@link File} is not a file.
      */
-    private static void checkFileExists(final File file, final String name) throws FileNotFoundException {
+    private static void checkFileExists(final File file, final String name) throws IOException {
         Objects.requireNonNull(file, name);
         if (!file.isFile()) {
             if (file.exists()) {
-                throw new IllegalArgumentException("Parameter '" + name + "' is not a file: " + file);
+                throw new IOException("Parameter '" + name + "' is not a file: " + file);
             }
             if (!Files.isSymbolicLink(file.toPath())) {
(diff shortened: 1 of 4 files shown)


jochenw:
Quoting from the Javadoc of IllegalArgumentException: "Thrown to indicate that a method has been passed an illegal or inappropriate argument."

In other words, throwing an IAE is perfectly alright, there is no condition on how to detect the fact, that is the argument is illegal, or inappropriate.

I conclude, that this patch adds nothing to the existing code, at the cost of changing method signatures. I will not merge it.

garydgregory:
Closing, I agree with @jochenw .

elharo:
When the problem is detected a core difference between runtime and checked exceptions. Runtime exceptions indicate program bugs, like passing the wrong argument. Checked exceptions indicate environmental problems that are not known at compile time.

A file that does not exist is not an illegal or inappropriate argument. It's not the argument that's wrong. It's the file system. Consider this.

The file could exist when the argument is passed and then be deleted by a different process before the method tries to write or read it. The argument was 100% correct but there's still an exception. That shouldn't be an IllegalArgumentException. The argument wasn't the problem he I/O was.

jochenw:
Quoting @elharo: "A file that does not exist is not an illegal or inappropriate argument. "

That is, in my opinion, a matter of policy, hence questionable.

My personal opinion is this: It is the callers responsibility to ensure, that a file exists, before invoking an operation like copyFile, or moveFile. How that is ensured, is up to the caller. However, if you adopt that opinion, then it is clearly an illegal argument.

Whatever, I do not wish to continue this discussion. In the end, it is pointless. The current code works,  as documented. (@throws IAE.) Even if we agreed on your policies, I'd still hold to the statement "Works as documented (perhaps not exactly as expected by you), so no need to fix anything."

elharo:
The "policy" is set by the design of Java since 1.0. The current approach fights against how exceptions were designed to work in Java. Sure, you can do it that way and it compiles and perhaps is documented, but it's more than little like building a car where first gear is full highway cruising speed and fourth gear is for pulling a horse trailer up a mountain. It's going to confuse everyone who tries to drive it and probably cause a few accidents, and documenting this in the owners' manual really wouldn't help.

PR #609, opened 2024-04-04
ThresholdingOutputStream: a negative threshold should behave like a zero threshold and trigger the event on the first write

rproserpio:
#587 introduces a change in behavior regarding cases where a `ThresholdingOutputStream` is initialized with a negative threshold.

Previously, the `thresholdConsumer` would be notified upon the first write, now it is not.

The previous behavior was relied upon at least by the `DeferredFileOutputStream` implementation.
In this implementation the underlying `currentBuffer` isn't switched from the memory one until the threshold exceeded callback is triggered.
If the thresholding output stream has already exceeded the threshold on initialization, the switch never happens. This causes the writes to go to the wrong buffer.

I'm not sure what the right approach for a fix would be. On one hand the `isThresholdExceeded` method was incorrect. On the other hand, even if we fix `DeferredFileOutputStream`, other implementations could still rely on the previous behavior.

DeferredFileOutputStream is used by commons-fileupload, that is broken when used with commons-io 2.16.0.

@elharo @garydgregory, can I ask your opinion on this?

### DIFF
--- a/src/main/java/org/apache/commons/io/output/ThresholdingOutputStream.java
+++ b/src/main/java/org/apache/commons/io/output/ThresholdingOutputStream.java
@@ -98,7 +98,6 @@ public class ThresholdingOutputStream extends OutputStream {
     /**
      * Checks to see if writing the specified number of bytes would cause the configured threshold to be exceeded. If
      * so, triggers an event to allow a concrete implementation to take action on this.
-     * If the count is not positive, the check is skipped.
      *
      * @param count The number of bytes about to be written to the underlying output stream.
      *
--- a/src/main/java/org/apache/commons/io/output/ThresholdingOutputStream.java
+++ b/src/main/java/org/apache/commons/io/output/ThresholdingOutputStream.java
@@ -104,7 +104,7 @@ public class ThresholdingOutputStream extends OutputStream {
      * @throws IOException if an error occurs.
      */
     protected void checkThreshold(final int count) throws IOException {
-        // thresholdReached is no bytes are to be actually written
+        // thresholdReached is not called if no bytes are to be actually written
         // this check handles the case of a negative threshold
         if (!thresholdExceeded && count > 0 && written + count > threshold) {
             thresholdExceeded = true;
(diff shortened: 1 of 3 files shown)


garydgregory:
Yeah, I'm not sure how I missed that. I don't understand why we don't just map a negative value to just 0, mapping it to "threshold reached" is suspicious. This needs a closer look...

rproserpio:
Thank you. Let me know if I could be of any help.

garydgregory:
Hi @rproserpio
Feel free to fix the code in this PR.
The behavior for a negative threshold should be the same as 0 IMO. WDYT? What does it even mean that a threshold is negative?
Writing zero bytes writes nothing, so there is nothing to reach until you at least write one byte.
Am I missing something?

rproserpio:
I agree. However, this would amount to essentially revert #507 (minus the the tests) and I don't know if that PR was done for some other reasons.

garydgregory:
Hello @rproserpio 

> I agree. However, this would amount to essentially revert #587 (minus the the tests) and I don't know if that PR was done for some other reasons.

Yeah, well, don't get lost in the PR description ;-) 

I invite you to study the code a bit more and update this PR. You could add comments to the code and test that describe what a negative and zero threshold means or doesn't. It seems you have an app or example in mind and it's always good to have an IRL use case to describe in PR/comments.

rproserpio:
I still not sure about this method:
```
    public boolean isThresholdExceeded() {
        return written > threshold;
    }
```

rproserpio:
@garydgregory I'm confident on the part of keeping the same behavior for zero and negatives thresholds. However:

> Writing zero bytes writes nothing, so there is nothing to reach until you at least write one byte.

This  assertion leaves me a bit uneasy, it's a possible behavior change. Not that I think writing zero bytes is really something that is often done, but you never know what inheritors do with classes designed to be extended.
Before 2.16, writing a zero byte array to a stream with negative threshold would have triggered the event. However I can't think of a reason we would want this.

garydgregory:
Hi @rproserpio 
Thank you for your update. Unless there is actually meaning to having a negative threshold, I think it should be normalized to 0 in the constructor. 0 has meaning. WDYT?

rproserpio:
Yes, that sounds like the better solution. The only observable difference from before 2.16.0 should be in `isThresholdExceeded`, where a count of zero bytes and what was a negative threshold now returns false. But that makes sense. For the use case I was looking at (commons-fileupload) this solution is even better.

I've reverted the other changes in the class, cleaned up a test and added a phrase in the constructor doc.
I've used a ternary operator instead of a Math.max for the check, I hope it fits the conventions.
Thank you for your help

garydgregory:
Merged. I canceled 2.16.1 RC1; RC2 is on the way.
TY @rproserpio !

PR #629, opened 2024-05-19
[IO-783] Remove NTFS ADS separator check in FilenameUtils.getExtension

baezzys:
### Summary
This PR addresses issue [IO-783](https://issues.apache.org/jira/browse/IO-783) by removing the NTFS ADS separator (colon) check from the `FilenameUtils.getExtension` method to ensure OS independence.

### Changes
- Removed the NTFS ADS separator (colon) validation from the `indexOfExtension` method in `FilenameUtils`.
- Updated the `FilenameUtils.getExtension` method to process filenames regardless of the operating system.

### DIFF
--- a/src/main/java/org/apache/commons/io/FilenameUtils.java
+++ b/src/main/java/org/apache/commons/io/FilenameUtils.java
@@ -988,13 +988,6 @@ public class FilenameUtils {
         if (fileName == null) {
             return NOT_FOUND;
         }
-        if (isSystemWindows()) {
-            // Special handling for NTFS ADS: Don't accept colon in the file name.
-            final int offset = fileName.indexOf(':', getAdsCriticalOffset(fileName));
-            if (offset != -1) {
-                throw new IllegalArgumentException("NTFS ADS separator (':') in file name is forbidden.");
-            }
-        }
         final int extensionPos = fileName.lastIndexOf(EXTENSION_SEPARATOR);
         final int lastSeparator = indexOfLastSeparator(fileName);
         return lastSeparator > extensionPos ? NOT_FOUND : extensionPos;
--- a/src/test/java/org/apache/commons/io/FilenameUtilsTest.java
+++ b/src/test/java/org/apache/commons/io/FilenameUtilsTest.java
@@ -247,15 +247,7 @@ public class FilenameUtilsTest {
         assertEquals("", FilenameUtils.getExtension("a\\b\\c"));
         assertEquals("", FilenameUtils.getExtension("C:\\temp\\foo.bar\\README"));
         assertEquals("ext", FilenameUtils.getExtension("../filename.ext"));
-
-        if (FilenameUtils.isSystemWindows()) {
-            // Special case handling for NTFS ADS names
-            final IllegalArgumentException e = assertThrows(IllegalArgumentException.class, () -> FilenameUtils.getExtension("foo.exe:bar.txt"));
-            assertEquals("NTFS ADS separator (':') in file name is forbidden.", e.getMessage());
-        } else {
-            // Upwards compatibility:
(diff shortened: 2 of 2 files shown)


garydgregory:
I don't think we should ping back and forth between releases in behavior... While this class suffers from some fundamental issues, this change will cause more problems since a feature now disappeared...

baulea:
@garydgregory 
In https://issues.apache.org/jira/projects/IO/issues/IO-783 Jochen Wiedmann pointed out, that filenames like "C:file.txt" without backslash are valid under Windows (relative path from current directory). 

But currently `FilenameUtils.getExtension("C:file.txt")` and `FilenameUtils.indexOfExtension("C:file.txt")` throw an IllegalArgumentException because of the NTFS ADS separator check.

Would you please rethink, if that shouldn't be fixed?

If you agree, I could  create a revised pull request, which adds unit tests for this case and also removes private methods now being superfluous. You might check the diff here: https://github.com/baulea/commons-io/commit/31b73aee1075f75fbdcc0a9876562158f1edab6d

jkbkupczyk:
@baulea @baezzys 
What about marking this method as deprecated and making a new one, that is OS independent? Currently you are breaking code - some programs may rely on that feature.

baulea:
@jkbkupczyk You are right, deprecating the current method and adding a new one is the smarter solution.
@garydgregory Do you agree?
@baezzys Would you adjust your pull request or should I provide a new one?

garydgregory:
-1:
- The current PR is completely wrong on Windows:
  - The PR returns part of the stream name or type, NOT the file extension! 
  - The format is `<filename>:<stream name>:<stream type>`
  - The file name extension for `"foo.exe:bar.txt"` is `"exe"`, not `"txt"`.
- The test needs an explicit check for `"C:file.txt"` since this is the common example given in the Jira and here.
- The needs tests for each combination of `<filename>:<stream name>:<stream type>`
- Also, with and without driver letters.

Tangent: Note that we also have the class FileSystems for OS-specific implementations but it does not have this kind of API.

See [NTFS Streams ](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/c54dec26-1551-4d3a-a0ea-4fa40f848eb3)

baulea:
@garydgregory If I understand [NTFS Streams](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/c54dec26-1551-4d3a-a0ea-4fa40f848eb3) correctly, then NTFS streams always contain two colons inside the  filename and may not contain only a single colon.

Currently the IllegalArgumentException is thrown, if a single colon is found after the last separator (/ or \\). Couldn't we just throw the IllegalArgumentException in org.apache.commons.io.FilenameUtils#indexOfExtension only on two (or more) colons after the last separator?

For the same reason "foo.exe:bar.txt" in the unit test is no valid NTFS stream name. But it is a valid filename under Linux and therefore it should not cause an exception and org.apache.commons.io.FilenameUtils#getExtension("foo.exe:bar.txt") should always return "txt".

What do you think?

garydgregory:
Hello @baezzys 

This all feels confusing because I don't think we follow strict behavior to match the names we use in methods and parameter names. 

For example, strictly speaking, on [Windows](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file), "foo.txt" is a file _name_ but "c:dir\foo.txt" is not a file name, is a _path_ pointing to a _file_. Then you have _components_, including _drive volumes_ also know as _drive designator_, directory names, and so on.

It's going to be delicate to change this code without breaking existing apps to provide bug fixes. So let's be careful.

IMO, moving this along is going to need a pragmatic approach with unit tests checking all these variations. Just talking about tweaking some method in PR comments isn't going to work for me, you have to run the whole test suite to check for unintended effects.

It's a valuable line of experimentation and I'm sure there's room for improvement.

HTH

PR #630, opened 2024-05-23
[IO-831] Support getInputStream() for https & http in URIOrigin

thachlp:
#### Summary
This PR addresses issue IO-831, support `http` and `https`

#### Changes
Handle in case uri are `http` or `https`

### DIFF
--- a/src/main/java/org/apache/commons/io/build/AbstractOrigin.java
+++ b/src/main/java/org/apache/commons/io/build/AbstractOrigin.java
@@ -409,15 +409,12 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
             final URI uri = get();
             final String scheme = uri.getScheme();
             final FileSystemProvider fileSystemProvider = FileSystemProviders.installed().getFileSystemProvider(scheme);
-
             if (fileSystemProvider != null) {
                 return Files.newInputStream(fileSystemProvider.getPath(uri), options);
             }
-
             if ("http".equalsIgnoreCase(scheme) || "https".equalsIgnoreCase(scheme)) {
                 return get().toURL().openStream();
             }
-            
             return Files.newInputStream(getPath(), options);
         }
     }
--- a/src/main/java/org/apache/commons/io/build/AbstractOrigin.java
+++ b/src/main/java/org/apache/commons/io/build/AbstractOrigin.java
@@ -402,6 +402,13 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
             return Paths.get(get());
         }
 
+        @Override
+        public InputStream getInputStream(final OpenOption... options) throws IOException {
+            if ("http".equalsIgnoreCase(get().getScheme()) || "https".equalsIgnoreCase(get().getScheme())) {
+                return get().toURL().openStream();
+            }
+            return Files.newInputStream(getPath(), options);
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


garydgregory (changes requested):
@thachlp 
Thank you for your PR.
The PR does not test all the new code, specifically HTTP (as opposed to HTTPS).
Refactor duplicate method calls.
You should check if a file system is installed already that handles HTTP[S] before overriding NIO Path handling.
TY!

src/test/java/org/apache/commons/io/build/AbstractOriginTest.java
```diff
@@ -130,4 +133,12 @@ public void testGetWriter() throws IOException {
     public void testSize() throws IOException {
         assertEquals(Files.size(Paths.get(FILE_NAME_RO)), getOriginRo().getByteArray().length);
     }
+
+    @Test
+    public void testURI() throws Exception {
```
garydgregory, line 138: This test is in the wrong class, it does not do anything generic that would apply to all of its subclasses. Use `URIOriginTest`

thachlp: Yes, I updated.

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
```diff
@@ -402,6 +402,13 @@ public Path getPath() {
             return Paths.get(get());
         }
 
+        @Override
+        public InputStream getInputStream(final OpenOption... options) throws IOException {
+            if ("http".equalsIgnoreCase(get().getScheme()) || "https".equalsIgnoreCase(get().getScheme())) {
```
garydgregory, line 407: This method should only call `get()` and `getScheme()` once.

Don't override the file systems for these two schemes. You should check if a file system is installed already that handles HTTP[S] before overriding NIO Path handling; use `org.apache.commons.io.file.spi.FileSystemProviders`.

src/test/java/org/apache/commons/io/build/URIOriginTest.java
```diff
@@ -35,4 +39,21 @@ public void beforeEach() {
...
+    }
+
+    @Test
+    void testGetInputStreamHttps() throws Exception {
+        final String uri = "https://example.com";
+        final AbstractOrigin.URIOrigin origin = new AbstractOrigin.URIOrigin(new URI(uri));
+        try (final InputStream in = origin.getInputStream()) {
+            assertNotEquals(-1, in.read());
+        }
+    }
```
garydgregory, line 58: Add tests for 'http' just like you have 'https'. Use JUnit parameterized tests to reduce code duplication.

garydgregory (changes requested):
@thachlp 
Thank you for your updated. Please see my comments.

src/test/java/org/apache/commons/io/build/URIOriginTest.java
thachlp: Yes, updated

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
thachlp: Hope that I understand you correctly, please help review it 🙇

garydgregory:
https://issues.apache.org/jira/browse/IO-831

garydgregory:
@thachlp 
Thank you for your updates. Merged!

thachlp:
> @thachlp Thank you for your updates. Merged!

Thanks @garydgregory 🚀

PR #728, opened 2025-03-20
IO-870: PathUtils.copyFileToDirectory

dsmiley:
span FileSystem.

https://issues.apache.org/jira/browse/IO-870

> PathUtils.copyFileToDirectory is a simple method calling Files.copy, the latter of which supports cross-FileSystem copy.  But copyFileToDirectory gets the file name from the source as a Path and then resolves this against the target, which is invalid.  It needs to get the string form of the source file name.

This problem was encountered in Apache Solr which uses Apache Lucene's extensive test infrastructure, including custom FileSystem impls that emulate various behaviors to help tests see things that some users might experience.

### DIFF
--- a/src/main/java/org/apache/commons/io/file/PathUtils.java
+++ b/src/main/java/org/apache/commons/io/file/PathUtils.java
@@ -322,8 +322,8 @@ public final class PathUtils {
      */
     public static Path copyFileToDirectory(final Path sourceFile, final Path targetDirectory, final CopyOption... copyOptions) throws IOException {
         // Path.resolve() naturally won't work across FileSystem unless we convert to a String
-        Path sourceFileName = sourceFile.getFileName();
-        Path targetFile;
+        final Path sourceFileName = sourceFile.getFileName();
+        final Path targetFile;
         if (sourceFileName.getFileSystem() == targetDirectory.getFileSystem()) {
             targetFile = targetDirectory.resolve(sourceFileName);
         } else {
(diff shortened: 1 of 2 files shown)


dsmiley:
I might actually be able to test this using Zip FileSystem from a JAR to copy to a temp dir (standard/default FileSystem).  Could just use a dependency's JAR.  Hmmm.

garydgregory:
@dsmiley 

You did not follow the PR template:

- [ ] Read the [contribution guidelines](CONTRIBUTING.md) for this project.
- [ ] Run a successful build using the default [Maven](https://maven.apache.org/) goal with `mvn`; that's `mvn` on the command line by itself.
- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible but is a best-practice.
- [ ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [ ] Each commit in the pull request should have a meaningful subject line and body. Note that commits might be squashed by a maintainer on merge.

dsmiley:
I didn't run "mvn" but otherwise I believe I'm compliant with this projects many rules.  If there's still a problem after the tests pass but I haven't done something specifically, just let me know (which).

dsmiley:
At first I thought "test" phase might be enough (I pushed) -- it wasn't.  Then "verify" (I pushed) -- but apparently not!  The project should hookup "spotbugs" to verify.  It ran when I used "mvn" alone.

PR #732, opened 2025-03-23
IO-872: PathUtils.directoryAndFileContentEquals with different FileSystems

dsmiley:
[IO-872](https://issues.apache.org/jira/browse/IO-872)

### DIFF
--- a/src/main/java/org/apache/commons/io/file/PathUtils.java
+++ b/src/main/java/org/apache/commons/io/file/PathUtils.java
@@ -57,10 +57,10 @@ import java.time.chrono.ChronoZonedDateTime;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
@@ -133,12 +133,12 @@ public final class PathUtils {
                     } else {
                         tmpRelativeDirList1 = visitor1.relativizeDirectories(dir1, true, null);
                         tmpRelativeDirList2 = visitor2.relativizeDirectories(dir2, true, null);
-                        if (!tmpRelativeDirList1.equals(tmpRelativeDirList2)) {
+                        if (!equals(tmpRelativeDirList1, tmpRelativeDirList2)) {
                             equals = false;
                         } else {
                             tmpRelativeFileList1 = visitor1.relativizeFiles(dir1, true, null);
                             tmpRelativeFileList2 = visitor2.relativizeFiles(dir2, true, null);
-                            equals = tmpRelativeFileList1.equals(tmpRelativeFileList2);
+                            equals = equals(tmpRelativeFileList1, tmpRelativeFileList2);
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


garydgregory:
@dsmiley Please see the build failures.

dsmiley:
I'll get back to this eventually... I can see a Java 8 issue, and a Windows issue.

garydgregory:
@dsmiley 

OK, sounds good. We'll be here 😉

src/main/java/org/apache/commons/io/file/PathUtils.java
```diff
@@ -148,6 +148,51 @@ private RelativeSortedPaths(final Path dir1, final Path dir2, final int maxDepth
...
+                final Path pathA = listAIter.next();
+                final Path pathB = listBIter.next();
+                if (pathA.getFileSystem() == pathB.getFileSystem()) {
+                    if (!pathA.equals(pathB)) {
+                        return false;
+                    }
+                } else if (pathA.getFileSystem().getSeparator().equals(pathB.getFileSystem().getSeparator())) {
+                    // Separators are the same, so we can use toString comparison
+                    if (!pathA.toString().equals(pathB.toString())) {
+                        return false;
```
garydgregory, line 176: @dsmiley 
This doesn't (based on local testing) when you compare two directories like "foo" and "foo/".

garydgregory: Hm, the binary search in `directoryAndFileContentEquals` will also need a custom comparator or you get a `ProviderMismatchException` like:
```
java.nio.file.ProviderMismatchException
	at com.sun.nio.zipfs.ZipPath.checkPath(ZipPath.java:393)
	at com.sun.nio.zipfs.ZipPath.compareTo(ZipPath.java:620)
	at com.sun.nio.zipfs.ZipPath.compareTo(ZipPath.java:59)
	at java.util.Collections.indexedBinarySearch(Collections.java:228)
	at java.util.Collections.binarySearch(Collections.java:215)
	at org.apache.commons.io.file.PathUtils.directoryAndFileContentEquals(PathUtils.java:711)
	at org.apache.commons.io.file.PathUtils.directoryAndFileContentEquals(PathUtils.java:675)
	at org.apache.commons.io.file.PathUtilsContentEqualsTest.testDirectoryAndFileContentEqualsDifferentFileSystems(PathUtilsContentEqualsTest.java:115)
	at java.lang.reflect.Method.invoke(Method.java:503)
	at java.util.ArrayList.forEach(ArrayList.java:1259)
	at java.util.ArrayList.forEach(ArrayList.java:1259)
```

dsmiley: Hmm.  In my (limited) experience looking at Path with say mac/unix, there's never a trailing slash in path.toString().  (Maybe I could be wrong).   Maybe this isn't universally true... which sucks because Path is supposed to be an abstraction

garydgregory: @dsmiley 
It's worse than that: Path implementations are OS specific and separators are not normalized such that, for example, on macOS: `Paths.get("\\foo").normalize().equals(Paths.get("/foo").normalize())` returns `false`.

garydgregory:
Hello @dsmiley 

I think we can close this PR based on the state of git master.
WDYT?

garydgregory:
Closing: No feedback in months.

PR #740, opened 2025-04-15
[javadoc] Rename parameter of `ProxyOutputStream.write(int)`

jglick:
Very trivial, but I noticed that the parameter of this method was named `idx`, which would usually be construed as an abbreviation for `index`, which in this context would be presumed to be an index into a `byte[]`. But that is not what it is; it is a byte (typed as `int` for what I suppose are historical reasons). The supertypes `OutputStream` and `FilterOutputStream` name the parameter `b`, which seems more intuitive.


garydgregory:
@jglick 
Thank you for your PR 👍 Merged 🚀

PR #744, opened 2025-05-05
Handling null charset in `FileUtils.readLines(File, Charset)`

ryanmkurtz:
I recently upgraded to the latest commons-io and the code where we were passing in null Charsets blew up.  The javadoc for this method says null should be handled, and this seems to be how you handle it elsewhere in the file.


ryanmkurtz:
The issue seems to have originated at dcb09db922e1edbd6dc6cb1531b827451d58aa83 in version 2.12.  We happened to upgraded from 2.11.

garydgregory (changes requested):
@ryanmkurtz 
Thank you for the PR. Please add a unit test to avoid regressions in the future.

ryanmkurtz:
@garydgregory I added a new test.

PR #748, opened 2025-05-17
QueueInputStream reads all but the first byte without waiting.

maxxedev:
```
QueueInputStream reads all but the first byte without waiting.

Fix so that bulk reads avoid getting stuck if a timeout is set and at least one byte is available.
```

### DIFF
--- a/src/main/java/org/apache/commons/io/input/QueueInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/QueueInputStream.java
@@ -224,4 +224,49 @@ public class QueueInputStream extends InputStream {
         }
     }
 
+    /**
+     * Reads up to {@code length} bytes of data from the input stream into
+     * an array of bytes.  The first byte is read while honoring the timeout; the rest are read while <i>not</i> honoring
+     * the timeout. The number of bytes actually read is returned as an integer.
+     *
+     * @param b     the buffer into which the data is read.
+     * @param offset   the start offset in array {@code b}
+     *                   at which the data is written.
+     * @param length   the maximum number of bytes to read.
+     * @return     the total number of bytes read into the buffer, or
+     *             {@code -1} if there is no more data because the end of
+     *             the stream has been reached.
+     * @throws IllegalStateException if thread is interrupted while waiting for the first byte.
+     */
+    @Override
+    public int read(byte[] b, int offset, int length) {
+        if (offset > b.length || offset < 0) {
+            throw new ArrayIndexOutOfBoundsException("Offset out of bounds: " + offset);
+        }
+        if (length < 0 || length > b.length - offset) {
+            throw new ArrayIndexOutOfBoundsException("Length out of bounds: " + length);
+        }
+
+        // read the first value, honoring the timeout
+        final int firstValue = read();
+        if (firstValue == EOF) {
+            return EOF;
+        }
+
+        int i = 0;
+        b[offset + i] = (byte) (0xFF & firstValue);
+        i++;
+
+        for (; i < length; i++) {
+            // read the rest, ignoring the timeout
+            Integer value = blockingQueue.poll();
+            value = value == null ? EOF : 0xFF & value;
+            if (value == EOF) {
+                return i;
+            }
+            b[offset + i] = (byte) (0xFF & value);
+        }
+        return i;
+    }
+
 }
--- a/src/test/java/org/apache/commons/io/input/QueueInputStreamTest.java
+++ b/src/test/java/org/apache/commons/io/input/QueueInputStreamTest.java
@@ -16,6 +16,7 @@
  */
 package org.apache.commons.io.input;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTimeout;
@@ -23,10 +24,15 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.BufferedInputStream;
 import java.io.BufferedOutputStream;
+import java.io.BufferedReader;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.time.Duration;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
@@ -123,11 +129,50 @@ public class QueueInputStreamTest {
     @MethodSource("inputData")
     public void testBufferedReads(final String inputData) throws IOException {
         final BlockingQueue<Integer> queue = new LinkedBlockingQueue<>();
-        try (BufferedInputStream inputStream = new BufferedInputStream(new QueueInputStream(queue));
-                QueueOutputStream outputStream = new QueueOutputStream(queue)) {
+        try (BufferedInputStream inputStream = new BufferedInputStream(new QueueInputStream(queue)); QueueOutputStream outputStream = new QueueOutputStream(queue)) {
             outputStream.write(inputData.getBytes(StandardCharsets.UTF_8));
             final String actualData = IOUtils.toString(inputStream, StandardCharsets.UTF_8);
             assertEquals(inputData, actualData);
+        }
+    }
+
+    @ParameterizedTest(name = "inputData={0}")
+    @MethodSource("inputData")
+    public void testReadLineByLineQueue(final String inputData) throws IOException {
+        final String[] lines = inputData.split("\n");
+        final BlockingQueue<Integer> queue = new LinkedBlockingQueue<>();
+        try (InputStream inputStream = QueueInputStream.builder().setBlockingQueue(queue)
+                                                       .setTimeout(Duration.ofHours(1))
+                                                       .get();
+             QueueOutputStream outputStream = new QueueOutputStream(queue)) {
+
+            doTestReadLineByLine(inputData, inputStream, outputStream);
+        }
+    }
+
+    @ParameterizedTest(name = "inputData={0}")
+    @MethodSource("inputData")
+    public void testReadLineByLineFile(final String inputData) throws IOException {
+        final Path tempFile = Files.createTempFile(getClass().getSimpleName(), ".txt");
+        try (InputStream inputStream = Files.newInputStream(tempFile);
+             OutputStream outputStream = Files.newOutputStream(tempFile)) {
+
+            doTestReadLineByLine(inputData, inputStream, outputStream);
+        } finally {
+            Files.delete(tempFile);
+        }
+    }
+
+    private void doTestReadLineByLine(final String inputData, final InputStream inputStream, final OutputStream outputStream) throws IOException {
+        final String[] lines = inputData.split("\n");
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, UTF_8))) {
+            for (String line : lines) {
+                outputStream.write(line.getBytes(UTF_8));
+                outputStream.write('\n');
+
+                String actualLine = reader.readLine();
+                assertEquals(line, actualLine);
+            }
         }
     }
 
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -47,6 +47,7 @@ The <action> type attribute can be add,update,fix,remove.
   <body>
     <release version="2.20.0" date="YYYY-MM-DD" description="Version 2.19.1: Java 8 or later is required.">
       <!-- FIX -->
+      <action dev="maxxedev" type="fix"                due-to="maxxedev">QueueInputStream reads all but the first byte without waiting</action>
       <action dev="ggregory" type="fix"                due-to="Jesse Glick">[javadoc] Rename parameter of ProxyOutputStream.write(int) #740.</action>
       <action dev="ggregory" type="fix" issue="IO-875" due-to="Pierre Baumard, Gary Gregory">CopyDirectoryVisitor ignores fileFilter #743.</action>
       <action dev="ggregory" type="fix"                due-to="Gary Gregory">org.apache.commons.io.build.AbstractOrigin.getReader(Charset) now maps a null Charset to the default Charset.</action>


src/test/java/org/apache/commons/io/input/QueueInputStreamTest.java
```diff
@@ -123,14 +129,53 @@ public void testAvailableAfterOpen(final String inputData) throws IOException {
...
 
+    @ParameterizedTest(name = "inputData={0}")
+    @MethodSource("inputData")
+    public void testReadLineByLineQueue(final String inputData) throws IOException {
+        final String[] lines = inputData.split("\n");
+        final BlockingQueue<Integer> queue = new LinkedBlockingQueue<>();
+        try (InputStream inputStream = QueueInputStream.builder().setBlockingQueue(queue)
+                                                       .setTimeout(Duration.ofHours(1))
+                                                       .get();
+             QueueOutputStream outputStream = new QueueOutputStream(queue)) {
```
ppkarwasz, line 147: _Nitpick_: by using `QueueInputStream.newQueueOutputStream()` we gain one more line of coverage:

```suggestion
        try (QueueInputStream inputStream = QueueInputStream.builder().setBlockingQueue(queue)
                                                            .setTimeout(Duration.ofHours(1))
                                                            .get();
             OutputStream outputStream = inputStream.newQueueOutputStream()) {
```

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -224,4 +224,49 @@ public int read() {
...
+        i++;
+
+        for (; i < length; i++) {
+            // read the rest, ignoring the timeout
+            Integer value = blockingQueue.poll();
+            value = value == null ? EOF : 0xFF & value;
+            if (value == EOF) {
+                return i;
+            }
+            b[offset + i] = (byte) (0xFF & value);
```
ppkarwasz, line 267:
```suggestion
            if (value == null) {
                return i;
            }
            b[offset + i] = (byte) (0xFF & value);
```

ppkarwasz (approved):
@maxxedev,

Thanks for the pull request — it adds valuable functionality and includes all the essential elements.

A couple of suggestions and minor nitpicks below.

**Note:** Have you considered using [`BlockingQueue.drainTo`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/BlockingQueue.html#drainTo-java.util.Collection-int-) instead of repeated `poll()` calls? It locks the queue only once, which could improve performance.  
The tradeoff would be representing a `byte[]` slice as a `List<Integer>`.

ppkarwasz:
I ran a small benchmark (also included in 35d1ce4893a979745f98372fd3bb7fed85679093) that measures the time it takes to read 1 MiB from the queue, while another thread is writing to it. The results look promising.

### Before

| Benchmark                                  | Mode   | Cnt | Score   | Error   | Units |
|--------------------------------------------|--------|-----|---------|---------|-------|
| QueueStreamBenchmark.streams:input         | sample | 639 | 78.261  | ± 1.892 | ms/op |
| QueueStreamBenchmark.streams:input:p0.00   | sample |     | 60.293  |         | ms/op |
| QueueStreamBenchmark.streams:input:p0.50   | sample |     | 69.861  |         | ms/op |
| QueueStreamBenchmark.streams:input:p0.90   | sample |     | 96.600  |         | ms/op |
| QueueStreamBenchmark.streams:input:p0.95   | sample |     | 104.858 |         | ms/op |
| QueueStreamBenchmark.streams:input:p0.99   | sample |     | 123.627 |         | ms/op |
| QueueStreamBenchmark.streams:input:p0.999  | sample |     | 156.500 |         | ms/op |
| QueueStreamBenchmark.streams:input:p0.9999 | sample |     | 156.500 |         | ms/op |
| QueueStreamBenchmark.streams:input:p1.00   | sample |     | 156.500 |         | ms/op |

### After

**Note**: The high values in some benchmark runs are due to a synchronization problem between the two threads, which led the `QueueInputStream` thread to time out.

| Benchmark                                  | Mode   | Cnt  | Score    | Error   | Units |
|--------------------------------------------|--------|------|----------|---------|-------|
| QueueStreamBenchmark.streams:input         | sample | 9577 | 21.248   | ± 5.848 | ms/op |
| QueueStreamBenchmark.streams:input:p0.00   | sample |      | 3.650    |         | ms/op |
| QueueStreamBenchmark.streams:input:p0.50   | sample |      | 12.222   |         | ms/op |
| QueueStreamBenchmark.streams:input:p0.90   | sample |      | 12.386   |         | ms/op |
| QueueStreamBenchmark.streams:input:p0.95   | sample |      | 12.468   |         | ms/op |
| QueueStreamBenchmark.streams:input:p0.99   | sample |      | 33.787   |         | ms/op |
| QueueStreamBenchmark.streams:input:p0.999  | sample |      | 2292.976 |         | ms/op |
| QueueStreamBenchmark.streams:input:p0.9999 | sample |      | 8128.561 |         | ms/op |
| QueueStreamBenchmark.streams:input:p1.00   | sample |      | 8128.561 |         | ms/op |

src/test/java/org/apache/commons/io/input/QueueInputStreamTest.java
ppkarwasz: Fixed in e91989163f7a5e856ae1b168b00c14ef49b8d8fe

src/main/java/org/apache/commons/io/input/QueueInputStream.java
ppkarwasz: Outdated after the changes in e91989163f7a5e856ae1b168b00c14ef49b8d8fe

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -224,4 +224,49 @@ public int read() {
...
+     * @throws IllegalStateException if thread is interrupted while waiting for the first byte.
+     */
+    @Override
+    public int read(byte[] b, int offset, int length) {
+        if (offset > b.length || offset < 0) {
+            throw new ArrayIndexOutOfBoundsException("Offset out of bounds: " + offset);
+        }
+        if (length < 0 || length > b.length - offset) {
+            throw new ArrayIndexOutOfBoundsException("Length out of bounds: " + length);
+        }
```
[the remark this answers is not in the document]
ppkarwasz: Fixed in e91989163f7a5e856ae1b168b00c14ef49b8d8fe

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -224,4 +224,49 @@ public int read() {
...
+     * the timeout. The number of bytes actually read is returned as an integer.
+     *
+     * @param b     the buffer into which the data is read.
+     * @param offset   the start offset in array {@code b}
+     *                   at which the data is written.
+     * @param length   the maximum number of bytes to read.
+     * @return     the total number of bytes read into the buffer, or
+     *             {@code -1} if there is no more data because the end of
+     *             the stream has been reached.
+     * @throws IllegalStateException if thread is interrupted while waiting for the first byte.
```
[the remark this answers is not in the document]
ppkarwasz: Fixed in e91989163f7a5e856ae1b168b00c14ef49b8d8fe

ppkarwasz (approved):
Thanks for the changes — using `BlockingQueue.drainTo` significantly reduces lock contention and improves performance ~4x.

We might optimize further by reordering operations. For `length > 1`, we could:

1. Try `drainTo()` first — if it returns elements, skip the next point.
2. Otherwise, call `read()` for one byte, then `drainTo()` for the rest.
3. Copy data from `List<Integer>` to the byte array.

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -224,4 +226,52 @@ public int read() {
...
+                return EOF;
+            }
+            b[offset + i] = (byte) (0xFF & firstValue);
+            i++;
+        }
+
+        if (length > 1) {
+            // read the rest, ignoring the timeout
+            final List<Integer> drain = new ArrayList<>(length * 2);
+            blockingQueue.drainTo(drain, length - 1);
```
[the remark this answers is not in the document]
maxxedev: > Why is the initial capacity of the list twice the number of expected elements?

Ah, I got confused with HashMap loadFactor. ArrayList does not need such buffer
&nbsp;

> Would it make sense to store the list in a field to avoid recreating it on each call? Unless I'm missing something, QueueInputStream doesn't appear to support concurrent reads, so we might not need to worry about thread safety in this case.

I would prefer to keep it simple and also maintain strict thread-safety. The javadoc says: "queue input/output streams may be used safely in a single thread or multiple threads". In practice, that means QueueInputStream and QueueOutputStream _together_ can be used concurrently (i.e. one thread reading and one thread writing). There is no practical reason someone would want to read from QueueInputStream concurrently (i.e. two threads reading). Regardless I want to avoid having to document such theoretical thread un-safety and avoid confusing the users.

garydgregory (commented):
Hi All,

I've made a few minor changes; see the commit comments.

The remaining major issue is why the new `read(...)` method ignores the configured timeout some of the time and not some other times? 

From a user's POV this is quite confusing and random behavior: I construct an instance with a timeout, I pass it on to other APIs, and sometimes the timeout applies and sometimes it doesn't. 

Isn't this bound to be a source of bug reports and confusion?

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -224,4 +226,48 @@ public int read() {
...
+     * @throws IndexOutOfBoundsException if {@code offset} is negative, {@code length} is negative, or {@code length} is
+     *             greater than {@code b.length - offset}.
+     * @since 2.20.0
+     */
+    @Override
+    public int read(final byte[] b, final int offset, final int length) {
+        if (b == null) {
+            throw new NullPointerException();
+        } else if (offset < 0 || length < 0 || length > b.length - offset) {
+            throw new IndexOutOfBoundsException();
```
ppkarwasz, line 250: _Nitpick_: Let us use the same message as the overridden method. See also https://github.com/apache/commons-io/pull/748#discussion_r2094044658

```suggestion
            throw new IndexOutOfBoundsException(
                    String.format("Range [%d, %<d + %d) out of bounds for length %d", offset, length, b.length));
```

ppkarwasz (approved):

src/main/java/org/apache/commons/io/input/QueueInputStream.java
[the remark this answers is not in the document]
ppkarwasz: I updated the suggestion in https://github.com/apache/commons-io/pull/748#discussion_r2094562967

src/main/java/org/apache/commons/io/input/QueueInputStream.java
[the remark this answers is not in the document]
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/748/commits/a2c080ceab3a2a0153cf3e41a6ff2b4a6a9f2037

ppkarwasz:
> The remaining major issue is why the new `read(...)` method ignores the configured timeout some of the time and not some other times?

From my perspective, the intended contract of the `read(...)` method is to attempt to read **at least one byte** from the queue, waiting **up to the configured timeout** for data to become available. If no data arrives within that period, it should return `-1`.
As far as I can tell, the current implementation adheres to this contract. It waits for the timeout at most **once**, not per byte, which ensures consistent and predictable behavior.

In contrast, the previous implementation—inherited from the super class—effectively applied the timeout **per byte requested**—so if, for example, the caller attempted to read into a buffer of size 1000, the method could block for up to 1000 times the configured timeout. This behavior could easily lead to unexpectedly long delays and violated the "at most the configured timeout" expectation.

maxxedev:
> The remaining major issue is why the new read(...) method ignores the configured timeout some of the time and not some other times?

In addition to what [`@ppkarwasz` said above](https://github.com/apache/commons-io/pull/748#issuecomment-2889075382), see [these tests](https://github.com/apache/commons-io/blob/1ae78339d10e0b9f65fa7c7c0dbd0c489cb31c25/src/test/java/org/apache/commons/io/input/QueueInputStreamTest.java#L146-L185) that compare behavior of QueueInputStream and FileInputStream as reference. The previous implementation would block for long durations even though there was some data always available. The new implementation and reference implementation don't block if data is available.

src/main/java/org/apache/commons/io/input/QueueInputStream.java
garydgregory: @ppkarwasz 
But that's _not_ what's in the superclass in Java 8. On Java 11, if you drill down, you end up in `jdk.internal.util.Preconditions.outOfBoundsMessage(String, List<Integer>)` which has the kind of message you show. I imagine subsequent Java LTS versions might yet do something else. 
Which one should we pick? If Java 11 gives more detail they must have decided that there are no security implications, so I'd be OK with that version.

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -224,4 +226,49 @@ public int read() {
...
+    public int read(final byte[] b, final int offset, final int length) {
+        if (b == null) {
+            throw new NullPointerException();
+        } else if (offset < 0 || length < 0 || length > b.length - offset) {
+            throw new IndexOutOfBoundsException(
+                    String.format("Range [%d, %<d + %d) out of bounds for length %d", offset, length, b.length));
+        } else if (length == 0) {
+            return 0;
+        }
+        final List<Integer> drain = new ArrayList<>(length);
```
garydgregory, line 255: Does it make sense to lower the memory allocation request like this:
```suggestion
        final List<Integer> drain = new ArrayList<>(Math.min(length, blockingQueue.size()));
```
?

src/main/java/org/apache/commons/io/input/QueueInputStream.java
ppkarwasz: Good point!

src/test/java/org/apache/commons/io/input/QueueInputStreamTest.java
```diff
@@ -124,13 +136,124 @@ public void testAvailableAfterOpen(final String inputData) throws IOException {
...
+                .setTimeout(Duration.ofHours(1))
+                .get()) {
+            final QueueOutputStream queueOutputStream = queueInputStream.newQueueOutputStream();
+            CompletableFuture.runAsync(() -> {
+                try {
+                    onPollLatch.await();
+                    queueOutputStream.write(inputData.getBytes(StandardCharsets.UTF_8));
+                    afterWriteLatch.countDown();
+                } catch (final Exception e) {
+                    throw new RuntimeException(e);
```
garydgregory, line 245: Hello @maxxedev 
Unless you expect this exception for the test to pass, the test should be clearer and call JUnit's `fail(...)` method.

garydgregory (changes requested):

src/test/java/org/apache/commons/io/input/QueueInputStreamTest.java
```diff
@@ -124,13 +136,124 @@ public void testAvailableAfterOpen(final String inputData) throws IOException {
...
+    public void testBulkReadZeroLength() {
+        final QueueInputStream queueInputStream = new QueueInputStream();
+        final int read = queueInputStream.read(EMPTY_BYTE_ARRAY, 0, 0);
+        assertEquals(0, read);
+    }
+
+    @ParameterizedTest(name = "inputData={0}")
+    @MethodSource("inputData")
+    public void testBulkReadWaiting(final String inputData) throws IOException {
+        assumeTrue(!inputData.isEmpty());
```
garydgregory, line 219: Simplify with:
```
assumeFalse(inputData.isEmpty());
```

src/main/java/org/apache/commons/io/input/QueueInputStream.java
```diff
@@ -224,4 +226,49 @@ public int read() {
...
+        final List<Integer> drain = new ArrayList<>(Math.min(length, blockingQueue.size()));
+        blockingQueue.drainTo(drain, length);
+        if (drain.isEmpty()) {
+            // no data immediately available. wait for first byte
+            final int value = read();
+            if (value == EOF) {
+                return EOF;
+            }
+            drain.add(value);
+            blockingQueue.drainTo(drain, length - 1);
```
garydgregory, line 264: I think this call to `drainTo(...)` could be wrong because it can cause EOF markers to end up in the `drain` data.
We want to drain until we get the requested length _or_ the EOF marker is seen in the drained data.
For example:
- Thread A is paused just before, on, or during the `drain.add(value)` above, but before `blockingQueue.drainTo(drain, length - 1);`
- Thread B adds a `-1` (an EOF marker).
- Thread A resumes, calls `blockingQueue.drainTo(drain, length - 1);` and now contains the EOF marker in `drain` and includes it in `b`.

ppkarwasz: The queue itself does not contain any `-1` elements:

https://github.com/apache/commons-io/blob/5080fa310dc19c20f9631bd18cfc3a5a63ef160e/src/main/java/org/apache/commons/io/output/QueueOutputStream.java#L92-L101

The `EOF` value that appears on lines 260 and 261 is generated by the `read()` method to indicate a missing value.

src/test/java/org/apache/commons/io/input/QueueInputStreamTest.java
ppkarwasz: This code is executed asynchronously, so any exception it throws will not be captured by the test.

Since, by the end of the test case, this code must execute, it would be useful to:

1. Assign the return value of `CompletableFuture.runAsync()` to a variable (e.g. `future`).
2. Call `assertDoesNotThrow(future::get)` at the very end of the test case.

**Note**: Since other parts of the test count on the `afterWriteLatch.countDown()` call, if something happens here, the test case will simply hang. I think that we should add some reasonable timeouts to the `CountDownLatch.wait()` calls.

src/test/java/org/apache/commons/io/input/QueueInputStreamTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/748/commits/e315a98906d2d75dcf412abc9cef0126e6d53447

src/main/java/org/apache/commons/io/input/QueueInputStream.java
maxxedev: That's an interesting question on whether to specially treat -1 in the underlying data stream. Looking at the behavior of PipedInputStream and ByteArrayInputStream, no special meaning is attached to -1. If there is a -1 and if there is data subsequently available, all data is returned as is.

I think QueueInputStream should behave similarly.

maxxedev:
@garydgregory what needs to be done to get the PR merged?

garydgregory:
I'll take a look again tonight or tomorrow morning.

PR #757, opened 2025-07-03
[IO-279] Add Tailer ignoreTouch option 

JoergBudi:
…oose to ignore touching of watched file

So I got late to the old io-279 Tailer issue. In the long history, some changes were already made long ago.
However from reading the last comments, some users want to avoid to get the whole file replayed 
when the timestamp of the watched file changes without a content update. I added an ignoreTouch option, the default is false and resembles current behaviour (file gets replayed) (inspired from the  2015 patch by gh user myyron).
When changed to true via the Builder pattern, a newertimestamp with constant content gets ignored.
That makes the Tailer more similar to "tail -f" (issuing touch on the watched file is ignored then), but note that subtle differences remain. When you copy the identical file onto watched  file, "tail -f" realizes that and replays the whole file. With ignoreTouch=true, the Tailer ignores that, as it looks identical to touching. 

Added 2 Junit tests to verify both behaviours explictly.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/Tailer.java
+++ b/src/main/java/org/apache/commons/io/input/Tailer.java
@@ -323,6 +323,7 @@ public class Tailer implements Runnable, AutoCloseable {
          *        ignoreTouch=true does nothing in that case.
          *
          * @return {@code this} instance.
+         * @since 2.20.0
          */
         public Builder setIgnoreTouch(final boolean ignoreTouch) {
             this.ignoreTouch = ignoreTouch;
--- a/src/main/java/org/apache/commons/io/input/Tailer.java
+++ b/src/main/java/org/apache/commons/io/input/Tailer.java
@@ -138,6 +138,7 @@ import org.apache.commons.io.file.attribute.FileTimes;
  *        VFS</a>.
  */
 public class Tailer implements Runnable, AutoCloseable {
+    private static final boolean DEFAULT_IGNORE_TOUCH = false;
 
     // @formatter:off
     /**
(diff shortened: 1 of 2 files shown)


garydgregory:
Jira: https://issues.apache.org/jira/browse/IO-279

src/main/java/org/apache/commons/io/input/Tailer.java
```diff
@@ -310,6 +312,23 @@ public Builder setTailFromEnd(final boolean end) {
...
+         * Sets ignoreTouch behaviour
+         *
+         * @param ignoreTouch This can be useful when your watched file gets touched (meaning it gets more recent timestamps
+         *        without changing the file) for some reason or when you are working on file systems where timestamp
+         *        is updated before content.
+         *        The default behaviour (ignoreTouch=false) would then reissue the whole current file, while
+         *        ignoreTouch=true does nothing in that case.
+         *
+         * @return {@code this} instance.
+         * @since 2.20.0
```
garydgregory, line 326: I added the missing Javadoc tag.

src/main/java/org/apache/commons/io/input/Tailer.java
```diff
@@ -695,6 +714,8 @@ public static Tailer create(final File file, final TailerListener listener, fina
      */
     private volatile boolean run = true;
 
+    private boolean ignoreTouch = DEFAULT_IGNORE_TOUCH;
```
garydgregory, line 717: Should be: `private final boolean ignoreTouch;`

garydgregory:
TY @JoergBudi !
Merged.

src/main/java/org/apache/commons/io/input/Tailer.java
JoergBudi: thx, will remember that.

src/main/java/org/apache/commons/io/input/Tailer.java
JoergBudi: yup, correct, thx for changing.

JoergBudi:
@garydgregory I will work on the unstable test and try not to slow it down for the other platforms, I'll  provide a new PR for that soon.

PR #758, opened 2025-07-03
Fix infinite loop in AbstractByteArrayOutputStream.

qwerty4030:
Fix infinite loop in AbstractByteArrayOutputStream.
When an AbstractByteArrayOutputStream is initialized with a 0 size buffer and writeImpl(InputStream) is called an infinite loop may occur and eventually cause an OOM exception.
This is due to passing 0 as the maximum bytes to read in InputStream#read which returns 0, then a new 0 size buffer is created, and then InputStream#read is called again therefore repeating the loop forever.
Fix this infinite loop by updating needNewBuffer to use DEFAULT_SIZE as the initial buffer size instead of 0.
This will allow InputStream#read to properly populate the buffer as expected and terminate the loop when EOF is reached.
Set initial value of currentBufferIndex to -1 so it is properly initialized to 0 when needNewBuffer is called; unrelated to the infinite loop bug.

### DIFF
--- a/src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java
+++ b/src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java
@@ -92,7 +92,7 @@ public abstract class AbstractByteArrayOutputStream<T extends AbstractByteArrayO
     private byte[] currentBuffer;
 
     /** The index of the current buffer. */
-    private int currentBufferIndex;
+    private int currentBufferIndex = -1;
 
     /** The total count of bytes in all the filled buffers. */
     private int filledBufferSum;
@@ -147,7 +147,8 @@ public abstract class AbstractByteArrayOutputStream<T extends AbstractByteArrayO
             // Creating new buffer
             final int newBufferSize;
             if (currentBuffer == null) {
-                newBufferSize = newCount;
+                // prevents 0 size buffers
+                newBufferSize = newCount > 0 ? newCount : DEFAULT_SIZE;
                 filledBufferSum = 0;
             } else {
                 newBufferSize = Math.max(currentBuffer.length << 1, newCount - filledBufferSum);
--- a/src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
+++ b/src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
@@ -28,12 +28,15 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.stream.Stream;
 
 import org.apache.commons.io.IOUtils;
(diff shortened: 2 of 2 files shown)


src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java
```diff
@@ -92,7 +92,7 @@ protected interface InputStreamConstructor<T extends InputStream> {
     private byte[] currentBuffer;
 
     /** The index of the current buffer. */
-    private int currentBufferIndex;
+    private int currentBufferIndex = -1;
```
qwerty4030, line 95: unrelated to the bug but I found this value was wrong while investigating. I dont think its causing any issues because it appears to only be used after `reset` is called which properly sets this value.

src/main/java/org/apache/commons/io/output/AbstractByteArrayOutputStream.java
```diff
@@ -147,7 +147,8 @@ protected void needNewBuffer(final int newCount) {
             // Creating new buffer
             final int newBufferSize;
             if (currentBuffer == null) {
-                newBufferSize = newCount;
+                // prevents 0 size buffers
+                newBufferSize = newCount > 0 ? newCount : DEFAULT_SIZE;
```
qwerty4030, line 151: simplest fix I found. Another option is to not allow ANY buffers smaller than DEFAULT_SIZE. Some things I ran into trying other fixes:

1.  if the logic allows an initial 0 size buffer; the next buffer must be >0 size. Then when `toBufferedInputStream` is called the tests fail because `.available()` returns 0 (for the initial buffer) instead of the actual length of the data in the next buffer. Also seems like a waste to even call `InputStream.read` with a 0 size buffer.
2. Technically there is nothing wrong with a BAOS with an initial size of 0, so we don't want to throw an exception. The docs say the buffer is increased as needed so hopefully it's fine to default it to `DEFAULT_SIZE` in this case.

src/test/java/org/apache/commons/io/output/ByteArrayOutputStreamTest.java
```diff
@@ -198,7 +205,23 @@ void testToBufferedInputStreamEmpty(final String baosName, final IOFunction<Inpu
...
+        }
+    }
+
+    @ParameterizedTest(name = "[{index}] {0}")
+    @MethodSource("toBufferedInputStreamFunctionFactories")
+    void testToBufferedInputStreamEmptyFile(final String baosName, final IOFunction<InputStream, InputStream> toBufferedInputStreamFunction,
+                                            final @TempDir Path temporaryFolder) throws IOException {
+        final Path emptyFile = Files.createTempFile(temporaryFolder, getClass().getSimpleName(), "-empty.txt");
+
+        try (InputStream is = Files.newInputStream(emptyFile)) {
```
qwerty4030, line 218: interesting corner case if you have an empty `ByteArrayInputStream` and you call `read` with a 0 size buffer it will return `EOF` BUT if you have an empty `FileInputStream` in that same scenario it will return 0.

garydgregory:
TY @qwerty4030 

Good find 👍, merged.

PR #760, opened 2025-07-05
[io-279] fixing macos-x-13-java8 unstable test due to posix file time…

JoergBudi:
… resolution is only 1s. Adding TestUtils.sleepTillNextFullSecond()

@garydgregory 
followup to PR https://github.com/apache/commons-io/pull/757 . The test was unstable, as the java.nio.file.Files.getLastModifiedTime() on macosx-13-java8 only has a time resolution of 1s (and it is used in Tailer.TailablePath.lastModifiedFIleTIme). 

Note: it seems, that java.io.File.lastModified() has  millisecond resolution on that platform, but it doesn't make sense to change the real code for everyone because of an odd platform.

So tests are now adapated, that the file date is updated after the curent second has passed to comply with posix standard Additionaly the waittime is shortened a little by polling instead of waiting max time.

see https://github.com/JoergBudi/commons-io/actions/runs/16092357707/job/45410764092 for passed test.

### DIFF
--- a/src/test/java/org/apache/commons/io/input/TailerTest.java
+++ b/src/test/java/org/apache/commons/io/input/TailerTest.java
@@ -724,19 +724,29 @@ class TailerTest {
 
             // Write some lines to the file
             writeLines(file, "Line one");
-            final long testDelayMillis = delayMillis * 10;
-            TestUtils.sleep(testDelayMillis);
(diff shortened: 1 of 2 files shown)


garydgregory:
@JoergBudi 

Thank you for the follow-up PR 👍

PR #776, opened 2025-09-04
feat: Add incremental `toByteArray` method

ppkarwasz:
This introduces `toByteArray(InputStream input, int size, int bufferSize)`, which reads the stream in chunks of `bufferSize` instead of allocating the full array up front.

By reading incrementally, the method:

* Validates that the stream actually contains `size` bytes before completing the allocation.
* Prevents excessive memory usage if a corrupted or malicious `size` value is provided.
* Offers safer handling for untrusted input compared to the direct-allocation variant.

- [x] I used AI to create the Javadoc.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -57,6 +57,7 @@ The <action> type attribute can be add,update,fix,remove.
       <action dev="ggregory" type="add"                due-to="Gary Gregory">Add org.apache.commons.io.output.ProxyOutputStream.writeRepeat(byte[], int, int, long).</action>
       <action dev="ggregory" type="add"                due-to="Gary Gregory">Add org.apache.commons.io.output.ProxyOutputStream.writeRepeat(byte[], long).</action>
       <action dev="ggregory" type="add"                due-to="Gary Gregory">Add org.apache.commons.io.output.ProxyOutputStream.writeRepeat(int, long).</action>
+      <action dev="pkarwasz" type="add"                due-to="Piotr P. Karwasz">Added toByteArray(InputStream, int, int) for safer incremental reading with size validation.</action>
       <!-- UPDATE -->
       <action type="update" dev="ggregory"             due-to="Gary Gregory, Dependabot">Bump org.apache.commons:commons-parent from 85 to 87 #774.</action>
       <action type="update" dev="ggregory"             due-to="Gary Gregory">[test] Bump commons-codec:commons-codec from 1.18.0 to 1.19.0.</action>
--- a/src/main/java/org/apache/commons/io/IOUtils.java
+++ b/src/main/java/org/apache/commons/io/IOUtils.java
@@ -65,6 +65,7 @@ import org.apache.commons.io.channels.FileChannels;
 import org.apache.commons.io.function.IOConsumer;
 import org.apache.commons.io.function.IOSupplier;
 import org.apache.commons.io.function.IOTriFunction;
+import org.apache.commons.io.input.BoundedInputStream;
 import org.apache.commons.io.input.CharSequenceReader;
 import org.apache.commons.io.input.QueueInputStream;
 import org.apache.commons.io.output.AppendableWriter;
@@ -2659,37 +2660,60 @@ public class IOUtils {
     }
 
     /**
-     * Gets the contents of an {@link InputStream} as a {@code byte[]}. Use this method instead of
-     * {@link #toByteArray(InputStream)} when {@link InputStream} size is known.
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
      *
-     * @param input the {@link InputStream} to read.
-     * @param size the size of {@link InputStream} to read, where 0 &lt; {@code size} &lt;= length of input stream.
-     * @return byte [] of length {@code size}.
-     * @throws IOException if an I/O error occurs or {@link InputStream} length is smaller than parameter {@code size}.
-     * @throws IllegalArgumentException if {@code size} is less than zero.
+     * <p>
+     *   This variant allocates the target array immediately and attempts to fill it in one pass.
+     *   It assumes that {@code size} is correct.
+     *   If the stream ends prematurely, an {@link EOFException} is thrown.
+     * </p>
+     *
+     * <p>
+     *   <strong>Important:</strong> This method does <em>not</em> defend against corrupted
+     *   or untrusted {@code size} values.
+     *   For untrusted input, use {@link #toByteArray(InputStream, int, int)} instead,
+     *   which validates that the stream contains at least {@code size} bytes before allocating the target array.
+     * </p>
+     *
+     * @param input the {@link InputStream} to read; must not be {@code null}.
+     * @param size  the exact number of bytes to read; must be {@code >= 0}.
+     * @return a new byte array of length {@code size}.
+     * @throws IllegalArgumentException if {@code size} is negative.
+     * @throws EOFException             if the stream ends before {@code size} bytes are read.
+     * @throws IOException              if an I/O error occurs while reading.
+     * @throws NullPointerException     if {@code input} is {@code null}.
      * @since 2.1
      */
     public static byte[] toByteArray(final InputStream input, final int size) throws IOException {
-        if (size == 0) {
-            return EMPTY_BYTE_ARRAY;
+        Objects.requireNonNull(input, "input");
+        if (size < 0) {
+            throw new IllegalArgumentException("Size must be equal or greater than zero: " + size);
         }
-        return toByteArray(Objects.requireNonNull(input, "input")::read, size);
+        return toByteArray(input::read, size);
     }
 
     /**
-     * Gets contents of an {@link InputStream} as a {@code byte[]}.
-     * Use this method instead of {@link #toByteArray(InputStream)}
-     * when {@link InputStream} size is known.
-     * <strong>NOTE:</strong> the method checks that the length can safely be cast to an int without truncation
-     * before using {@link IOUtils#toByteArray(InputStream, int)} to read into the byte array.
-     * (Arrays can have no more than Integer.MAX_VALUE entries anyway.)
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
      *
-     * @param input the {@link InputStream} to read.
-     * @param size the size of {@link InputStream} to read, where 0 &lt; {@code size} &lt;= min(Integer.MAX_VALUE, length of input stream).
-     * @return byte [] the requested byte array, of length {@code size}.
-     * @throws IOException              if an I/O error occurs or {@link InputStream} length is less than {@code size}.
-     * @throws IllegalArgumentException if size is less than zero or size is greater than Integer.MAX_VALUE.
-     * @see IOUtils#toByteArray(InputStream, int)
+     * <p>
+     *   This is a convenience overload of {@link #toByteArray(InputStream, int, int)} that accepts a
+     *   {@code long} size parameter. The value is checked to ensure it does not exceed
+     *   {@link Integer#MAX_VALUE} before being safely converted to {@code int}.
+     * </p>
+     *
+     * <p>
+     *   All behavior, validation rules, and exceptions are otherwise identical to
+     *   {@link #toByteArray(InputStream, int, int)}.
+     * </p>
+     *
+     * @param input the {@link InputStream} to read; must not be {@code null}.
+     * @param size  the exact number of bytes to read; must be {@code >= 0} and {@code <= Integer.MAX_VALUE}.
+     * @return a new byte array of length {@code size}.
+     * @throws IllegalArgumentException if {@code size} is negative or greater than {@link Integer#MAX_VALUE}.
+     * @throws EOFException             if the stream ends before {@code size} bytes are read.
+     * @throws IOException              if an I/O error occurs while reading.
+     * @throws NullPointerException     if {@code input} is {@code null}.
+     * @see #toByteArray(InputStream, int, int)
      * @since 2.1
      */
     public static byte[] toByteArray(final InputStream input, final long size) throws IOException {
@@ -2697,6 +2721,62 @@ public class IOUtils {
             throw new IllegalArgumentException("Size cannot be greater than Integer max value: " + size);
         }
         return toByteArray(input, (int) size);
+    }
+
+    /**
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
+     *
+     * <p>
+     *   This variant validates that the stream actually contains {@code size} bytes.
+     *   It is suitable for untrusted input because it prevents oversized allocations when the provided {@code size}
+     *   is corrupted or malicious.
+     * </p>
+     *
+     * <ul>
+     *   <li>If {@code size <= bufferSize}, the array is allocated directly and filled in a single pass.</li>
+     *   <li>
+     *     If {@code size > bufferSize}, the stream is read incrementally using a buffer of length {@code bufferSize}.
+     *     This avoids allocating an excessively large array up front,
+     *     but may temporarily double memory usage due to buffering.
+     *   </li>
+     * </ul>
+     *
+     * @param input      the {@link InputStream} to read; must not be {@code null}.
+     * @param size       the exact number of bytes to read; must be {@code >= 0}.
+     *                   The actual bytes read are validated to equal {@code size}.
+     * @param bufferSize the buffer size for incremental reading; must be {@code > 0}.
+     * @return a new byte array of length {@code size}.
+     * @throws IllegalArgumentException if {@code size} is negative or {@code bufferSize <= 0}.
+     * @throws EOFException             if the stream ends before {@code size} bytes are read.
+     * @throws IOException              if an I/O error occurs while reading.
+     * @throws NullPointerException     if {@code input} is {@code null}.
+     * @since 2.21.0
+     */
+    public static byte[] toByteArray(final InputStream input, final int size, final int bufferSize) throws IOException {
+        Objects.requireNonNull(input, "input");
+        if (size < 0) {
+            throw new IllegalArgumentException("Size must be equal or greater than zero: " + size);
+        }
+        if (bufferSize <= 0) {
+            throw new IllegalArgumentException("Chunk size must be greater than zero: " + bufferSize);
+        }
+        if (size <= bufferSize) {
+            return toByteArray(input::read, size);
+        }
+        try (UnsynchronizedByteArrayOutputStream output = UnsynchronizedByteArrayOutputStream.builder()
+                        .setBufferSize(bufferSize)
+                        .get();
+                InputStream boundedInput = BoundedInputStream.builder()
+                        .setMaxCount(size)
+                        .setPropagateClose(false)
+                        .setInputStream(input)
+                        .get()) {
+            output.write(boundedInput);
+            if (output.size() != size) {
+                throw new EOFException("Unexpected read size, current: " + output.size() + ", expected: " + size);
+            }
+            return output.toByteArray();
+        }
     }
 
     /**
@@ -2709,11 +2789,6 @@ public class IOUtils {
      * @throws IllegalArgumentException if {@code size} is less than zero.
      */
     static byte[] toByteArray(final IOTriFunction<byte[], Integer, Integer, Integer> input, final int size) throws IOException {
-
-        if (size < 0) {
-            throw new IllegalArgumentException("Size must be equal or greater than zero: " + size);
-        }
-
         if (size == 0) {
             return EMPTY_BYTE_ARRAY;
         }
--- a/src/main/java/org/apache/commons/io/RandomAccessFiles.java
+++ b/src/main/java/org/apache/commons/io/RandomAccessFiles.java
@@ -76,7 +76,11 @@ public class RandomAccessFiles {
      *                     other I/O error occurs.
      */
     public static byte[] read(final RandomAccessFile input, final long position, final int length) throws IOException {
+        Objects.requireNonNull(input, "input");
         input.seek(position);
+        if (length < 0) {
+            throw new IllegalArgumentException("Size must be equal or greater than zero: " + length);
+        }
         return IOUtils.toByteArray(input::read, length);
     }
 
--- a/src/test/java/org/apache/commons/io/IOUtilsTest.java
+++ b/src/test/java/org/apache/commons/io/IOUtilsTest.java
@@ -90,6 +90,9 @@ import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 /**
  * This is used to test {@link IOUtils} for correctness. The following checks are performed:
@@ -1657,6 +1660,48 @@ class IOUtilsTest {
             assertNotNull(out, "Out cannot be null");
             assertEquals(0, out.length, "Out length must be 0");
         }
+    }
+
+    @ParameterizedTest
+    @MethodSource
+    void testToByteArray_InputStream_Size_BufferSize_Succeeds(byte[] data, int size, int bufferSize) throws IOException {
+        final ByteArrayInputStream input = new ByteArrayInputStream(data);
+        final byte[] expected = Arrays.copyOf(data, size);
+        final byte[] actual = IOUtils.toByteArray(input, size, bufferSize);
+        assertArrayEquals(expected, actual);
+    }
+
+    private static Stream<Arguments> testToByteArray_InputStream_Size_BufferSize_Succeeds() {
+        final byte[] data = new byte[1024];
+        for (int i = 0; i < 1024; i++) {
+            data[i] = (byte) i;
+        }
+        return Stream.of(
+                // Eager reading
+                Arguments.of(data.clone(), 512, 1024),
+                // Incremental reading
+                Arguments.of(data.clone(), 1024, 512),
+                // No reading
+                Arguments.of(data.clone(), 0, 128));
+    }
+
+    @ParameterizedTest
+    @MethodSource
+    void testToByteArray_InputStream_Size_BufferSize_Throws(
+            int size, int bufferSize, Class<? extends Exception> exceptionClass) throws IOException {
+        try (InputStream input = new NullInputStream(0)) {
+            assertThrows(exceptionClass, () -> IOUtils.toByteArray(input, size, bufferSize));
+        }
+    }
+
+    static Stream<Arguments> testToByteArray_InputStream_Size_BufferSize_Throws() {
+        return Stream.of(
+                // Negative size
+                Arguments.of(-1, 128, IllegalArgumentException.class),
+                // Invalid buffer size
+                Arguments.of(0, 0, IllegalArgumentException.class),
+                // Huge size: should not cause OutOfMemoryError
+                Arguments.of(Integer.MAX_VALUE, 128, EOFException.class));
     }
 
     @Test


src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2707,9 +2790,6 @@ public static byte[] toByteArray(final InputStream input, final long size) throw
      * @throws IllegalArgumentException if {@code size} is less than zero.
      */
     static byte[] toByteArray(final IOTriFunction<byte[], Integer, Integer, Integer> input, final int size) throws IOException {
-        if (size < 0) {
```
garydgregory: This check should remain here; a negative array size never makes sense here.

src/main/java/org/apache/commons/io/RandomAccessFiles.java
```diff
@@ -76,7 +76,11 @@ private static long length(final RandomAccessFile raf) throws IOException {
      *                     other I/O error occurs.
      */
     public static byte[] read(final RandomAccessFile input, final long position, final int length) throws IOException {
+        Objects.requireNonNull(input, "input");
         input.seek(position);
+        if (length < 0) {
```
garydgregory: This change is unrelated to the feature, and I don't understand the benefit of moving the check from the method actually using the value to here.

src/changes/changes.xml
```diff
@@ -57,6 +57,7 @@ The <action> type attribute can be add,update,fix,remove.
       <action dev="ggregory" type="add"                due-to="Gary Gregory">Add org.apache.commons.io.output.ProxyOutputStream.writeRepeat(byte[], int, int, long).</action>
       <action dev="ggregory" type="add"                due-to="Gary Gregory">Add org.apache.commons.io.output.ProxyOutputStream.writeRepeat(byte[], long).</action>
       <action dev="ggregory" type="add"                due-to="Gary Gregory">Add org.apache.commons.io.output.ProxyOutputStream.writeRepeat(int, long).</action>
+      <action dev="pkarwasz" type="add"                due-to="Piotr P. Karwasz">Added toByteArray(InputStream, int, int) for safer incremental reading with size validation.</action>
```
garydgregory: "Added" -> "Add", like all other descriptions that add something.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2659,35 +2660,61 @@ public static byte[] toByteArray(final InputStream inputStream) throws IOExcepti
...
+     * @throws EOFException             if the stream ends before {@code size} bytes are read.
+     * @throws IOException              if an I/O error occurs while reading.
+     * @throws NullPointerException     if {@code input} is {@code null}.
      * @since 2.1
      */
     @SuppressWarnings("resource")
     public static byte[] toByteArray(final InputStream input, final int size) throws IOException {
-        return toByteArray(Objects.requireNonNull(input, "input")::read, size);
+        Objects.requireNonNull(input, "input");
+        if (size < 0) {
```
garydgregory: This check should remain in the underlying method.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2659,35 +2660,61 @@ public static byte[] toByteArray(final InputStream inputStream) throws IOExcepti
...
-     * {@link #toByteArray(InputStream)} when {@link InputStream} size is known.
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
      *
-     * @param input the {@link InputStream} to read.
-     * @param size the size of {@link InputStream} to read, where 0 &lt; {@code size} &lt;= length of input stream.
-     * @return byte [] of length {@code size}.
-     * @throws IOException if an I/O error occurs or {@link InputStream} length is smaller than parameter {@code size}.
-     * @throws IllegalArgumentException if {@code size} is less than zero.
+     * <p>
+     *   This variant allocates the target array immediately and attempts to fill it in one pass.
```
garydgregory: Saying "one pass" feels off since the underlying worker method _loops_ until EOF or the right amount is read. I don't think we should describe the _how_ of this method.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2659,35 +2660,61 @@ public static byte[] toByteArray(final InputStream inputStream) throws IOExcepti
...
+     *   It assumes that {@code size} is correct.
+     *   If the stream ends prematurely, an {@link EOFException} is thrown.
+     * </p>
+     *
+     * <p>
```
garydgregory: This feels like FUD to me, and I would remove the whole paragraph. I don't think we want to start documenting which methods fall into which defensive category, especially when the actual issue would be in the call site, not this implementation.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2659,35 +2660,61 @@ public static byte[] toByteArray(final InputStream inputStream) throws IOExcepti
...
-     * @param size the size of {@link InputStream} to read, where 0 &lt; {@code size} &lt;= min(Integer.MAX_VALUE, length of input stream).
-     * @return byte [] the requested byte array, of length {@code size}.
-     * @throws IOException              if an I/O error occurs or {@link InputStream} length is less than {@code size}.
-     * @throws IllegalArgumentException if size is less than zero or size is greater than Integer.MAX_VALUE.
-     * @see IOUtils#toByteArray(InputStream, int)
+     * <p>
+     *   This is a convenience overload of {@link #toByteArray(InputStream, int, int)} that accepts a
+     *   {@code long} size parameter. The value is checked to ensure it does not exceed
+     *   {@link Integer#MAX_VALUE} before being safely converted to {@code int}.
+     * </p>
```
garydgregory: This whole paragraph (above) is wrong.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2659,35 +2660,61 @@ public static byte[] toByteArray(final InputStream inputStream) throws IOExcepti
...
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
      *
-     * @param input the {@link InputStream} to read.
-     * @param size the size of {@link InputStream} to read, where 0 &lt; {@code size} &lt;= length of input stream.
-     * @return byte [] of length {@code size}.
-     * @throws IOException if an I/O error occurs or {@link InputStream} length is smaller than parameter {@code size}.
-     * @throws IllegalArgumentException if {@code size} is less than zero.
+     * <p>
+     *   This variant allocates the target array immediately and attempts to fill it in one pass.
+     *   It assumes that {@code size} is correct.
```
garydgregory: Let's not say we assume a parameter is correct; after all, we don't say when we assume a parameter incorrect in other cases. That's what exceptions are for, and sometimes the return value.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2659,35 +2660,61 @@ public static byte[] toByteArray(final InputStream inputStream) throws IOExcepti
...
+     *   If the stream ends prematurely, an {@link EOFException} is thrown.
```
garydgregory: You can remove this since it duplicates the exception documentation.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2659,35 +2660,61 @@ public static byte[] toByteArray(final InputStream inputStream) throws IOExcepti
...
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
      *
-     * @param input the {@link InputStream} to read.
-     * @param size the size of {@link InputStream} to read, where 0 &lt; {@code size} &lt;= min(Integer.MAX_VALUE, length of input stream).
-     * @return byte [] the requested byte array, of length {@code size}.
-     * @throws IOException              if an I/O error occurs or {@link InputStream} length is less than {@code size}.
-     * @throws IllegalArgumentException if size is less than zero or size is greater than Integer.MAX_VALUE.
-     * @see IOUtils#toByteArray(InputStream, int)
+     * <p>
+     *   This is a convenience overload of {@link #toByteArray(InputStream, int, int)} that accepts a
```
garydgregory: Let's not define convenience versus not; the convenience is only due to the implementation, which feels wrong, we should probably (and separately) fix the implementation to use `Math.toExactInt()`.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2697,6 +2724,62 @@ public static byte[] toByteArray(final InputStream input, final long size) throw
         return toByteArray(input, (int) size);
     }
 
+    /**
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
+     *
+     * <p>
+     *   This variant validates that the stream actually contains {@code size} bytes.
+     *   It is suitable for untrusted input because it prevents oversized allocations when the provided {@code size}
```
garydgregory: I don't think we should start to document which APIs are designed for trusted versus untrusted input. The documentation should just describe the intended functionality.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2697,6 +2724,62 @@ public static byte[] toByteArray(final InputStream input, final long size) throw
...
+     * @throws NullPointerException     if {@code input} is {@code null}.
+     * @since 2.21.0
+     */
+    public static byte[] toByteArray(final InputStream input, final int size, final int bufferSize) throws IOException {
+        Objects.requireNonNull(input, "input");
+        if (size < 0) {
+            throw new IllegalArgumentException("Size must be equal or greater than zero: " + size);
+        }
+        if (bufferSize <= 0) {
+            throw new IllegalArgumentException("Chunk size must be greater than zero: " + bufferSize);
```
garydgregory: The message text and parameter names should match: Either "buffer" or "chunk", not both.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2697,6 +2724,62 @@ public static byte[] toByteArray(final InputStream input, final long size) throw
...
+     * @return a new byte array of length {@code size}.
+     * @throws IllegalArgumentException if {@code size} is negative or {@code bufferSize <= 0}.
+     * @throws EOFException             if the stream ends before {@code size} bytes are read.
+     * @throws IOException              if an I/O error occurs while reading.
+     * @throws NullPointerException     if {@code input} is {@code null}.
+     * @since 2.21.0
+     */
+    public static byte[] toByteArray(final InputStream input, final int size, final int bufferSize) throws IOException {
+        Objects.requireNonNull(input, "input");
+        if (size < 0) {
```
garydgregory: You can remove this check since the worker method does it (in master).

garydgregory (changes requested):
Hi @ppkarwasz 

I have lots of comments! 😉

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: This method is package-private, so it’s reasonable to push the responsibility for validating array size onto the callers. At one point most callers already performed this check, so I added it to the one missing call site and removed it here.

After reviewing it again, I agree the check belongs here, so I’ll restore it and remove the duplicates in the callers. Fixed in https://github.com/apache/commons-io/pull/776/commits/d748e99d390eef117d54bf175dcaa569e2b9106d

src/main/java/org/apache/commons/io/RandomAccessFiles.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/d748e99d390eef117d54bf175dcaa569e2b9106d

src/changes/changes.xml
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/33a4fe15a7c05a3b850b6a9872ac6dd3b7c0aaf6

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/d748e99d390eef117d54bf175dcaa569e2b9106d

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/7362d3e34801662b029ea6ef4e005dfb1194c57b

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: I removed details about which methods can be used for defensive programming in https://github.com/apache/commons-io/pull/776/commits/7362d3e34801662b029ea6ef4e005dfb1194c57b

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/7362d3e34801662b029ea6ef4e005dfb1194c57b

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Since this method throws an unchecked exception if the `long` does not fit into an `int`, most callers have to check the parameter, before calling this. So this methods looks like a convenient composition of a size check and `toByteArray(InputStream, int, int)`.

Fixed in https://github.com/apache/commons-io/pull/776/commits/7362d3e34801662b029ea6ef4e005dfb1194c57b

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/7362d3e34801662b029ea6ef4e005dfb1194c57b

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/7362d3e34801662b029ea6ef4e005dfb1194c57b

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Removed in https://github.com/apache/commons-io/pull/776/commits/7362d3e34801662b029ea6ef4e005dfb1194c57b

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/fe39b777c447f2eeb185ae2520ca6b547d84e719

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/c6c79a3dc4fcfcc84a82189f9ae41e3a051a5d35

ppkarwasz:
@garydgregory,

I removed a lot of details from the Javadoc and let only these:

- For `toByteArray(InputStream)`:
  https://github.com/apache/commons-io/blob/c6c79a3dc4fcfcc84a82189f9ae41e3a051a5d35/src/main/java/org/apache/commons/io/IOUtils.java#L2643-L2644
- For `toByteArray(InputStream, int/long)`:
  https://github.com/apache/commons-io/blob/c6c79a3dc4fcfcc84a82189f9ae41e3a051a5d35/src/main/java/org/apache/commons/io/IOUtils.java#L2666-L2667
- For `toByteArray(InputStream, int, int)`:
  https://github.com/apache/commons-io/blob/c6c79a3dc4fcfcc84a82189f9ae41e3a051a5d35/src/main/java/org/apache/commons/io/IOUtils.java#L2708-L2709

I believe that these behaviors will not change and should remain in the method contracts, what do you think?

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2638,61 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
+     * containing the entire contents once the end of the stream is reached.</p>
+     *
+     * @param input the {@link InputStream} to read; must not be {@code null}.
+     * @return a new byte array.
+     * @throws IllegalArgumentException if the size of the stream is greater than {@code Integer.MAX_VALUE}.
+     * @throws IOException          if an I/O error occurs while reading.
+     * @throws NullPointerException if {@code input} is {@code null}.
      */
-    public static byte[] toByteArray(final InputStream inputStream) throws IOException {
+    public static byte[] toByteArray(final InputStream input) throws IOException {
```
garydgregory: Please leave the parameter name unchanged. We can normalize the other names later to "inputStream" for consistency.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2638,61 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
-     * <p>
-     * This method buffers the input internally, so there is no need to use a {@link BufferedInputStream}.
-     * </p>
+     * Reads all remaining bytes from the given {@link InputStream} into a new {@code byte[]}.
      *
-     * @param inputStream the {@link InputStream} to read.
-     * @return the requested byte array.
-     * @throws NullPointerException if the InputStream is {@code null}.
-     * @throws IOException          if an I/O error occurs or reading more than {@link Integer#MAX_VALUE} occurs.
+     * <p>The method accumulates the data in temporary buffers and returns a single array
```
garydgregory: This comment is not needed IMO, the user cannot affect the behavior about it by passing in different values. It's a low-level implementation detail.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2638,61 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
     }
 
     /**
-     * Gets the contents of an {@link InputStream} as a {@code byte[]}.
-     * <p>
-     * This method buffers the input internally, so there is no need to use a {@link BufferedInputStream}.
-     * </p>
+     * Reads all remaining bytes from the given {@link InputStream} into a new {@code byte[]}.
```
garydgregory: What follows is just about the finer aspects of documenting methods, but it's still nice to talk about IMO:

The text can be simpler and more direct, for example, `java.nio.file.Files.read(InputStream, int)` says "Reads all the bytes from an input stream." After all, the only kinds of bytes in an input stream are the "remaining" kind, and it only makes sense to return a "new" array. So I would say: "Reads all the bytes from an input stream in a byte array".

As an aside, Java seems to have called out the "remaining" concept formally in one API name: "Iterator.forEachRemaining(...)", but that might have been to avoid overriding or shadowing the "forEach(...)".

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2638,61 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
-     * Gets the contents of an {@link InputStream} as a {@code byte[]}. Use this method instead of
-     * {@link #toByteArray(InputStream)} when {@link InputStream} size is known.
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
      *
-     * @param input the {@link InputStream} to read.
-     * @param size the size of {@link InputStream} to read, where 0 &lt; {@code size} &lt;= length of input stream.
-     * @return byte [] of length {@code size}.
-     * @throws IOException if an I/O error occurs or {@link InputStream} length is smaller than parameter {@code size}.
-     * @throws IllegalArgumentException if {@code size} is less than zero.
+     * <p>The method allocates a single array of the requested size and fills it directly
```
garydgregory: This comment is not needed IMO, the user cannot affect the behavior by passing in different values. It's a low-level implementation detail.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2638,61 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
-     * (Arrays can have no more than Integer.MAX_VALUE entries anyway.)
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
      *
-     * @param input the {@link InputStream} to read.
-     * @param size the size of {@link InputStream} to read, where 0 &lt; {@code size} &lt;= min(Integer.MAX_VALUE, length of input stream).
-     * @return byte [] the requested byte array, of length {@code size}.
-     * @throws IOException              if an I/O error occurs or {@link InputStream} length is less than {@code size}.
-     * @throws IllegalArgumentException if size is less than zero or size is greater than Integer.MAX_VALUE.
-     * @see IOUtils#toByteArray(InputStream, int)
+     * <p>The method allocates a single array of the requested size and fills it directly
```
garydgregory: This comment is not needed IMO, the user cannot affect the behavior by passing in different values. It's a low-level implementation detail.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2697,6 +2702,48 @@ public static byte[] toByteArray(final InputStream input, final long size) throw
         return toByteArray(input, (int) size);
     }
 
+    /**
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
+     *
+     * <p>The method accumulates the data in temporary buffers of size at most {@code bufferSize}
```
garydgregory: I think it would be nice to explain _why_ this method is useful, instead of _how_ it works. For example, "When reading from an untrusted input stream, you reduce the risk of an OOME by calling this API with...".

garydgregory (changes requested):
Hi @ppkarwasz ,
I have comments scattered about.
TY!

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/9439095923e901ef3e4eac9ae4f06c00268a280c

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Removed in https://github.com/apache/commons-io/pull/776/commits/97d37a94de8c935ea8e4961e354a1cd1d5c49b6a

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Removed in https://github.com/apache/commons-io/pull/776/commits/97d37a94de8c935ea8e4961e354a1cd1d5c49b6a

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Removed in https://github.com/apache/commons-io/pull/776/commits/97d37a94de8c935ea8e4961e354a1cd1d5c49b6a

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Replaced with guidance about OOME in https://github.com/apache/commons-io/pull/776/commits/97d37a94de8c935ea8e4961e354a1cd1d5c49b6a

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/cbfa307f526f8c6fac4744fea135de16edbdb1b0

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2697,6 +2693,52 @@ public static byte[] toByteArray(final InputStream input, final long size) throw
...
 
+    /**
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
+     *
+     * <p>When reading from an untrusted stream, this variant lowers the risk of
+     * {@link OutOfMemoryError} by allocating data in buffers of up to {@code bufferSize}
+     * bytes rather than in one large array.</p>
+     *
+     * <p>Note, however, that this approach requires additional temporary memory
+     * compared to {@link #toByteArray(InputStream, int)}.</p>
```
ppkarwasz: If we document `OutOfMemoryError` risk in the Javadoc…

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2697,6 +2693,52 @@ public static byte[] toByteArray(final InputStream input, final long size) throw
...
+     */
+    public static byte[] toByteArray(final InputStream input, final int size, final int bufferSize) throws IOException {
+        Objects.requireNonNull(input, "input");
+        if (bufferSize <= 0) {
+            throw new IllegalArgumentException("Buffer size must be greater than zero: " + bufferSize);
+        }
+        if (size <= bufferSize) {
+            // throws if size < 0
+            return toByteArray(input::read, size);
+        }
```
ppkarwasz: …should we also extend the "fast path" case to situations where the requested size is already available in the stream?

```suggestion
        // Fast path: use direct allocation if the size fits in one buffer
        // or if at least `size` bytes are already available without blocking.
        if (size <= Math.max(bufferSize, input.available())) {
            // throws if size < 0
            return toByteArray(input::read, size);
        }
```

In this case what would you suggest I add to the Javadoc?

ppkarwasz:
Hi @garydgregory,

I’ve refactored this proposal a bit further:

* Extended chunked reading to the legacy `toByteArray(InputStream, int/long)` methods as well.
* Revised the Javadoc to clarify the contract. As you mentioned earlier, we should **not** guide users based on how *trusted* the size parameter is. I’ve also removed explicit references to `OutOfMemoryError`, which can always occur. Instead, the docs now emphasize that memory allocation is **proportional to the number of bytes actually read** (previously it was _de facto_ proportional to the `size` argument).

Open questions:

* **Chunking threshold**: Currently set to 128 KiB, matching CPython. JDK 11+ uses 16 KiB. We could also consider raising our default buffer size (currently 8 KiB, and in some places as low as 1 KiB, which feels outdated).
* **Use of `available()`**: Should we consult `available()` in `toByteArray(InputStream, int)` and `toByteArray(InputStream)` (see https://github.com/apache/commons-io/pull/776#discussion_r2326598113)? The `toByteArray(InputStream, int, int)` method explicitly promises chunks up to `chunkSize`, but the others do not.

garydgregory:
> Hi @garydgregory,
> 
> I’ve refactored this proposal a bit further:
> 
> * Extended chunked reading to the legacy `toByteArray(InputStream, int/long)` methods as well.
> * Revised the Javadoc to clarify the contract. As you mentioned earlier, we should **not** guide users based on how _trusted_ the size parameter is. I’ve also removed explicit references to `OutOfMemoryError`, which can always occur. Instead, the docs now emphasize that memory allocation is **proportional to the number of bytes actually read** (previously it was _de facto_ proportional to the `size` argument).
> 
> Open questions:
> 
> * **Chunking threshold**: Currently set to 128 KiB, matching CPython. JDK 11+ uses 16 KiB. We could also consider raising our default buffer size (currently 8 KiB, and in some places as low as 1 KiB, which feels outdated).
> * **Use of `available()`**: Should we consult `available()` in `toByteArray(InputStream, int)` and `toByteArray(InputStream)` (see [feat: Add incremental `toByteArray` method #776 (comment)](https://github.com/apache/commons-io/pull/776#discussion_r2326598113))? The `toByteArray(InputStream, int, int)` method explicitly promises chunks up to `chunkSize`, but the others do not.

CPython is irrelevant IMO. Java's `Files.BUFFER_SIZE` class in Java 21 is 8K, so 8K is consistent as the default.

> JDK 11+ uses 16 KiB.

Where do you see that?

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -221,6 +221,21 @@ public class IOUtils {
      */
     private static final char[] SCRATCH_CHAR_BUFFER_WO = charArray();
 
+    /**
+     * The maximum size of an array in many Java VMs.
+     */
+    private static final int MAX_ARRAY_LENGTH = Integer.MAX_VALUE - 8;
```
garydgregory: Java 12's `ArraySupport` class calls this `SOFT_MAX_ARRAY_LENGTH`, so I think we should use the same name.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -221,6 +221,21 @@ public class IOUtils {
...
+
+    /*
+     * Default maximum chunk size used when copying large streams into a byte array.
+     * <p>
+     * This value is somewhat arbitrary, currently aligned with the value used by
+     * <a href="https://github.com/python/cpython/blob/3.14/Lib/_pyio.py">Python</a>
```
garydgregory: CPython is completely irrelevant here; it should not be mentioned, as we certainly do not want to give the impression that we are aligning anything with CPython. I also like to mention the value with `{@value}` in this kind of Javadoc.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2697,6 +2720,69 @@ public static byte[] toByteArray(final InputStream input, final long size) throw
...
+        final UnsynchronizedByteArrayOutputStream output = copyToOutputStream(input, size, chunkSize);
+         if (output.size() != size) {
+            throw new EOFException("Unexpected read size, current: " + output.size() + ", expected: " + size);
+        }
+        return output.toByteArray();
+    }
+
+    /**
+     * Copies up to {@code size} bytes from the given {@link InputStream} into a new {@link UnsynchronizedByteArrayOutputStream}.
+     *
```
garydgregory, line 2754: Could you remove extra line?

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2652,65 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
     public static byte[] toByteArray(final InputStream inputStream) throws IOException {
-        // We use a ThresholdingOutputStream to avoid reading AND writing more than Integer.MAX_VALUE.
-        try (UnsynchronizedByteArrayOutputStream ubaOutput = UnsynchronizedByteArrayOutputStream.builder().get();
-                ThresholdingOutputStream thresholdOutput = new ThresholdingOutputStream(Integer.MAX_VALUE, os -> {
-                    throw new IllegalArgumentException(String.format("Cannot read more than %,d into a byte array", Integer.MAX_VALUE));
-                }, os -> ubaOutput)) {
-            copy(inputStream, thresholdOutput);
-            return ubaOutput.toByteArray();
+        final UnsynchronizedByteArrayOutputStream output =
+                copyToOutputStream(inputStream, MAX_ARRAY_LENGTH + 1, DEFAULT_CHUNK_SIZE);
```
garydgregory: You can use line lengths up to 160 per our Checkstyle, so the code doesn't end up reading like a newspaper 😉

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2652,65 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
+     * @throws IllegalArgumentException if {@code size} is negative.
+     * @throws EOFException             if the stream ends before {@code size} bytes are read.
+     * @throws IOException              if an I/O error occurs while reading.
+     * @throws NullPointerException     if {@code input} is {@code null}.
      * @since 2.1
      */
-    @SuppressWarnings("resource")
     public static byte[] toByteArray(final InputStream input, final int size) throws IOException {
-        return toByteArray(Objects.requireNonNull(input, "input")::read, size);
+        return toByteArray(input, size, DEFAULT_CHUNK_SIZE);
```
garydgregory: Woa, this is surprising; now we are double-buffering and slowing down existing call sites that don't want this new feature? It seems to like we should _only_ do the double buffering when a call site chooses to use the new API.

ppkarwasz:
> > JDK 11+ uses 16 KiB.
> 
> Where do you see that?

In the [source code of `InputStream`](https://github.com/openjdk/jdk/blob/bea2b029a77e126171d17c3a44baec6d5cafed4a/src/java.base/share/classes/java/io/InputStream.java#L58). This is the value used by `readAllBytes/readNBytes`.

src/main/java/org/apache/commons/io/IOUtils.java
vy: A reference to `ArraysSupport#SOFT_MAX_ARRAY_LENGTH` in the Javadoc would be handy for maintainers.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -221,6 +221,21 @@ public class IOUtils {
...
+
+    /*
+     * Default maximum chunk size used when copying large streams into a byte array.
+     * <p>
+     * This value is somewhat arbitrary, currently aligned with the value used by
+     * <a href="https://github.com/python/cpython/blob/3.14/Lib/_pyio.py">Python</a>
+     * for copying streams.
+     * </p>
+     */
+    private static final int DEFAULT_CHUNK_SIZE = 128 * 1024;
```
vy: Food for thought: `Files.BUFFER_SIZE` is 8 KiB – maybe we shall stick to Java defaults? Note that 8 KiB will theoretically allow more threads consuming in parallel.

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: I aligned this method with the JDK methods [`InputStream.readNBytes()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/InputStream.html#readNBytes(int)), which performs a similar task, but is only available since JDK 11.

Our Javadoc in this method does not prevent us from making the switch:

> Gets the contents of an `InputStream` as a byte[]. Use this method instead of `toByteArray(InputStream)` when `InputStream` size is known.

I think that with the right value for the threshold, the change might be profitable for users, what about using `Math.max(in.available(), DEFAULT_CHUNK_SIZE)` as threshold? This should guarantee that loading from local sources will **not** double-buffer (`FileInputStream` reports the remaining bytes in the file as “available”). What do you think?

Personally I would rather not rely on the new behavior in my projects, since there is not guarantee users don't downgrade Commons IO to `2.20.0`, so I am mostly interested in `toByteArray(InputStream, int, int)`, but we might consider improving this method too.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2652,65 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
-        // We use a ThresholdingOutputStream to avoid reading AND writing more than Integer.MAX_VALUE.
-        try (UnsynchronizedByteArrayOutputStream ubaOutput = UnsynchronizedByteArrayOutputStream.builder().get();
-                ThresholdingOutputStream thresholdOutput = new ThresholdingOutputStream(Integer.MAX_VALUE, os -> {
-                    throw new IllegalArgumentException(String.format("Cannot read more than %,d into a byte array", Integer.MAX_VALUE));
-                }, os -> ubaOutput)) {
-            copy(inputStream, thresholdOutput);
-            return ubaOutput.toByteArray();
+        final UnsynchronizedByteArrayOutputStream output =
+                copyToOutputStream(inputStream, MAX_ARRAY_LENGTH + 1, DEFAULT_CHUNK_SIZE);
+        if (output.size() > MAX_ARRAY_LENGTH) {
```
vy: Can this ever happen? If not, I'd replace `if` with `assert`.

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/38a6a2c6b7dcb3cf0646a54ce4a762913f6b7627

src/main/java/org/apache/commons/io/IOUtils.java
vy: I don't think this necessarily implies double buffering: we can avoid the ephemeral buffer if `size <= chunkSize` – and this is what @ppkarwasz already does:

https://github.com/apache/commons-io/pull/776/files#diff-b106f51e7daf5b40ed10dccc4622f425f3e82afc52fb736547a8f28efa6478e4R2750-R2753

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/38a6a2c6b7dcb3cf0646a54ce4a762913f6b7627

I didn't use `{@value}`, since as far as I remember, the `javadoc` tool in JDK 8 fails to reference values, which are not public constants.

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/29f365be5d485dce7d03f0d0384dbf36fd40cc9d

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/29f365be5d485dce7d03f0d0384dbf36fd40cc9d

This was folded by my formatter and I usually don't change anything, unless there is a CheckStyle rule that fails.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2652,65 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
-     * Gets the contents of an {@link InputStream} as a {@code byte[]}. Use this method instead of
-     * {@link #toByteArray(InputStream)} when {@link InputStream} size is known.
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
      *
-     * @param input the {@link InputStream} to read.
-     * @param size the size of {@link InputStream} to read, where 0 &lt; {@code size} &lt;= length of input stream.
-     * @return byte [] of length {@code size}.
-     * @throws IOException if an I/O error occurs or {@link InputStream} length is smaller than parameter {@code size}.
-     * @throws IllegalArgumentException if {@code size} is less than zero.
+     * <p>The memory used by this method is <strong>proportional</strong> to the number
```
vy: Can you explain what is meant with `proportional` here? As a user reading this to understand the API, I find it more confusing than helpful. Can't we simply say something as follows?

> This method allocates at most `size` bytes, plus {@value MAX_ARRAY_LENGTH} bytes for the temporary buffer used to consume the stream.

vy: Okay... I see we're using `UnsynchronizedByteArrayOutputStream`, which allocates along the way. Now I understand why you avoided to provide an elaborate explanation on the memory cap. I was about to suggest you to shape the Javadoc similar to `InputStream::readNBytes`, but then I realized that you already did by using the `proportional` term here. I see you've done your homework. :+1:

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2714,13 +2800,9 @@ static byte[] toByteArray(final IOTriFunction<byte[], Integer, Integer, Integer>
...
         }
         final byte[] data = byteArray(size);
-        int offset = 0;
-        int read;
-        while (offset < size && (read = input.apply(data, offset, size - offset)) != EOF) {
-            offset += read;
-        }
-        if (offset != size) {
-            throw new IOException("Unexpected read size, current: " + offset + ", expected: " + size);
+        final int read = read(input, data, 0, size);
```
vy: AFAICT, this _improvement_ feels unrelated with this PR. You might consider carrying this out in a separate PR.

src/test/java/org/apache/commons/io/IOUtilsTest.java
vy, line 1: I don't see a test that verifies the exception on insufficient data. Am I mistaken?

src/main/java/org/apache/commons/io/IOUtils.java
garydgregory: I think this method's behavior should remain unchanged. The current goal is to use the new method in Commons  Compress, so let's focus on that IMO.

ppkarwasz: I restored the previous behavior in https://github.com/apache/commons-io/pull/776/commits/3afae073c62d55f241f3893612e85ce650b79bbe, but I added a recommendation to use `toByteArray(InputStream, int, int)` for large arrays:

https://github.com/apache/commons-io/blob/4aba097d871332ca1ac537abb5f6880c0bf0ffc3/src/main/java/org/apache/commons/io/IOUtils.java#L2691-L2693

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: I switched to 8 KiB in https://github.com/apache/commons-io/pull/776/commits/4aba097d871332ca1ac537abb5f6880c0bf0ffc3

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Yes, it's possible since `MAX_ARRAY_LENGTH` is `Integer.MAX_VALUE - 8`.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2645,61 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
+     * @throws IllegalArgumentException if {@code size} is negative.
+     * @throws EOFException             if the stream ends before {@code size} bytes are read.
+     * @throws IOException              if an I/O error occurs while reading.
+     * @throws NullPointerException     if {@code input} is {@code null}.
      * @since 2.1
      */
-    @SuppressWarnings("resource")
     public static byte[] toByteArray(final InputStream input, final int size) throws IOException {
-        return toByteArray(Objects.requireNonNull(input, "input")::read, size);
+        return toByteArray(input::read, size);
```
garydgregory: With this change, we are losing the better NPE message on Java < 14.

src/changes/changes.xml
```diff
@@ -57,6 +57,7 @@ The <action> type attribute can be add,update,fix,remove.
       <action dev="ggregory" type="add"                due-to="Gary Gregory">Add org.apache.commons.io.output.ProxyOutputStream.writeRepeat(byte[], int, int, long).</action>
       <action dev="ggregory" type="add"                due-to="Gary Gregory">Add org.apache.commons.io.output.ProxyOutputStream.writeRepeat(byte[], long).</action>
       <action dev="ggregory" type="add"                due-to="Gary Gregory">Add org.apache.commons.io.output.ProxyOutputStream.writeRepeat(int, long).</action>
+      <action dev="pkarwasz" type="add"                due-to="Piotr P. Karwasz">Add IOUtils.toByteArray(InputStream, int, int) for safer incremental reading with size validation.</action>
```
ecki: Btw what’s incremental about reading a while stream into a single array?

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -221,6 +221,14 @@ public class IOUtils {
      */
     private static final char[] SCRATCH_CHAR_BUFFER_WO = charArray();
 
+    /**
+     * The maximum size of an array in many Java VMs.
+     * <p>
+     *     The constant is copied from OpenJDK's {@link jdk.internal.util.ArraysSupport#SOFT_MAX_ARRAY_LENGTH}.
+     * </p>
+     */
+    private static final int SOFT_MAX_ARRAY_LENGTH = Integer.MAX_VALUE - 8;
```
ecki, line 230: I wonder what’s soft and what’s -8 is about, but I guess the question is for openjdk (for the review it’s only important that it does not overflow, see below)

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2645,61 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
-     * <p>
-     * This method buffers the input internally, so there is no need to use a {@link BufferedInputStream}.
-     * </p>
+     * Reads all the bytes from an input stream in a byte array.
      *
-     * @param inputStream the {@link InputStream} to read.
-     * @return the requested byte array.
-     * @throws NullPointerException if the InputStream is {@code null}.
-     * @throws IOException          if an I/O error occurs or reading more than {@link Integer#MAX_VALUE} occurs.
+     * <p>The memory used by this method is <strong>proportional</strong> to the number
```
ecki, line 2650: For me “proportional” could also mean 2x so maybe “is the same”? And “large input streams” is also relatife with 2gb. Maybe say it directly “Only streams which fit into a single byte array with roughly 2GB limit can be processed with this method”?

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2645,61 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
+     * of bytes read, which is only limited by {@link Integer#MAX_VALUE}. This makes it unsuitable for
+     * processing large input streams, unless sufficient heap space is available.</p>
+     *
+     * @param inputStream The {@link InputStream} to read; must not be {@code null}.
+     * @return A new byte array.
+     * @throws IllegalArgumentException If the size of the stream is greater than the maximum array size.
```
ecki, line 2657: IAE is a bit strange but since it’s the existing APi it’s fine to keep that.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2645,61 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
      */
     public static byte[] toByteArray(final InputStream inputStream) throws IOException {
-        // We use a ThresholdingOutputStream to avoid reading AND writing more than Integer.MAX_VALUE.
-        try (UnsynchronizedByteArrayOutputStream ubaOutput = UnsynchronizedByteArrayOutputStream.builder().get();
-                ThresholdingOutputStream thresholdOutput = new ThresholdingOutputStream(Integer.MAX_VALUE, os -> {
-                    throw new IllegalArgumentException(String.format("Cannot read more than %,d into a byte array", Integer.MAX_VALUE));
-                }, os -> ubaOutput)) {
-            copy(inputStream, thresholdOutput);
-            return ubaOutput.toByteArray();
+        final UnsynchronizedByteArrayOutputStream output = copyToOutputStream(inputStream, SOFT_MAX_ARRAY_LENGTH + 1, DEFAULT_BUFFER_SIZE);
```
ecki, line 2663: Maybe add a comment that the soft max is guranteed to not overflow on +1?

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2645,61 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
+        if (output.size() > SOFT_MAX_ARRAY_LENGTH) {
```
ecki, line 2664: Hm, the UABOS.size() feels a bit overloaded here and why does the write() not throw?

src/main/java/org/apache/commons/io/IOUtils.java
ecki, line 2663: Overall it looks like a simplification!

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2637,57 +2645,61 @@ public static BufferedReader toBufferedReader(final Reader reader, final int siz
...
-     * (Arrays can have no more than Integer.MAX_VALUE entries anyway.)
+     * Reads exactly {@code size} bytes from the given {@link InputStream} into a new {@code byte[]}.
      *
-     * @param input the {@link InputStream} to read.
-     * @param size the size of {@link InputStream} to read, where 0 &lt; {@code size} &lt;= min(Integer.MAX_VALUE, length of input stream).
-     * @return byte [] the requested byte array, of length {@code size}.
-     * @throws IOException              if an I/O error occurs or {@link InputStream} length is less than {@code size}.
-     * @throws IllegalArgumentException if size is less than zero or size is greater than Integer.MAX_VALUE.
-     * @see IOUtils#toByteArray(InputStream, int)
+     * <p>This variant provides no safeguards against allocating very large arrays.
```
ecki: Why not? If size is small it allocates no large array?

src/test/java/org/apache/commons/io/IOUtilsTest.java
```diff
@@ -1659,6 +1662,48 @@ void testToByteArray_InputStream_SizeZero() throws Exception {
...
+    }
+
+    private static Stream<Arguments> testToByteArray_InputStream_Size_BufferSize_Succeeds() {
+        final byte[] data = new byte[1024];
+        for (int i = 0; i < 1024; i++) {
+            data[i] = (byte) i;
+        }
+        return Stream.of(
+                // Eager reading
+                Arguments.of(data.clone(), 512, 1024),
```
ecki, line 1681: Those clones and array copies are pretty defensive, but I guess it’s fine here the test only uses small ones

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Yes, I borrowed the **proportional** keyword, which seems to be the magical word to dismiss DoS reports.

Of course `<bytes read> + Integer.MAX_VALUE` is also proportional to the number of bytes read. :wink:

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Nice catch! :100:

Fixed in https://github.com/apache/commons-io/pull/776/commits/15b249da4f4e6bc8238dfce9ae34f4d21f7f88a4

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/624de9fa1a2d376e944219249c56639e02da441b

src/test/java/org/apache/commons/io/IOUtilsTest.java
ppkarwasz: This test case shows the exception on insufficient data:

https://github.com/apache/commons-io/blob/624de9fa1a2d376e944219249c56639e02da441b/src/test/java/org/apache/commons/io/IOUtilsTest.java#L1703-L1704

Surefire is configured with 25 MiB of heap (probably should be 256 MiB), so having a smaller chunk size prevents an OOME here.

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Yes, it is the current API.

@garydgregory, should we change it in a separate PR, so it throws an `IOException`?

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Yes, `size()` is quite a generic method name, however `UBAOS.size()` is in line with what `ByteArrayOutputStream.size()` from the JDK does.

Which `write` are you referring to? The one used in `copyToOutputStream`? I does throw `IOException` on error.

src/changes/changes.xml
ppkarwasz: Should I say “chunked”?

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: The total memory allocation for this method is actually about **twice** the real size of the stream, since data is first loaded into chunks and only then copied to a `byte` array of the appropriate size.

I fixed the mention of “large input streams” in https://github.com/apache/commons-io/pull/776/commits/eb6c2bccb61dda3b586e7f5828ae7cf009970a70

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/eb6c2bccb61dda3b586e7f5828ae7cf009970a70

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: The difference between this method and `toByteArray(InputStream, size, chunkSize)` is that this method always allocates a byte array of size `size`, while the other allocates `2 * min(size, remaining bytes in the stream)`.

These are the safeguards I am mentioning in the Javadoc: if the stream is corrupted this method could allocate 2 GiB of memory and cause an OOME, while the other method will allocate up to twice what is left in the stream. How could I express this better?

src/main/java/org/apache/commons/io/IOUtils.java
ecki: I mean on reaching the byte array limit. Would be a cleaner signalling compared to returning with size one bigger than a limit

src/changes/changes.xml
ecki: Ah yeah chunked or maybe dynamic, if the idea it works for partial results.

src/main/java/org/apache/commons/io/IOUtils.java
ecki: How about “This variant always allocates the whole requested array size, for a dynamic growing variant use ..”?

ppkarwasz: Thanks! Fixed in https://github.com/apache/commons-io/pull/776/commits/e6e2a4d6e1dafb74ef8162116183862fdd9f8d70

src/changes/changes.xml
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/776/commits/e6e2a4d6e1dafb74ef8162116183862fdd9f8d70

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: I believe that under the hood both the current code (below) and the proposed change reads 1 byte more than the limit.

https://github.com/apache/commons-io/blob/4dc17b6cb8c794cd15acb2bc99fe30d63306d5f9/src/main/java/org/apache/commons/io/IOUtils.java#L2651-L2657

I prefer the new approach because:

* It writes directly into `UnsynchronizedByteArrayOutputStream` buffers without requiring additional buffers in the `copy` call.
* the control flow is more explicity (`copy(limit + 1) -> size check -> throw`),
* and it avoids the overhead of extra wrappers and callbacks.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -2697,6 +2711,68 @@ public static byte[] toByteArray(final InputStream input, final long size) throw
...
+        if (chunkSize <= 0) {
+            throw new IllegalArgumentException("Chunk size must be greater than zero: " + chunkSize);
+        }
+        if (size <= chunkSize) {
+            // throws if size < 0
+            return toByteArray(input::read, size);
+        }
+        final UnsynchronizedByteArrayOutputStream output = copyToOutputStream(input, size, chunkSize);
+         if (output.size() != size) {
+            throw new EOFException("Unexpected read size, current: " + output.size() + ", expected: " + size);
```
garydgregory, line 2747: Hi @ppkarwasz 
We now have two calls to the new `copyToOutputStream` method, and in both cases, we check the return value and throw an exception (with different exceptions and styles in building the message). What about moving the check and exception throwing inside the private method, which could throw IAE.

ppkarwasz: Those two calls have slightly different checks:

- `toByteArray(InputStream)` checks if the size is **at most** `SOFT_MAX_ARRAY_LENGTH` and throws an IAE if the check fails.
- `toByteArray(InputStream, int, int)` checks if the size is **equal** `size` and throws an EOF if the check fails.

To move the check inside the method, I would need to pass an `IOConsumer` as parameter. In practice this would not change the location of the check code.

garydgregory: OK.

src/main/java/org/apache/commons/io/IOUtils.java
garydgregory: Sounds good. Let's finish this one first.

src/main/java/org/apache/commons/io/IOUtils.java
garydgregory: OK.

MarkEWaite:
As far as I can tell from `git bisect` with builds of commons-io bundled into Jenkins core, this is the pull request that caused a major regression in [Jenkins 2.537](https://www.jenkins.io/changelog/2.537/).  The major regression was:

* https://github.com/jenkinsci/jenkins/issues/16845

Jenkins agents failed to connect with SSH until we reverted Apache Commons IO 2.21.0 and returned to Apache Commons IO 2.20.0.  We released [Jenkins 2.538](https://www.jenkins.io/changelog/2.538/) less than 24 hours after 2.537 due to the severity of the issue.

The stack trace reported in the failure is:

```
[SSH] Starting sftp client.
[SSH] Copying latest remoting.jar...
java.io.IOException: Could not copy remoting.jar into '/home/jenkins/.jenkins-cd-control' on agent
    at PluginClassLoader for ssh-slaves//hudson.plugins.sshslaves.SSHLauncher.copyAgentJar(SSHLauncher.java:738)
    at PluginClassLoader for ssh-slaves//hudson.plugins.sshslaves.SSHLauncher.lambda$launch$0(SSHLauncher.java:462)
    at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
    at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
    at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
    at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: java.lang.IllegalArgumentException: invalid len argument
    at PluginClassLoader for trilead-api//com.trilead.ssh2.SFTPv3Client.read(SFTPv3Client.java:1250)
    at PluginClassLoader for trilead-api//com.trilead.ssh2.jenkins.SFTPClient$SFTPInputStream.read(SFTPClient.java:172)
    at org.apache.commons.io.input.ProxyInputStream.read(ProxyInputStream.java:337)
    at org.apache.commons.io.input.BoundedInputStream.read(BoundedInputStream.java:536)
    at org.apache.commons.io.output.AbstractByteArrayOutputStream.writeImpl(AbstractByteArrayOutputStream.java:405)
    at org.apache.commons.io.output.UnsynchronizedByteArrayOutputStream.write(UnsynchronizedByteArrayOutputStream.java:227)
    at org.apache.commons.io.IOUtils.copyToOutputStream(IOUtils.java:1958)
    at org.apache.commons.io.IOUtils.toByteArray(IOUtils.java:2918)
    at PluginClassLoader for ssh-slaves//hudson.plugins.sshslaves.SSHLauncher.readInputStreamIntoByteArrayAndClose(SSHLauncher.java:796)
    at PluginClassLoader for ssh-slaves//hudson.plugins.sshslaves.SSHLauncher.copyAgentJar(SSHLauncher.java:705)
    ... 5 more
Launch failed - cleaning up connection
[SSH] Connection closed.
```

PR #778, opened 2025-09-05
fix: correct `BoundedInputStream.getRemaining()` for unbounded streams

ppkarwasz:
`BoundedInputStream.getRemaining()` previously returned `0` when no read limit was set, which was misleading. It now returns `Long.MAX_VALUE` to more clearly represent an unbounded stream.

The Javadoc was also updated to clarify that the method reflects only the configured limit and does **not** report the actual number of bytes available in the underlying stream.

Although `Long.MAX_VALUE` represents almost 8 EiB, the behavior of `getMaxCount()` is unchanged: it still returns `EOF` (`-1`) to signal an unbounded stream. The Javadoc was updated to make this explicit, and new unit tests were added to clearly differentiate an unbounded stream from one explicitly limited to `Long.MAX_VALUE`:

* When `getMaxCount()` returns `-1`, `getRemaining()` always reports `Long.MAX_VALUE`, regardless of how many bytes have been read.
* When `getMaxCount()` returns `Long.MAX_VALUE`, `getRemaining()` begins at `Long.MAX_VALUE` but decreases as bytes are consumed.

- [x] I used AI to create Javadoc and unit test comments.


ppkarwasz:
> This PR duplicates some of or all of #779 , so pick one to work on and close the other I think.

Yes, as I noted in the description of #779, that PR is based on this change (it contains the only commit from this PR + a new one). I meant to leave that as “Draft” until this has been reviewed, but I forgot… :wink:

If you are OK with reviewing both changes together, I am closing this one.

PR #779, opened 2025-09-05
fix: respect read limit in BoundedInputStream.available()

ppkarwasz:
> [!NOTE]
> This PR depends on the fixes proposed in #778. It should not be marked as ready for review until that PR has been approved or rejected.

`BoundedInputStream.available()` previously returned the underlying stream’s availability without applying the configured read limit. It now caps the reported value by the remaining bytes allowed by the bound.

Additional changes:

* Removed the `onMaxLength` call from `available()`, since `available()` does not consume data and cannot exceed the maximum count.
* Reworked the `testAvailableAfterClose` and `testReadAfterClose` tests:

  * Old versions relied too heavily on `CharSequenceInputStream`’s specific behavior.
  * New parameterized tests cover all three typical cases of the underlying stream after close: no-op, return special value, or throw an exception.
* Added a dedicated unit test for the corrected `available()` behavior.

- [x] I used AI to improve unit tests and documentation.

### DIFF
--- a/src/main/java/org/apache/commons/io/input/BoundedInputStream.java
+++ b/src/main/java/org/apache/commons/io/input/BoundedInputStream.java
@@ -422,14 +422,16 @@ public class BoundedInputStream extends ProxyInputStream {
     }
 
     /**
-     * Gets the bytes remaining before the configured read limit is reached.
+     * Gets the number of bytes remaining to read before the maximum is reached.
      *
-     * <p>This method does <strong>not</strong> report the bytes available in the
+     * <p>
+     * This method does <strong>not</strong> report the bytes available in the
      * underlying stream; it only reflects the remaining allowance imposed by this
-     * {@code BoundedInputStream}.</p>
+     * {@code BoundedInputStream}.
+     * </p>
      *
-     * @return The number of bytes left until the read limit is reached,
-     *         or {@link Long#MAX_VALUE} if no limit is set.
+     * @return The number of bytes remaining to read before the maximum is reached,
+     *         or {@link Long#MAX_VALUE} if no bound is set.
      * @since 2.16.0
      */
     public long getRemaining() {


src/main/java/org/apache/commons/io/input/BoundedInputStream.java
```diff
@@ -405,9 +399,9 @@ public synchronized long getCount() {
     }
 
     /**
-     * Gets the max count of bytes to read.
+     * Returns the maximum number of bytes this stream is allowed to read.
```
garydgregory: In general, Commons uses the convention that a getter method "Gets something", a setter method "Sets something", and an is method "Tests something". Please do not change the style here.

src/main/java/org/apache/commons/io/input/BoundedInputStream.java
```diff
@@ -405,9 +399,9 @@ public synchronized long getCount() {
...
      *
-     * @return The max count of bytes to read.
+     * @return the maximum number of bytes permitted, or {@value IOUtils#EOF} if unbounded.
```
garydgregory: Please do not change the Commons convention that a sentence starts with a capital letter.

src/main/java/org/apache/commons/io/input/BoundedInputStream.java
```diff
@@ -427,13 +421,19 @@ public long getMaxLength() {
     }
 
     /**
-     * Gets how many bytes remain to read.
+     * Returns the bytes remaining before the configured read limit is reached.
```
garydgregory: See above about documenting getter methods.

src/test/java/org/apache/commons/io/input/BoundedInputStreamTest.java
```diff
@@ -80,21 +91,75 @@ void testAfterReadConsumer() throws Exception {
...
+        final InputStream throwsAfterClose = mock(InputStream.class);
+        when(throwsAfterClose.available()).thenReturn(42).thenThrow(new IOException("Stream closed"));
+
+        return Stream.of(
+                Arguments.of("no-op close → still returns 42", noOpClose, 42),
+                Arguments.of("after close → returns 0", returnsZeroAfterClose, 42),
+                Arguments.of("after close → throws IOException", throwsAfterClose, 42));
+    }
+
+    @DisplayName("available() after close follows underlying stream semantics")
```
garydgregory: In general, we don't use `@DisplayName`; using Javadoc is both richer in formatting and more versatile.

src/test/java/org/apache/commons/io/input/BoundedInputStreamTest.java
```diff
@@ -80,21 +91,75 @@ void testAfterReadConsumer() throws Exception {
...
+        // Case 2: returns 0 after close (Commons memory-backed streams that ignore close but report 0 when exhausted).
+        final InputStream returnsZeroAfterClose = mock(InputStream.class);
+        when(returnsZeroAfterClose.available()).thenReturn(42, 0);
+
+        // Case 3: throws IOException after close (e.g., FileInputStream-like behavior).
+        final InputStream throwsAfterClose = mock(InputStream.class);
+        when(throwsAfterClose.available()).thenReturn(42).thenThrow(new IOException("Stream closed"));
+
+        return Stream.of(
+                Arguments.of("no-op close → still returns 42", noOpClose, 42),
```
garydgregory: The arrow Unicode character is unnecessary IMO, elsewhere in this PR, you use "->", are either better explained with a word? Does arrow mean "convert", or "returns", or...?

src/test/java/org/apache/commons/io/input/BoundedInputStreamTest.java
```diff
@@ -203,10 +268,10 @@ void testCounts(final long startCount) throws Exception {
 
     @Test
     void testMarkReset() throws Exception {
-        final byte[] helloWorld = "Hello World".getBytes(StandardCharsets.UTF_8);
+        final byte[] helloWorld = "Hello World".getBytes(UTF_8);
```
garydgregory: In general, I would not change the access style in a PR because there is now more to review.

src/main/java/org/apache/commons/io/input/BoundedInputStream.java
```diff
@@ -370,16 +370,10 @@ protected synchronized void afterRead(final int n) throws IOException {
...
-     * {@inheritDoc}
-     */
     @Override
     public int available() throws IOException {
-        if (isMaxCount()) {
-            onMaxLength(maxCount, getCount());
-            return 0;
-        }
-        return in.available();
+        final int remaining = Math.toIntExact(Math.min(getRemaining(), Integer.MAX_VALUE));
```
garydgregory: This method now throws `ArithmeticException`; should we:
- Javadoc it, or,
- Convert it to an `IllegalStateException`, or,
- Something else?

garydgregory (changes requested):
Hello @ppkarwasz 

Good bug fix! 😄  I have comments scattered throughout.

src/main/java/org/apache/commons/io/input/BoundedInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/779/commits/c831b562f656559e246da43197cf8fe7d5751f45

src/main/java/org/apache/commons/io/input/BoundedInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/779/commits/c831b562f656559e246da43197cf8fe7d5751f45

src/main/java/org/apache/commons/io/input/BoundedInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/779/commits/c831b562f656559e246da43197cf8fe7d5751f45

src/test/java/org/apache/commons/io/input/BoundedInputStreamTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/779/commits/42208f18145e07a8a5b6660b593d1a8d0c4f29dd

src/test/java/org/apache/commons/io/input/BoundedInputStreamTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/779/commits/42208f18145e07a8a5b6660b593d1a8d0c4f29dd

src/test/java/org/apache/commons/io/input/BoundedInputStreamTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/779/commits/403000c3087583c673ab5b6d8b56099394edd703

src/main/java/org/apache/commons/io/input/BoundedInputStream.java
ppkarwasz: The value is between `0` (minimum value of `getRemaining()`) and `Integer.MAX_VALUE`, so the call never throws.

In https://github.com/apache/commons-io/pull/779/commits/ac15e4c30c1693df110ea754696ef702b01b59c8 I replaced it with a simple cast.

garydgregory:
Merged, thank you @ppkarwasz.

PR #781, opened 2025-09-05
feat: add length unit support in FileSystem limits

ppkarwasz:
Different filesystems and operating systems measure file and path lengths in different units:

* <s>macOS and</s> Windows filesystems typically count **UTF-16 code units**.
* Linux and other UNIX filesystems typically count **bytes**.

This change introduces explicit unit support so these limits can be interpreted consistently.

### Key changes

* **New API**

  * <s>Added a `LengthUnit` enum and `FileSystem.getLengthUnit()` to expose the unit of measure used by `getMaxFileNameLength()` and `getMaxPathLength()`</s>.
  * Added new overloads for `isLegalFileName` and `toLegalFileName` that accept a `Charset`, making conversions between bytes and UTF-16 explicit.

* **Adjusted defaults**

  * Reduced the `GENERIC` filesystem defaults:

    * File name length → **1020 bytes** (covers 255 UTF-16 characters encoded as up to 3 UTF-8 bytes).
    * Path length → **1 MiB** (covers 32,767 UTF-16 code units, again at 3 UTF-8 bytes each).

* **Testing**

  * Added unit tests to validate the new API and updated limits.

### DIFF
--- a/src/main/java/org/apache/commons/io/FileSystem.java
+++ b/src/main/java/org/apache/commons/io/FileSystem.java
@@ -424,7 +424,8 @@ public enum FileSystem {
     /**
      * Gets the maximum length for file paths (may include folders).
      *
-     * <p>This may include folder names as well as the file name.</p>
+     * <p>This value is inclusive of all path components and separators.
+     * For a limit of each path component see {@link #getMaxFileNameLength()}.</p>
      *
      * <p>The value is expressed in Java {@code char} units (UTF-16 code units)
      * and represents the longest path that can be safely passed to Java
--- a/src/main/java/org/apache/commons/io/FileSystem.java
+++ b/src/main/java/org/apache/commons/io/FileSystem.java
@@ -676,10 +676,8 @@ public enum FileSystem {
 
             private int safeCutPoint(final CharSequence value, final int limit) {
                 // Ensure we do not cut a surrogate pair in half.
-                if (Character.isHighSurrogate(value.charAt(limit - 1))) {
-                    if (Character.isLowSurrogate(value.charAt(limit))) {
-                        return limit - 1;
-                    }
+                if (Character.isHighSurrogate(value.charAt(limit - 1)) && Character.isLowSurrogate(value.charAt(limit))) {
+                    return limit - 1;
                 }
                 return limit;
             }
--- a/src/main/java/org/apache/commons/io/FileSystem.java
+++ b/src/main/java/org/apache/commons/io/FileSystem.java
@@ -17,6 +17,8 @@
 
 package org.apache.commons.io;
 
+import static java.nio.charset.StandardCharsets.UTF_16;
+
 import java.nio.ByteBuffer;
 import java.nio.CharBuffer;
 import java.nio.charset.CharacterCodingException;
@@ -653,6 +655,11 @@ public enum FileSystem {
 
             @Override
             CharSequence truncate(final CharSequence value, final int limit, final Charset charset) {
+                if (!UTF_16.newEncoder().canEncode(value)) {
+                    throw new IllegalArgumentException(
+                            "The value " + value + " can not be encoded using " + UTF_16.name());
+                }
+
                 // Fast path: no truncation needed.
                 if (value.length() <= limit) {
                     return value;
@@ -675,7 +682,7 @@ public enum FileSystem {
 
             private int safeCutPoint(final CharSequence value, final int limit) {
                 // Ensure we do not cut a surrogate pair in half.
-                if (Character.isHighSurrogate(value.charAt(limit - 1)) && Character.isLowSurrogate(value.charAt(limit))) {
+                if (Character.isHighSurrogate(value.charAt(limit - 1))) {
                     return limit - 1;
                 }
                 return limit;
--- a/src/main/java/org/apache/commons/io/FileSystem.java
+++ b/src/main/java/org/apache/commons/io/FileSystem.java
@@ -26,6 +26,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.CharsetEncoder;
 import java.nio.charset.CoderResult;
 import java.nio.charset.CodingErrorAction;
+import java.text.BreakIterator;
 import java.util.Arrays;
 import java.util.Locale;
 import java.util.Objects;
@@ -570,10 +571,43 @@ public enum FileSystem {
         return new String(array, 0, array.length);
     }
 
-    CharSequence trimExtension(final CharSequence cs) {
+    static CharSequence trimExtension(final CharSequence cs) {
         final int index = indexOfFirstDot(cs);
         // An initial dot is not an extension
         return index < 1 ? cs : cs.subSequence(0, index);
+    }
+
+    static CharSequence[] splitExtension(final CharSequence value) {
+        final int index = indexOfFirstDot(value);
+        // An initial dot is not an extension
+        return index < 1
+                ? new CharSequence[] {value, ""}
+                : new CharSequence[] {value.subSequence(0, index), value.subSequence(index, value.length())};
+    }
+
+    /**
+     * Truncates a string respecting grapheme cluster boundaries.
+     *
+     * @param value The value to truncate.
+     * @param limit The maximum length.
+     * @return The truncated value.
+     * @throws IllegalArgumentException If the first grapheme cluster is longer than the limit.
+     */
+    private static CharSequence safeTruncate(final CharSequence value, final int limit) {
+        if (value.length() <= limit) {
+            return value;
+        }
+        final BreakIterator boundary = BreakIterator.getCharacterInstance(Locale.ROOT);
+        final String text = value.toString();
+        boundary.setText(text);
+        final int end = boundary.preceding(limit + 1);
+        assert end != BreakIterator.DONE;
+        if (end == 0) {
+            final String limitMessage = limit <= 1 ? "1 character" : limit + " characters";
+            throw new IllegalArgumentException("The value " + value + " can not be truncated to " + limitMessage
+                    + " without breaking the first codepoint or grapheme cluster");
+        }
+        return text.substring(0, end);
     }
 
     /**
@@ -614,22 +648,19 @@ public enum FileSystem {
 
                 // Slow path: encode into a fixed-size byte buffer.
                 // 1. Compute length of extension in bytes (if any).
-                final int extensionStart = indexOfFirstDot(value);
-                final boolean hasExtension = extensionStart > 0;
-                final int extensionLength = hasExtension ? getLength(value.subSequence(extensionStart, value.length()), charset) : 0;
-                if (hasExtension && extensionLength >= limit) {
+                final CharSequence[] parts = splitExtension(value);
+                final int extensionLength = getLength(parts[1], charset);
+                if (extensionLength > 0 && extensionLength >= limit) {
                     // Extension itself does not fit
                     throw new IllegalArgumentException(
                             "The extension of " + value + " is too long to fit within " + limit + " bytes");
                 }
 
-                // 2. Compute the byte size of the non-extension part.
+                // 2. Compute the character part that fits within the remaining byte budget.
                 final ByteBuffer byteBuffer = ByteBuffer.allocate(limit - extensionLength);
-                final CharBuffer charBuffer = CharBuffer.allocate(value.length());
+                final CharBuffer charBuffer = CharBuffer.wrap(parts[0]);
 
                 // Encode until the first character that would exceed the byte budget.
-                charBuffer.append(value, 0, hasExtension ? extensionStart : value.length());
-                charBuffer.rewind();
                 final CoderResult cr = encoder.encode(charBuffer, byteBuffer, true);
 
                 if (cr.isUnderflow()) {
@@ -637,12 +668,8 @@ public enum FileSystem {
                     return value;
                 }
 
-                // We ran out of space mid-encode: append the extension (if any) and return the charBuffer.
-                if (hasExtension) {
-                    charBuffer.append(value, extensionStart, value.length());
-                }
-                charBuffer.flip();
-                return charBuffer;
+                final CharSequence truncated = safeTruncate(value, charBuffer.position());
+                return extensionLength == 0 ? truncated : truncated.toString() + parts[1];
             }
         },
 
@@ -664,28 +691,20 @@ public enum FileSystem {
                 if (value.length() <= limit) {
                     return value;
                 }
+
                 // Slow path: truncate to limit.
                 // 1. Compute length of extension in chars (if any).
-                final int extensionStart = indexOfFirstDot(value);
-                final int extensionLength = extensionStart > 0 ? value.length() - extensionStart : 0;
-                // 2. Truncate the non-extension part and append the extension (if any).
-                if (extensionLength >= limit) {
+                final CharSequence[] parts = splitExtension(value);
+                final int extensionLength = parts[1].length();
+                if (extensionLength > 0 && extensionLength >= limit) {
                     // Extension itself does not fit
-                    throw new IllegalArgumentException("The extension of " + value + " is too long to fit within " + limit + " characters");
+                    throw new IllegalArgumentException(
+                            "The extension of " + value + " is too long to fit within " + limit + " characters");
                 }
-                final int safeLimit = safeCutPoint(value, limit - extensionLength);
-                if (extensionLength == 0) {
-                    return value.subSequence(0, safeLimit);
-                }
-                return value.subSequence(0, safeLimit).toString() + value.subSequence(extensionStart, value.length());
-            }
 
-            private int safeCutPoint(final CharSequence value, final int limit) {
-                // Ensure we do not cut a surrogate pair in half.
-                if (Character.isHighSurrogate(value.charAt(limit - 1))) {
-                    return limit - 1;
-                }
-                return limit;
+                // 2. Truncate the non-extension part and append the extension (if any).
+                final CharSequence truncated = safeTruncate(value, limit - extensionLength);
+                return extensionLength == 0 ? truncated : truncated.toString() + parts[1];
             }
         };
 
--- a/src/main/java/org/apache/commons/io/FileSystem.java
+++ b/src/main/java/org/apache/commons/io/FileSystem.java
@@ -212,7 +212,7 @@ public enum FileSystem {
         }
     }
 
-    /**
+    /*
      * Finds the index of the first dot in a CharSequence.
      */
     private static int indexOfFirstDot(final CharSequence cs) {
--- a/src/main/java/org/apache/commons/io/FileSystem.java
+++ b/src/main/java/org/apache/commons/io/FileSystem.java
@@ -48,7 +48,7 @@ public enum FileSystem {
             // ASCII NUL
             0
             // @formatter:on
-    }, new String[] {}, false, false, '/', LengthUnit.BYTES),
+    }, new String[] {}, false, false, '/', NameLengthStrategy.BYTES),
 
     /**
      * Linux file system.
@@ -60,7 +60,7 @@ public enum FileSystem {
             0,
              '/'
             // @formatter:on
-    }, new String[] {}, false, false, '/', LengthUnit.BYTES),
+    }, new String[] {}, false, false, '/', NameLengthStrategy.BYTES),
 
     /**
      * MacOS file system.
@@ -73,7 +73,7 @@ public enum FileSystem {
             '/',
              ':'
             // @formatter:on
-    }, new String[] {}, false, false, '/', LengthUnit.CHARS),
+    }, new String[] {}, false, false, '/', NameLengthStrategy.UTF16_CHARS),
 
     /**
      * Windows file system.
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -329,10 +342,12 @@ private static String replace(final String path, final char oldChar, final char
      * @param reservedFileNamesExtensions TODO
      * @param supportsDriveLetter Whether this file system support driver letters.
      * @param nameSeparator The name separator, '\\' on Windows, '/' on Linux.
+     * @param lengthUnit The unit of measurement for length limits.
```
garydgregory: The Javadoc should make explicit _what_ the measurement applies to. The length of ...

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -380,23 +396,45 @@ public int[] getIllegalFileNameCodePoints() {
...
     /**
-     * Gets the maximum length for file names. The file name does not include folders.
+     * Gets the maximum length for file names (excluding any folder path).
+     *
+     * <p><strong>Note:</strong> This excludes any folder path. The unit depends on the
+     * filesystem or OS; see {@link #getLengthUnit()} to check whether the value is in
+     * bytes or UTF-16 characters.</p>
      *
-     * @return the maximum length for file names.
+     * @return the maximum file name length.
```
garydgregory: Should we now specify the scale of this length? Length in ... `char`s?

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -380,23 +396,45 @@ public int[] getIllegalFileNameCodePoints() {
...
     /**
-     * Gets the maximum length of the path to a file. This can include folders.
+     * Gets the maximum length for file paths (may include folders).
      *
-     * @return the maximum length of the path to a file.
+     * <p><strong>Note:</strong> This may include folder names as well as the file name.
+     * The unit is the same as {@link #getMaxFileNameLength()} and can be obtained
+     * from {@link #getLengthUnit()}.</p>
+     *
+     * @return the maximum file path length.
```
garydgregory: Should we now specify the scale of this length? Length in ... chars?

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -530,4 +623,76 @@ CharSequence trimExtension(final CharSequence cs) {
...
+        // Character-based limit: simple substring if needed.
+        if (lengthUnit == LengthUnit.CHARS) {
+            return candidate.length() <= maxFileNameLength ? candidate : candidate.subSequence(0, maxFileNameLength);
+        }
+
+        // Byte-based limit
+        return truncateByBytes(candidate, charset, maxFileNameLength);
+    }
+
+    static CharSequence truncateByBytes(final CharSequence candidate, final Charset charset, final int maxBytes) {
```
garydgregory: What does "truncateByBytes" mean?

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -530,4 +623,76 @@ CharSequence trimExtension(final CharSequence cs) {
         final int index = indexOf(cs, '.', 0);
         return index < 0 ? cs : cs.subSequence(0, index);
     }
+
+    private boolean isLegalFileLength(final CharSequence candidate, final Charset charset) {
+        if (candidate == null || candidate.length() == 0) {
+            return false;
+        }
+        if (lengthUnit == LengthUnit.CHARS) {
```
garydgregory: If we are going to add a LengthUnit class, then we should delegate these if/else computations to the class.

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -530,4 +623,76 @@ CharSequence trimExtension(final CharSequence cs) {
...
+        } catch (CharacterCodingException e) {
+            // If we can't encode, it's not legal
+            return false;
+        }
+    }
+
+    CharSequence truncateFileName(final CharSequence candidate, final Charset charset) {
+        final int maxFileNameLength = getMaxFileNameLength();
+        // Character-based limit: simple substring if needed.
+        if (lengthUnit == LengthUnit.CHARS) {
```
garydgregory: If we are going to add a LengthUnit class, then we should delegate these if/else computations to the class.

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -530,4 +623,76 @@ CharSequence trimExtension(final CharSequence cs) {
...
+            return candidate;
+        }
+
+        // We ran out of space mid-encode: truncate BEFORE the offending character.
+        return candidate.subSequence(0, in.position());
+    }
+
+    /**
+     * Units of length for the file name and path length limits.
+     *
```
garydgregory: I can't quite understand why this is needed and how it works. Is this saying that different OSs count file names on disk differently? Would this counting depend on the OS file encoding?

garydgregory (changes requested):
Hi @ppkarwasz 

I have a question on the LengthUnit class and a few comments.

TY!

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: I should have explained it more clearly.

**File name length**

* The maximum length of a *single file or directory name* is defined by the **filesystem**, not the OS.
* On Windows (NTFS) and macOS (APFS, HFS+), names are stored internally as **UTF-16 code units**. The limit is expressed in those code units (e.g. 255). Because the filesystem enforces UTF-16 internally, the user-visible encoding doesn’t affect the limit.
* On most Linux/UNIX filesystems (ext4, XFS, Btrfs), names are just opaque **byte strings** in directory entries. The usual limit is 255 bytes. If your system locale is UTF-8, that translates to 255 ASCII characters, but only 85 `★` characters (3 bytes each).

These are **hard limits**: you simply cannot store a longer name on those filesystems.

**Path length**

* The length of an *entire path* (all components combined) is typically an **OS-level API restriction**, not a filesystem restriction. The filesystem doesn’t know where it has been mounted, so it cannot enforce a total path length.
* On POSIX systems, the constants `PATH_MAX` (often 4096) and `NAME_MAX` (255) are defined in bytes, because paths are passed as `char*`.
* On Windows, path limits are defined in UTF-16 characters. NTFS itself allows very long paths (32,767 chars), but older Win32 APIs traditionally limited paths to 260 characters unless special prefixes are used.
* On macOS, it depends which API you use: the POSIX layer (`open`, `getcwd`) works in UTF-8 byte strings and enforces `PATH_MAX` in bytes, while Apple’s higher-level APIs often work in UTF-16 and measure in code units.

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/781/commits/9a1e9e41df93bf2e2c442637e9c5ba2c78d7dc72

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: I don't know what the original intention was, but `char`s seems the only length unit useful for users.

The maximum length is always reachable by using ASCII characters, I just added a note in https://github.com/apache/commons-io/pull/781/commits/9a1e9e41df93bf2e2c442637e9c5ba2c78d7dc72 that satisfying the limit in `char`s might not be enough.

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/781/commits/9a1e9e41df93bf2e2c442637e9c5ba2c78d7dc72

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: I moved this method to `LengthUnit` and renamed it `truncate`: https://github.com/apache/commons-io/pull/781/commits/9a1e9e41df93bf2e2c442637e9c5ba2c78d7dc72

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: I moved the methods to `LengthUnit` (and renamed it `NameLengthStrategy`): https://github.com/apache/commons-io/pull/781/commits/9a1e9e41df93bf2e2c442637e9c5ba2c78d7dc72

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: Same as above, fixed in https://github.com/apache/commons-io/pull/781/commits/9a1e9e41df93bf2e2c442637e9c5ba2c78d7dc72

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -107,7 +107,7 @@ public enum FileSystem {
                     "LPT1", "LPT2", "LPT3", "LPT4", "LPT5", "LPT6", "LPT7", "LPT8", "LPT9",
                     "LPT\u00b2", "LPT\u00b3", "LPT\u00b9", // Superscript 2 3 1 in that order
                     "NUL", "PRN"
-            }, true, true, '\\', LengthUnit.CHARS);
+            }, true, true, '\\', NameLengthStrategy.UTF16_CHARS);
```
ecki: Maybe differentiate code points and code units (and clusters?) instead of using “chars”? Also should it support max path length and path components and depth etc? )I think the problem is generally too big tomwddress)

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -342,12 +343,12 @@ private static String replace(final String path, final char oldChar, final char
      * @param reservedFileNamesExtensions TODO
      * @param supportsDriveLetter Whether this file system support driver letters.
      * @param nameSeparator The name separator, '\\' on Windows, '/' on Linux.
-     * @param lengthUnit The unit of measurement for length limits.
+     * @param nameLengthStrategy The strategy for measuring and truncating file and path names.
```
ecki, line 290: I like this general change, it can cover more limits

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -411,30 +424,26 @@ public int getMaxFileNameLength() {
     /**
      * Gets the maximum length for file paths (may include folders).
      *
-     * <p><strong>Note:</strong> This may include folder names as well as the file name.
-     * The unit is the same as {@link #getMaxFileNameLength()} and can be obtained
-     * from {@link #getLengthUnit()}.</p>
+     * <p>This may include folder names as well as the file name.</p>
```
ecki: “May include folder names” sounds a bit unclear to me, is this the path length (including separators and roots) for folders and files or is it for path components?

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -519,13 +528,11 @@ public boolean isLegalFileName(final CharSequence candidate) {
...
-        if (!isLegalFileLength(candidate, charset)) {
-            return false;
-        }
-        if (isReservedFileName(candidate)) {
-            return false;
-        }
-        return candidate.chars().noneMatch(this::isIllegalFileNameChar);
+        return candidate != null
+                && candidate.length() != 0
+                && nameLengthStrategy.isWithinLimit(candidate, getMaxFileNameLength(), charset)
```
ecki, line 478: I think passing the candidate to the strategy would be better, this could then also implements things like “no more than 255 utf16 code units and no more than 300mbytes utf8 serialized”

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -624,75 +631,102 @@ CharSequence trimExtension(final CharSequence cs) {
...
-        return candidate.subSequence(0, in.position());
+        /**
+         * Truncates to {@code limit} in this strategy’s unit (no-op if already within limit).
+         *
+         * @param value The value to truncate, not null.
+         * @param limit The limit to truncate to.
+         * @param charset The charset to use when measuring in bytes.
+         * @return The truncated value, not null.
+         */
+        abstract CharSequence truncate(CharSequence value, int limit, Charset charset);
```
ecki, line 740: Should truncate at least be able to preserve extensions? Also make sure that after truncation file name is still valid. And can truncation also affect parents? Should there be a minimum length the remaining name has (don’t truncate to a single char, etc)

ppkarwasz: Good point, I will fix it.

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: Could you expand a bit on what you mean?

The main goal of my PR is to highlight that UTF-16 code units are almost never the unit of measure used for file name limits. That severely constrains how the values returned by `getMaxFileNameLength()` and `getMaxPathLength()` can be interpreted.

My immediate motivation is to provide reasonable, conservative limits for file names inside archives, so that corrupted or malicious values can be detected early. For example a file entry with a name larger than `getMaxPathLength() * maxBytesPerChar` could be immediately marked as invalid.

At first I looked only at **filesystem constraints**, but I realized I should also consider the **Java APIs** that sit on top of them, since the JDK’s abstractions can impose their own effective limits. That naturally leads to two broad categories:

* **Windows**: API calls use UTF-16 code units directly. That makes it relatively straightforward to apply file name and path length checks in Java. Interestingly, the JDK `Path` implementation lowers the theoretical maximum from 32,767 UTF-16 units to 32,000 and throws if the limit is exceeded.
* **POSIX-based OSes (Linux, macOS)**: limits are defined in **bytes** (`NAME_MAX`, `PATH_MAX`), not UTF-16 units. The behavior for names longer than those limits is unspecified. For example, HFS+ allows 255 UTF-16 code units per name, but macOS’s POSIX layer only guarantees 255 **UTF-8 bytes**, so names that are legal on disk may still be inaccessible via standard APIs.

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -315,6 +327,8 @@ private static String replace(final String path, final char oldChar, final char
     private final boolean supportsDriveLetter;
     private final char nameSeparator;
     private final char nameSeparatorOther;
+    // package-private for testing
+    final NameLengthStrategy nameLengthStrategy;
```
garydgregory: This should be private like the other instance variables.

garydgregory (changes requested):
Hi @ppkarwasz 

I have one small comment. Have @ecki's issues been resolved?

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/781/commits/a0b99bfdb01ce7cc218475100f33ff182a37a8ed.

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/781/commits/b1cdfdb834cd795d37d9ac26b4469d4e070789b5

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: This is used in a test. I changed the test and made this private in https://github.com/apache/commons-io/pull/781/commits/ff205f8d09cadd2508797a6ae3a95c4941a976e8.

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: I am not entirely sure, what are you suggesting? The `candidate` is passed to the strategy.

Are you referring to the null check and `candidate.length() != 0`?

ecki: I am not sure either, probably missread the diff, please ignore the comment, looks good this way.

src/main/java/org/apache/commons/io/FileSystem.java
ecki: I agree with your differentiation, I was not sure the strategy addresses all of them (and I am not sure if the CHARS in the name is the right terminology. but maybe its clearer than CODEPOINTS).

src/main/java/org/apache/commons/io/FileSystem.java
ecki: in some conditions there are also different path styles, I think on windows the jdk even used \\?\ syntax if the path is longer than 256 to get around a api limitation. (Not sure I remeber that correctly). That was the reason why the prefix of the path also counts for special strategies. (So maybe you need a WINDOWS_UTF16_CHARS strategy (not a length mode)?

src/main/java/org/apache/commons/io/FileSystem.java
ppkarwasz: `CODEPOINTS` sounds good to me! After all, this is a private class and it can be changed later.

src/test/java/org/apache/commons/io/FileSystemTest.java
```diff
@@ -128,6 +190,42 @@ void testSorted() {
...
+                throw new IllegalStateException("Unexpected value: " + fs);
+        }
+        for (final String fileName : validNames) {
+            assertDoesNotThrow(() -> testFileName(tempDir, fileName), "OS accepts max length name");
+            assertTrue(fs.isLegalFileName(fileName, UTF_8), "Commons IO accepts max length name");
+            final String tooLongName = fileName + "a";
+            assertThrows(IOException.class, () -> testFileName(tempDir, tooLongName), "OS rejects too-long name");
+            assertFalse(fs.isLegalFileName(tooLongName, UTF_8), "Commons IO rejects too-long name");
+        }
+    }
```
ppkarwasz: @garydgregory,

Could you run this test on a MacOS with an APFS file system? MacOS is a little bit of a mystery to me:

- File access in Java is implemented using the POSIX API, so the limits are measured in UTF-8 bytes.
- HFS+ has a limit in UTF-16 code points.
- While the POSIX specification does not require calls to fail if the file name length is greater than `NAME_MAX`, the behavior is undefined. MacOS apparently allows the usage of more than 255 UTF-8 bytes if the underlying file system supports it.

src/main/java/org/apache/commons/io/FileSystem.java
```diff
@@ -504,30 +517,201 @@ public boolean supportsDriveLetter() {
...
+                try {
+                    return enc.encode(CharBuffer.wrap(value)).remaining();
+                } catch (CharacterCodingException e) {
+                    // Unencodable, does not fit any byte limit.
+                    return Integer.MAX_VALUE;
+                }
+            }
+
+            @Override
+            CharSequence truncate(final CharSequence value, final int limit, final Charset charset) {
```
garydgregory, line 634: We are probably biting off more than we can chew here, considering issues like https://issues.apache.org/jira/projects/LANG/issues/LANG-1770. Maybe we should do a "hard" truncate and Javadoc the possible gotcha.

ppkarwasz: The two `truncate` strategies already avoid splitting valid Unicode code points.

* In the `UTF16_CODE_UNITS` strategy we explicitly handle surrogate pairs. For example:

  * `"😀😀.txt"` is 8 UTF-16 code units. With a limit of 7, the result is `"😀.txt"` (6 code units): we step back so we don’t cut inside a surrogate pair.
  * If the input has an *invalid* sequence, e.g. `"\uD83Da"`, then the code allows truncation between `\uD83D` and `a`.

There is a unit test for this:

https://github.com/apache/commons-io/blob/b99664a6cf020da89df05af827c5f0274e69658a/src/test/java/org/apache/commons/io/FileSystemTest.java#L288-L295

src/test/java/org/apache/commons/io/FileSystemTest.java
ppkarwasz: I modified the tests in https://github.com/apache/commons-io/pull/781/commits/087654e6067daaae8d9309d6351df9603596ed68 to work both on HFS+ and APFS, but a confirmation that it works on APFS would be nice.

src/main/java/org/apache/commons/io/FileSystem.java
garydgregory: I'm not sure this works for more complex examples in https://issues.apache.org/jira/projects/LANG/issues/LANG-1770 for strings like "👩🏻‍👨🏻‍👦🏻‍👦🏻👩🏼‍👨🏼‍👦🏼‍👦🏼👩🏽‍👨🏽‍👦🏽‍👦🏽👩🏾‍👨🏾‍👦🏾‍👦🏾👩🏿‍👨🏿‍👦🏿‍👦🏿👩🏻‍👨🏻‍👦🏻‍👦🏻👩🏼‍👨🏼‍👦🏼‍👦🏼👩🏽‍👨🏽‍👦🏽‍👦🏽👩🏾‍👨🏾‍👦🏾‍👦🏾👩🏿‍👨🏿‍👦🏿‍👦🏿"

ppkarwasz: You are right! :100:

Grapheme clusters were not handled correctly. I fixed this in https://github.com/apache/commons-io/pull/781/commits/81a33ae082b87ea2f8d95feba7d3bdac1abb5bf7.

PR #784, opened 2025-09-21
feat: add NIO channel support to `AbstractStreamBuilder`

ppkarwasz:
Some builders derived from `AbstractStreamBuilder` (for example, `SevenZFile.Builder` in Commons Compress) need to produce a `SeekableByteChannel` as their data source. Until now this required ad-hoc `instanceof` switches across different origin types.

This change integrates channel support directly into the origin/builder abstraction, leading to a cleaner and more object-oriented design.

### Key changes

* Add `getReadableByteChannel()` and `getWritableByteChannel()` to `AbstractOrigin` and propagate to `AbstractStreamBuilder`.
* Introduce `ChannelOrigin`, an `AbstractOrigin` implementation backed by an existing `ReadableByteChannel`/`WritableByteChannel`.
* Add `ByteArrayChannel`, a simple in-memory `SeekableByteChannel` implementation.
* Extend unit tests to cover the new methods and types.

- [x] I used AI to improve the PR description.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -62,6 +62,8 @@ The <action> type attribute can be add,update,fix,remove.
       <action dev="ggregory" type="add"                due-to="Gary Gregory">Add org.apache.commons.io.output.ProxyOutputStream.writeRepeat(int, long).</action>
       <action dev="pkarwasz" type="add"                due-to="Piotr P. Karwasz">Add length unit support in FileSystem limits.</action>
       <action dev="pkarwasz" type="add"                due-to="Piotr P. Karwasz">Add IOUtils.toByteArray(InputStream, int, int) for safer chunked reading with size validation.</action>
+      <action dev="pkarwasz" type="add"                due-to="Piotr P. Karwasz">Add ByteArrayChannel, a simple in-memory `SeekableByteChannel` implementation.</action>
+      <action dev="pkarwasz" type="add"                due-to="Piotr P. Karwasz">Add NIO channel support to `AbstractStreamBuilder`.</action>
       <!-- UPDATE -->
       <action type="update" dev="ggregory"             due-to="Gary Gregory, Dependabot">Bump org.apache.commons:commons-parent from 85 to 87 #774.</action>
       <action type="update" dev="ggregory"             due-to="Gary Gregory">[test] Bump commons-codec:commons-codec from 1.18.0 to 1.19.0.</action>
--- a/src/main/java/org/apache/commons/io/build/AbstractOrigin.java
+++ b/src/main/java/org/apache/commons/io/build/AbstractOrigin.java
@@ -14,11 +14,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.apache.commons.io.build;
 
 import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
@@ -29,13 +29,17 @@ import java.io.RandomAccessFile;
 import java.io.Reader;
 import java.io.Writer;
 import java.net.URI;
+import java.nio.channels.Channel;
+import java.nio.channels.Channels;
+import java.nio.channels.ReadableByteChannel;
+import java.nio.channels.SeekableByteChannel;
+import java.nio.channels.WritableByteChannel;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.OpenOption;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
-import java.nio.file.spi.FileSystemProvider;
 import java.util.Arrays;
 import java.util.Objects;
 
@@ -44,7 +48,7 @@ import org.apache.commons.io.IORandomAccessFile;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.io.RandomAccessFileMode;
 import org.apache.commons.io.RandomAccessFiles;
-import org.apache.commons.io.file.spi.FileSystemProviders;
+import org.apache.commons.io.channels.ByteArrayChannel;
 import org.apache.commons.io.input.BufferedFileChannelInputStream;
 import org.apache.commons.io.input.CharSequenceInputStream;
 import org.apache.commons.io.input.CharSequenceReader;
@@ -53,15 +57,246 @@ import org.apache.commons.io.output.RandomAccessFileOutputStream;
 import org.apache.commons.io.output.WriterOutputStream;
 
 /**
- * Abstracts the origin of data for builders like a {@link File}, {@link Path}, {@link Reader}, {@link Writer}, {@link InputStream}, {@link OutputStream}, and
- * {@link URI}.
+ * Abstract base class that encapsulates the <em>origin</em> of data used by Commons IO builders.
  * <p>
- * Some methods may throw {@link UnsupportedOperationException} if that method is not implemented in a concrete subclass, see {@link #getFile()} and
- * {@link #getPath()}.
+ * An origin represents where bytes/characters come from or go to, such as a {@link File}, {@link Path},
+ * {@link Reader}, {@link Writer}, {@link InputStream}, {@link OutputStream}, or {@link URI}. Concrete subclasses
+ * expose only the operations that make sense for the underlying source or sink; invoking an unsupported operation
+ * results in {@link UnsupportedOperationException} (see, for example, {@link #getFile()} and {@link #getPath()}).
  * </p>
  *
- * @param <T> the type of instances to build.
- * @param <B> the type of builder subclass.
+ * <p>
+ * The table below summarizes which views and conversions are supported for each origin type.
+ * Column headers show the target view; cells indicate whether that view is available from the origin in that row.
+ * </p>
+ *
+ * <table>
+ *   <caption>Origin support matrix</caption>
+ *   <thead>
+ *     <tr>
+ *       <th>Origin Type</th>
+ *       <th>byte[]</th>
+ *       <th>CS</th>
+ *       <th>File</th>
+ *       <th>Path</th>
+ *       <th>RAF</th>
+ *       <th>IS</th>
+ *       <th>Reader</th>
+ *       <th>RBC</th>
+ *       <th>OS</th>
+ *       <th>Writer</th>
+ *       <th>WBC</th>
+ *     </tr>
+ *   </thead>
+ *   <tbody>
+ *     <tr>
+ *       <td>byte[]</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *     </tr>
+ *     <tr>
+ *       <td>CharSequence (CS)</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✔<sup>1</sup></td>
+ *       <td>✔</td>
+ *       <td>✔<sup>1</sup></td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *     </tr>
+ *     <tr>
+ *       <td>File</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *     </tr>
+ *     <tr>
+ *       <td>Path</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *     </tr>
+ *     <tr>
+ *       <td>IORandomAccessFile</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *     </tr>
+ *     <tr>
+ *       <td>RandomAccessFile (RAF)</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *     </tr>
+ *     <tr>
+ *       <td>InputStream (IS)</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *     </tr>
+ *     <tr>
+ *       <td>Reader</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✔<sup>1</sup></td>
+ *       <td>✔</td>
+ *       <td>✔<sup>1</sup></td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *     </tr>
+ *     <tr>
+ *       <td>ReadableByteChannel (RBC)</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *     </tr>
+ *     <tr>
+ *       <td>OutputStream (OS)</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *     </tr>
+ *     <tr>
+ *       <td>Writer</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✔<sup>1</sup></td>
+ *       <td>✔</td>
+ *       <td>✔<sup>1</sup></td>
+ *     </tr>
+ *     <tr>
+ *       <td>WritableByteChannel (WBC)</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *     </tr>
+ *     <tr>
+ *       <td>URI (FileSystem)</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *     </tr>
+ *     <tr>
+ *       <td>URI (http/https))</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✔</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *       <td>✖</td>
+ *     </tr>
+ *   </tbody>
+ * </table>
+ *
+ * <p><strong>Legend</strong></p>
+ * <ul>
+ *   <li>✔ = Supported</li>
+ *   <li>✖ = Not supported (throws {@link UnsupportedOperationException})</li>
+ *   <li><sup>1</sup> = Characters are converted to bytes using the default {@link Charset}.</li>
+ * </ul>
+ *
+ * @param <T> the type produced by the builder.
+ * @param <B> the concrete builder subclass type.
  * @since 2.12.0
  */
 public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends AbstractSupplier<T, B> {
@@ -138,6 +373,16 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
         }
 
         @Override
+        public ReadableByteChannel getReadableByteChannel(OpenOption... options) throws IOException {
+            return getRandomAccessFile(options).getChannel();
+        }
+
+        @Override
+        public WritableByteChannel getWritableByteChannel(OpenOption... options) throws IOException {
+            return getRandomAccessFile(options).getChannel();
+        }
+
+        @Override
         public long size() throws IOException {
             return origin.length();
         }
@@ -171,7 +416,7 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
          */
         @Override
         public InputStream getInputStream(final OpenOption... options) throws IOException {
-            return new ByteArrayInputStream(origin);
+            return new ByteArrayInputStream(getByteArray());
         }
 
         @Override
@@ -180,10 +425,14 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
         }
 
         @Override
+        public ReadableByteChannel getReadableByteChannel(OpenOption... options) throws IOException {
+            return ByteArrayChannel.wrap(getByteArray());
+        }
+
+        @Override
         public long size() throws IOException {
             return origin.length;
         }
-
     }
 
     /**
@@ -203,7 +452,7 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
         @Override
         public byte[] getByteArray() {
             // TODO Pass in a Charset? Consider if call sites actually need this.
-            return origin.toString().getBytes(Charset.defaultCharset());
+            return getCharSequence(null).toString().getBytes(Charset.defaultCharset());
         }
 
         /**
@@ -227,7 +476,7 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
         @Override
         public InputStream getInputStream(final OpenOption... options) throws IOException {
             // TODO Pass in a Charset? Consider if call sites actually need this.
-            return CharSequenceInputStream.builder().setCharSequence(getCharSequence(Charset.defaultCharset())).get();
+            return CharSequenceInputStream.builder().setCharSequence(getCharSequence(null)).get();
         }
 
         /**
@@ -238,7 +487,12 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
          */
         @Override
         public Reader getReader(final Charset charset) throws IOException {
-            return new CharSequenceReader(get());
+            return new CharSequenceReader(getCharSequence(charset));
+        }
+
+        @Override
+        public ReadableByteChannel getReadableByteChannel(OpenOption... options) throws IOException {
+            return ByteArrayChannel.wrap(getByteArray());
         }
 
         @Override
@@ -267,7 +521,7 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
 
         @Override
         public byte[] getByteArray(final long position, final int length) throws IOException {
-            try (RandomAccessFile raf = RandomAccessFileMode.READ_ONLY.create(origin)) {
+            try (RandomAccessFile raf = RandomAccessFileMode.READ_ONLY.create(getFile())) {
                 return RandomAccessFiles.read(raf, position, length);
             }
         }
@@ -280,9 +534,18 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
 
         @Override
         public Path getPath() {
-            return get().toPath();
+            return getFile().toPath();
         }
 
+        @Override
+        public ReadableByteChannel getReadableByteChannel(OpenOption... options) throws IOException {
+            return Files.newByteChannel(getPath(), options);
+        }
+
+        @Override
+        public WritableByteChannel getWritableByteChannel(OpenOption... options) throws IOException {
+            return Files.newByteChannel(getPath(), options);
+        }
     }
 
     /**
@@ -304,7 +567,7 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
 
         @Override
         public byte[] getByteArray() throws IOException {
-            return IOUtils.toByteArray(origin);
+            return IOUtils.toByteArray(getInputStream());
         }
 
         /**
@@ -324,6 +587,18 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
             return new InputStreamReader(getInputStream(), Charsets.toCharset(charset));
         }
 
+        @Override
+        public ReadableByteChannel getReadableByteChannel(OpenOption... options) throws IOException {
+            return Channels.newChannel(getInputStream(options));
+        }
+
+        public long size() throws IOException {
+            if (origin instanceof FileInputStream) {
+                final FileInputStream fileInputStream = (FileInputStream) origin;
+                return fileInputStream.getChannel().size();
+            }
+            throw unsupportedOperation("size");
+        }
     }
 
     /**
@@ -392,7 +667,12 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
          */
         @Override
         public Writer getWriter(final Charset charset, final OpenOption... options) throws IOException {
-            return new OutputStreamWriter(origin, Charsets.toCharset(charset));
+            return new OutputStreamWriter(getOutputStream(options), Charsets.toCharset(charset));
+        }
+
+        @Override
+        public WritableByteChannel getWritableByteChannel(OpenOption... options) throws IOException {
+            return Channels.newChannel(getOutputStream(options));
         }
     }
 
@@ -415,12 +695,12 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
 
         @Override
         public byte[] getByteArray(final long position, final int length) throws IOException {
-            return RandomAccessFileMode.READ_ONLY.apply(origin, raf -> RandomAccessFiles.read(raf, position, length));
+            return RandomAccessFileMode.READ_ONLY.apply(getPath(), raf -> RandomAccessFiles.read(raf, position, length));
         }
 
         @Override
         public File getFile() {
-            return get().toFile();
+            return getPath().toFile();
         }
 
         @Override
@@ -429,6 +709,15 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
             return get();
         }
 
+        @Override
+        public ReadableByteChannel getReadableByteChannel(OpenOption... options) throws IOException {
+            return Files.newByteChannel(getPath(), options);
+        }
+
+        @Override
+        public WritableByteChannel getWritableByteChannel(OpenOption... options) throws IOException {
+            return Files.newByteChannel(getPath(), options);
+        }
     }
 
     /**
@@ -474,7 +763,7 @@ public abstract class AbstractOrigin<T, B extends AbstractOrigin<T, B>> extends 
         @Override
         public byte[] getByteArray() throws IOException {
             // TODO Pass in a Charset? Consider if call sites actually need this.
-            return IOUtils.toByteArray(origin, Charset.defaultCharset());
+            return IOUtils.toByteArray(getReader(null), Charset.defaultCharset());
         }
 
         /**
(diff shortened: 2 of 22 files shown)


src/main/java/org/apache/commons/io/channels/ByteArrayChannel.java
```diff
@@ -0,0 +1,241 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
+import org.apache.commons.io.IOUtils;
+
+/**
+ * An in-memory {@link SeekableByteChannel} backed by a growable {@code byte[]} buffer.
+ *
+ * @since 2.21.0
+ */
+public class ByteArrayChannel implements SeekableByteChannel {
```
garydgregory: Hello @ppkarwasz 
This appears to duplicate the new ByteArraySeekableByteChannel.

src/test/java/org/apache/commons/io/build/ByteArrayOriginTest.java
```diff
@@ -82,11 +75,24 @@ void testGetRandomAccessFile(final OpenOption openOption) {
         assertThrows(UnsupportedOperationException.class, () -> super.testGetRandomAccessFile(openOption));
     }
 
+    @Override
+    @Test
+    void testGetOutputStream() {
+        // Cannot convert a byte[] to an OutputStream.
+        assertThrows(UnsupportedOperationException.class, super::testGetOutputStream);
```
garydgregory:
```java
// Cannot convert a byte[] to an OutputStream.
```
What can't a ByteArrayOutputStream be used?

src/changes/changes.xml
```diff
@@ -62,10 +62,12 @@ The <action> type attribute can be add,update,fix,remove.
...
-      <action dev="pkarwasz" type="add"                due-to="Piotr P. Karwasz">Add org.apache.commons.io.file.PathUtils.getPath(String, String).</action>
+      <action dev="ggregory" type="add"                due-to="Gary Gregory">Add org.apache.commons.io.file.PathUtils.getPath(String, String).</action>
       <action dev="ggregory" type="add"                due-to="Gary Gregory">Add org.apache.commons.io.channels.ByteArraySeekableByteChannel.</action>
+      <action dev="pkarwasz" type="add"                due-to="Piotr P. Karwasz">Add ByteArrayChannel, a simple in-memory `SeekableByteChannel` implementation.</action>
+      <action dev="pkarwasz" type="add"                due-to="Piotr P. Karwasz">Add NIO channel support to `AbstractStreamBuilder`.</action>
       <!-- UPDATE -->
-      <action type="update" dev="ggregory"             due-to="Gary Gregory, Dependabot">Bump org.apache.commons:commons-parent from 85 to 88 #774, #783.</action>
+      <action type="update" dev="ggregory"             due-to="Gary Gregory, Dependabot">Bump org.apache.commons:commons-parent from 85 to 87 #774.</action>
```
garydgregory: This change is backward: We are currently on commons-parent 88.

src/main/java/org/apache/commons/io/channels/ByteArrayChannel.java
ppkarwasz: I merged your version and mine in https://github.com/apache/commons-io/pull/784/commits/c7bfc8f978abc044e44da6cc7fa310eba01f60ef

src/test/java/org/apache/commons/io/build/ByteArrayOriginTest.java
ppkarwasz: A plain `byte[]` is not a reliable backing store for an `OutputStream`.

* With `ByteArrayOutputStream` (and similar classes), the underlying array is replaced whenever it needs to grow and the original byte array will not longer receive any updates after that.
* If we were to expose an `OutputStream` directly over a fixed `byte[]`, we’d have to either throw once the array’s capacity is exceeded or silently ignore extra writes. Neither option seems broadly useful.

That’s why `ByteArrayOrigin#getOutputStream` is expected to throw.

garydgregory:
> Hi @garydgregory,
> 
> I noticed you just added [`ArraySeekableByteChannel`](https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java) from Commons Compress into IO. I was considering doing the same, but ended up writing a fresh implementation because the original class doesn’t fully meet the NIO channel concurrency contract:
> 
> * [`Channel#close()`](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/Channel.html#close--) calls must be serialized, and the closed state must be immediately visible to other threads. This is correctly handled in `ByteArraySeekableByteChannel`.
> * [`ReadableByteChannel#read(...)`](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/ReadableByteChannel.html#read-java.nio.ByteBuffer-) invocations must be serialized: BASBC does not enforce this.
> * [`WritableByteChannel#write(...)`](https://docs.oracle.com/javase/8/docs/api/java/nio/channels/WritableByteChannel.html#write-java.nio.ByteBuffer-) invocations must also be serialized: BASBC doesn’t enforce this either.

Hi @ppkarwasz 

Yes, you're right of course 😄 
Please see git master for the classic reentrant lock approach.

Would you mind providing any additional modifications to `ByteArraySeekableByteChannel` either here or separately?

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
```diff
@@ -37,41 +37,67 @@
...
+     * @return A new channel that wraps the given byte array; never {@code null}.
+     * @throws NullPointerException If the byte array is {@code null}.
+     */
+    public static ByteArraySeekableByteChannel wrap(byte[] bytes) {
+        Objects.requireNonNull(bytes, "bytes");
+        return new ByteArraySeekableByteChannel(bytes, bytes.length);
+    }
+
+    // package-private for testing
+    byte[] data;
```
garydgregory: Please leave as private and use the `array()` method.

src/main/java/org/apache/commons/io/build/AbstractOriginSupplier.java
```diff
@@ -368,4 +376,26 @@ public B setURI(final URI origin) {
...
+    }
+
+    /**
+     * Sets a new origin.
+     *
+     * @param origin the new origin.
+     * @return {@code this} instance.
+     * @since 2.21.0
+     */
+    public B setWritableByteChannel(final WritableByteChannel origin) {
```
garydgregory: If `newChannelOrigin(Channel)` and `ChannelOrigin.ChannelOrigin(Channel)` both take a Channel, why not simply have `setChannel(Channel)`?

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
```diff
@@ -137,6 +372,16 @@ public Writer getWriter(final Charset charset, final OpenOption... options) thro
             return new OutputStreamWriter(getOutputStream(options), Charsets.toCharset(charset));
         }
 
+        @Override
+        public ReadableByteChannel getReadableByteChannel(OpenOption... options) throws IOException {
+            return getRandomAccessFile(options).getChannel();
```
garydgregory: If `RandomAccessFile.getChannel()` always returns a `FileChannel` and `FileChannel` implements
`SeekableByteChannel`, `GatheringByteChannel`, `ScatteringByteChannel`, why not just have a `getFileChannel()`? I guess that depends on the typing of the overriden method. Or maybe both of these should call a `getFileChannel()`.

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
```diff
@@ -227,7 +476,7 @@ public CharSequence getCharSequence(final Charset charset) {
         @Override
         public InputStream getInputStream(final OpenOption... options) throws IOException {
             // TODO Pass in a Charset? Consider if call sites actually need this.
-            return CharSequenceInputStream.builder().setCharSequence(getCharSequence(Charset.defaultCharset())).get();
+            return CharSequenceInputStream.builder().setCharSequence(getCharSequence(null)).get();
```
garydgregory: How is this change related to Channels? This is a giant PR already. Can it be focused on just the new feature? Or is this a bug fix?

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
```diff
@@ -203,7 +452,7 @@ public CharSequenceOrigin(final CharSequence origin) {
         @Override
         public byte[] getByteArray() {
             // TODO Pass in a Charset? Consider if call sites actually need this.
-            return origin.toString().getBytes(Charset.defaultCharset());
+            return getCharSequence(null).toString().getBytes(Charset.defaultCharset());
```
garydgregory: How is this change related to Channels? This is a giant PR already. Can it be focused on just the new feature? Or is this a bug fix?

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
```diff
@@ -267,7 +521,7 @@ public FileOrigin(final File origin) {
 
         @Override
         public byte[] getByteArray(final long position, final int length) throws IOException {
-            try (RandomAccessFile raf = RandomAccessFileMode.READ_ONLY.create(origin)) {
+            try (RandomAccessFile raf = RandomAccessFileMode.READ_ONLY.create(getFile())) {
```
garydgregory: How is this change related to Channels? This is a giant PR already. Can it be focused on just the new feature? Or is this a bug fix?

garydgregory (commented):
Hi @ppkarwasz 

I am expecting a `getSeekableByteChannel()` method instead of or in addtion to the getRead/Write methods, because Compress only works in a `SeekableByteChannel` in more than one place. Did I miss it?

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/784/commits/0c853487783a43abb85de3fcc204ecca3a5ca421

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
ppkarwasz: Sorry, fixed in https://github.com/apache/commons-io/pull/784/commits/24fff74dc1645e7b566fb167befe1ad167c1ba62

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/784/commits/24fff74dc1645e7b566fb167befe1ad167c1ba62

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/784/commits/24fff74dc1645e7b566fb167befe1ad167c1ba62

src/main/java/org/apache/commons/io/build/AbstractOriginSupplier.java
ppkarwasz: I considered a single `setChannel(Channel)` as well, but decided on separate `setReadableByteChannel` and `setWritableByteChannel` methods for type safety.

While any `Channel` can be wrapped in a `ChannelOrigin`, the actual use cases differ: you need a `ReadableByteChannel` to expose an `InputStream`, and a `WritableByteChannel` to expose an `OutputStream`. Keeping the methods separate lets the compiler catch mismatches instead of deferring to runtime.

On the other hand, many channel implementations declare both interfaces but only fully support one. In those cases, errors surface **only** at runtime. Do you think it would be clearer to collapse this into a single `setChannel` method instead?

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
ppkarwasz: Good point! :100:

Should we introduce an `AbstractFileBasedOrigin` that introduces a `getFileChannel()` method? Some origins support that (`RandomAccessFile`, `File`, `Path` and possibly `FileInputStream/FileOutputStream`) but others don't.

ppkarwasz:
Hi @garydgregory,

I originally had a `getSeekableByteChannel()` in this PR but removed it because it felt misleading. For example, you can construct a `SeekableByteChannel` from a `FileOutputStream`, but that is obviously *not* the kind of channel Compress expects.

My current approach is to call `getReadableByteChannel()` in Compress and, if necessary, check `instanceof SeekableByteChannel`. This works in practice: see the test that covers 9 different ways of configuring a `SeekableByteChannel`:

https://github.com/apache/commons-io/blob/0c853487783a43abb85de3fcc204ecca3a5ca421/src/test/java/org/apache/commons/io/build/AbstractStreamBuilderTest.java#L119-L133

BTW, I think that `AbstractOrigin` is a bit muddled right now:

* It mixes sources and sinks. We could clean this up by splitting into two interfaces (e.g. `InputOrigin` and `OutputOrigin`), or by adding `isReadable()` / `isWritable()` markers.
* Resource ownership semantics are unclear. Some origins wrap pre-opened resources (`InputStream`, `RandomAccessFile`, `Channel`), others create new resources on demand (`byte[]`, `File`, `Path`, `URI`). Right now it’s ambiguous who “owns” the underlying resource. In practice, the builder should not take ownership unless explicitly documented; otherwise returned resources should be close-shielded.
* In Compress, builders are typically used to create an `ArchiveInputStream`, which *does* close the underlying resource when closed. That behavior should probably be made explicit in the Javadocs for `getInputStream()` and `getReadableByteChannel()`.
* The semantics of `getByteArray()` are also fuzzy: it consumes the origin, but should that also imply closing it?

src/changes/changes.xml
ppkarwasz: This was some merge issue. I fixed it in https://github.com/apache/commons-io/pull/784/commits/32cb23976910a368985ce02025c14378d344651f

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
```diff
@@ -37,41 +37,66 @@
...
+    }
+
+    private static byte[] byteArray(int value) {
+        if (value < 0) {
+            throw new IllegalArgumentException("Size must be non-negative");
+        }
+        return new byte[value];
+    }
+
+    private ByteArraySeekableByteChannel(byte[] data, int size) {
```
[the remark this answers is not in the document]
ppkarwasz: The lack of validation is the reason the constructor is private…

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
```diff
@@ -212,13 +249,12 @@ public int write(final ByteBuffer b) throws IOException {
         checkOpen();
         lock.lock();
         try {
-            int wanted = b.remaining();
-            final int possibleWithoutResize = size - position;
+            final int wanted = b.remaining();
+            final int possibleWithoutResize = Math.max(0, size - position);
```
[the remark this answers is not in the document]
ppkarwasz: The `SeekableByteChannel` contract explicitly allows `position` to be larger than `size`.

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
```diff
@@ -324,6 +589,18 @@ public Reader getReader(final Charset charset) throws IOException {
...
+            return Channels.newChannel(getInputStream(options));
+        }
+
+        public long size() throws IOException {
+            if (origin instanceof FileInputStream) {
+                final FileInputStream fileInputStream = (FileInputStream) origin;
+                return fileInputStream.getChannel().size();
+            }
+            throw unsupportedOperation("size");
+        }
```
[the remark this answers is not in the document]
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/784/commits/a2dd032ef042b0fa35bf68ba49cd66636bf2d3bd

src/test/java/org/apache/commons/io/build/ByteArrayOriginTest.java
garydgregory: Good analysis, ty. Since this use case has not come up, I think we should be good for now.

src/main/java/org/apache/commons/io/build/AbstractOriginSupplier.java
garydgregory: Well, thinking of the Compress use case, the only interface that matters there is `SeekableByteChannel` which is both a read and write channel as you pointed out.

>On the other hand, many channel implementations declare both interfaces but only fully support one.

I don't think we should be concerned about implementations since we are supporting interfaces here. This feels a bit like `Path` which is an abstract class but behaves like an interface for call sites. We don't concern ourselves with the implementations of `Path`.

I would expect specific origins to test for read/write/seekable kinds of channels and use what they need.

So all of this to say it seems like you should be able to say `setChannel(Channel)` and getters like `get ReadableByteChannel()`, `getSeekableByteChannel()`, and `getSeekableByteChannel()` should do their best or throw.

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
garydgregory: I had considered such a hierarchy but initially rejected it for the simplicity of a "flat"  hierarchy, which is no longer true since the introduction of `AbstractRandomAccessFileOrigin`. It might be time to revisit this.

garydgregory:
All good topics, some comments below.

> * It mixes sources and sinks. We could clean this up by splitting into two interfaces (e.g. `InputOrigin` and `OutputOrigin`), or by adding `isReadable()` / `isWritable()` markers.

Hm, that would only work for a subset of origins. Would you leave origins like `Path`, `File`, and `SeekableByteChannel` under `AbstractOrigin`, or, create an `AbstractIoOrigin`? (I don't think we'd want to test a given file's attributes since these can change on the fly.) 

I can see that you could move `InputStreamOrigin` and `ReaderOrigin` under an `AbstractInputOrigin`. Would this really buy us anything?

> Resource ownership semantics are unclear. Some origins wrap pre-opened resources (InputStream, RandomAccessFile, Channel), others create new resources on demand (byte[], File, Path, URI). Right now it’s ambiguous who “owns” the underlying resource. In practice, the builder should not take ownership unless explicitly documented; otherwise returned resources should be close-shielded.

I thought this was all clear to: The allocator owns the resources and must close it.

Are there any bugs?

>  wrap pre-opened resources (InputStream, RandomAccessFile, Channel),

This falls under "the allocating call site owns the resource."

> others create new resources on demand (byte[], File, Path, URI).

There is nothing to close on these types, they are not "resources" in the traditional "implements AutoCloseable" sense. What am I missing?

> The semantics of getByteArray() are also fuzzy: it consumes the origin, but should that also imply closing it?

This falls under "the allocating call site owns the resource". This should be covered by existing unit tests. Do you see some missing or incorrect cases?

All good topics 😄

garydgregory:
> @garydgregory
> 
> Just a heads-up: in [474f5e8](https://github.com/apache/commons-io/commit/474f5e8cc80bf8557d87d48ef512b5a983832ccd) I refactored the `ByteArraySeekableByteChannel` constructors and factory methods to align more closely with the `ByteArrayInputStream`/`ByteArrayOutputStream` patterns:
> 
> * Default buffer size is now 32, matching `ByteArrayOutputStream`.
> * Javadoc now cross-references the corresponding BAIS/BAOS constructors.
> 
> I’m not strongly against exposing a public `ByteArraySeekableByteChannel(byte[])` constructor, though I personally prefer the `wrap` factory method.

Most if not all buffers in Commons IO are `IOUtils.DEFAULT_BUFFER_SIZE` (8K), attempting to align with OS file block sizes.

src/main/java/org/apache/commons/io/build/AbstractOriginSupplier.java
ppkarwasz: You make a good point. Since there are quite a few `Channel` sub-interfaces, how about introducing a single generic accessor on `AbstractOrigin` instead of separate getters? For example:

```java
public <T extends Channel> T getChannel(Class<T> type);
```

Each implementation could then attempt to provide the requested type or fail fast if it can’t.

On the builder side we can keep it simple with a single setter:

```java
public B setChannel(Channel channel);
```

ppkarwasz: In https://github.com/apache/commons-io/pull/784/commits/c4e3b4300f2221ccfbcb14adf5184c7be5154b8d I replaced the `getReadableByteChannel` and similar methods with those proposed above: the API looks much cleaner now.

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
ppkarwasz: In https://github.com/apache/commons-io/pull/784/commits/c4e3b4300f2221ccfbcb14adf5184c7be5154b8d I introduced `getChannel`, that can be used as:

```java
origin.getChannel(FileChannel.class)
```

garydgregory:
> ```java
> InputStreamOrigin
> ```

This is definitely not a bug and is the expected behavior. A couple of points:
- Maybe the type-specific origin classes like `InputStreamOrigin` shouldn't have been public, but it would be harder to provide new conversions outside the library itself, but maybe IO isn't used at that level of depth. A lesson learned for the future.
- The type-specific origin classes (here, `InputStreamOriogin`) are all _wrappers_ of that specific type, and always return the instance it was given on creation from its same-type getter (here, `getInputStream()`). In this case, an `InputStreamOrigin` is constructed with an `InputStream` and always returns that `InputStream` from `getInputStream()`
- The previous point highlights the need for better Javadocs I think.

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
```diff
@@ -755,6 +1120,28 @@ public Writer getWriter(final Charset charset, final OpenOption... options) thro
...
+    public <C extends Channel> C getChannel(Class<C> channelType, OpenOption... options) throws IOException {
+        Objects.requireNonNull(channelType, "channelType");
+        final Channel channel = doGetChannel(options);
+        if (channelType.isInstance(channel)) {
+            return channelType.cast(channel);
+        }
+        throw unsupportedChannelType(channelType);
+    }
+
+    abstract Channel doGetChannel(OpenOption... options) throws IOException;
```
garydgregory: What is this supposed to do? A Javadoc would help 😉 

I'm _really_ not a fan of the `doFoo` naming convention. To me it screams: "I couldn't think of a better name!". It looks like all of these implementations create a _new_ channel (except `ChannelOrigin` obviously), so... how about  calling it`newChannel()`? At least that name is pretty clear, even without a Javadoc 😉

src/main/java/org/apache/commons/io/build/AbstractStreamBuilder.java
```diff
@@ -258,6 +260,23 @@ public Writer getWriter() throws IOException {
...
+     * @return A channel of the specified type.
+     * @param <C>         The channel type.
+     * @throws IllegalStateException         if the {@code origin} is {@code null}.
+     * @throws UnsupportedOperationException if the origin cannot be converted to a {@link ReadableByteChannel}.
+     * @throws IOException                   if an I/O error occurs.
+     * @see AbstractOrigin#getChannel
+     * @see #getOpenOptions()
+     * @since 2.21.0
+     */
+    public <C extends Channel> C getChannel(Class<C> channelType) throws IOException {
```
garydgregory, line 276: Yeah, that's likely a good general solution 👍

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
```diff
@@ -37,41 +37,74 @@
  * accessed via {@link ByteArraySeekableByteChannel#array()}.
  * </p>
  *
- * @since 2.19.0
+ * @since 2.21.0
```
garydgregory, line 42: Good catch! 👍

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
```diff
@@ -37,41 +37,74 @@
...
     /**
-     * Constructs a new instance using a default empty buffer.
+     * Constructs a new instance, with a default internal buffer capacity.
+     * <p>
+     * The initial size and position of the channel are 0.
+     * </p>
+     * <p>
+     * The default initial capacity is 32 bytes, although the capacity will grow as needed when writing data.
+     * </p>
+     * @see java.io.ByteArrayOutputStream#ByteArrayOutputStream()
```
garydgregory: There should be a blank Javadoc line before the first `@` tag.

Don't use FQCN when you don't need to in `@see java.io...`

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
```diff
@@ -37,41 +37,74 @@
...
      */
     public ByteArraySeekableByteChannel() {
-        this(0);
+        this(32);
```
garydgregory: Hm, why not `IOUtils.DEFAULT_BUFFER_SIZE` (8 KiB)? Confusingly we use `1024` in our own `AbstractByteArrayOutputStream` and `ByteArrayOutputStream`.

I'm surprised this change doesn't break anything, because if you only make that change, `ByteArraySeekableByteChannelTest` breaks.
When I tried this kinf of change here and in Compress' `SeekableInMemoryByteChannel`, the tests broke when I changed the default size to `IOUtils.DEFAULT_BUFFER_SIZE`. But now it works in `SeekableInMemoryByteChannel`...

src/test/java/org/apache/commons/io/channels/ByteArraySeekableByteChannelCompressTest.java
```diff
@@ -62,7 +66,7 @@ void testCloseIsIdempotent() throws Exception {
     @ParameterizedTest
     @ValueSource(ints = { 0, 1, 2, 3, 4, 5, 6 })
     void testReadingFromAPositionAfterEndReturnsEOF(final int size) throws Exception {
-        try (SeekableByteChannel c = new ByteArraySeekableByteChannel(size)) {
+        try (SeekableByteChannel c = ByteArraySeekableByteChannel.wrap(new byte[size])) {
```
garydgregory, line 72: I don't think this is a good idea for reading tests. You want to test reading from the same array over and over when you are reading because you want to make sure there are no side-effects that incorrectly _write_ to the underlying byte array.

src/test/java/org/apache/commons/io/channels/ByteArraySeekableByteChannelCompressTest.java
```diff
@@ -42,6 +42,10 @@ class ByteArraySeekableByteChannelCompressTest {
 
     private final byte[] testData = "Some data".getBytes(StandardCharsets.UTF_8);
 
+    private byte[] getTestData() {
```
garydgregory: This would be better named `cloneTestData` IMO to make clear to tests what kind of data we are dealing with.

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
```diff
@@ -37,41 +37,74 @@
...
+     * @see java.io.ByteArrayInputStream#ByteArrayInputStream(byte[])
+     */
+    public static ByteArraySeekableByteChannel wrap(byte[] bytes) {
+        Objects.requireNonNull(bytes, "bytes");
+        return new ByteArraySeekableByteChannel(bytes);
+    }
+
     private byte[] data;
-    private final AtomicBoolean closed = new AtomicBoolean();
+    private volatile boolean closed;
```
garydgregory, line 70: This seems unrelated. Why is this better? Just because it's simpler? Just curious.

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
```diff
@@ -37,41 +37,74 @@
...
     /**
-     * Constructs a new instance from a byte array.
-     *
-     * @param data input data or pre-allocated array.
+     * Constructs a new instance, with an internal buffer of the given capacity, in bytes.
+     * <p>
+     * The initial size and position of the channel are 0.
+     * </p>
+     * @param size Capacity of the internal buffer to allocate, in bytes.
+     * @see java.io.ByteArrayOutputStream#ByteArrayOutputStream(int)
```
garydgregory: Don't use FQCN when you don't need to in `@see java.io...`

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
```diff
@@ -53,15 +58,246 @@
...
  *
- * @param <T> the type of instances to build.
- * @param <B> the type of builder subclass.
+ * <p>
+ * The table below summarizes which views and conversions are supported for each origin type.
+ * Column headers show the target view; cells indicate whether that view is available from the origin in that row.
+ * </p>
+ *
+ * <table>
+ *   <caption>Origin support matrix</caption>
```
garydgregory, line 75: Nice! I see entries for `WritableByteChannel`, but shouldn't there be entries for `ReadableByteChannel`, and `SeekableByteChannel`? 

Or should there be a single entry for `Channel`?

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
ppkarwasz: I renamed it to `getChannel` and changed the access to `protected` in https://github.com/apache/commons-io/pull/784/commits/b820c4497cab67fe757234b550634e89477e1974. This will allow users to reuse the `getChannel(Class<?>, OpenOption...)` implementation above.

In this case, should we make `getChannel(Class<?>, OpenOption...)` final?

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
ppkarwasz: I have Copilot enabled in my IDE, so I am always watching for this kind of typos: Copilot always suggests a random completion of `@since` tags.

BTW: Copilot also suggests random completions for SHA256 properties. :laughing:

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
ppkarwasz: I changed a little bit the semantics of the `ByteArraySeekableByteChannel()` and `ByteArraySeekableByteChannel(int)` constructors: they allocate a buffer of the given size, but they keep the `size()` of the `SeekableByteChannel` equal to 0.

This way, calling `toByteArray()` right after the constructor, returns an empty array. This aligns with how `ByteArrayOutputStream#toByteArray()` works. Of course calling `array()` returns the raw backing array of the given size.

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/784/commits/b2ac652df0f24614995e48fa7d6ba236149aa633

To remove the FQCN I had to add an import.

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/784/commits/b2ac652df0f24614995e48fa7d6ba236149aa633

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
ppkarwasz: I restored `DEFAULT_BUFFER_SIZE` in https://github.com/apache/commons-io/pull/784/commits/94f7f783ab0790fee3d374ce6c38e398711a1f20.

src/test/java/org/apache/commons/io/channels/ByteArraySeekableByteChannelCompressTest.java
ppkarwasz: This change is due to the different semantics of `wrap` vs the constructor: `wrap` also updates the logical size of the channel to the length of the byte array, while the constructor keeps it at 0.

src/test/java/org/apache/commons/io/channels/ByteArraySeekableByteChannelCompressTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/784/commits/42133ae14f0d227064f4c320b87ab8a68b588185

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
ppkarwasz: Yes, I replaced it because it is simpler (no additional Java object) and has the exact same semantics.

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
ppkarwasz: There should be entries for both `ReadableByteChannel` and `WritableByteChannel`. The table contains readable resources first (`InputStream`, `Reader`, RBC) and writable resources next (`OutputStream`, `Writer`, WBC).

Initially I also added `SeekableByteChannel`, but it seemed confusing.

ppkarwasz: In https://github.com/apache/commons-io/pull/784/commits/ffb3780f46763c5330d61710e9d439a18e1f6f3e I added an additional column, which summarizes what kind of `Channel` one can expect from various origins.

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
garydgregory: If you want, it's the more conservative choice. If we make it final now, we can always loosen it later, but not the other way around.

src/test/java/org/apache/commons/io/channels/ByteArraySeekableByteChannelCompressTest.java
garydgregory: Sure, but I am talking about the _all_ the tests that read from a byte array like the ones below. I should have made that clearer in my initial comment. For example, this is _not_ the pattern we want:
```diff
    @Test
    void testShouldReadContentsProperly() throws IOException {
-        try (ByteArraySeekableByteChannel c = new ByteArraySeekableByteChannel(testData)) {
+        try (ByteArraySeekableByteChannel c = ByteArraySeekableByteChannel.wrap(cloneTestData())) {
```
A read test should not affect the underlying byte array so you want to give it the raw array.
If we want to give it a copy, then maybe the PR should also make sure the copy has not been altered?

src/main/java/org/apache/commons/io/channels/ByteArraySeekableByteChannel.java
garydgregory: If you "Resolve a Conversation" in GH after answering a question, the person asking the question doesn't see the answer! It's hidden behind a "Show resolved" button. I think it makes sense for the person asking the question to do the "Resolve". Now I have to open every "Show resolved" to see if a question was truly resolved or not.

src/test/java/org/apache/commons/io/build/CharSequenceOriginTest.java
```diff
@@ -62,13 +62,6 @@ void testGetFile() {
         assertThrows(UnsupportedOperationException.class, super::testGetFile);
     }
 
-    @Override
-    @Test
-    void testGetOutputStream() {
```
garydgregory, line 67: Why was this method deleted and then added again? If you keep the methods in AB order, this shouln't happen.

src/test/java/org/apache/commons/io/build/CharSequenceOriginTest.java
```diff
@@ -109,11 +123,4 @@ void testGetReaderIgnoreCharsetNull() throws IOException {
         }
     }
 
-    @Override
```
garydgregory, line 112: It's confusing to see a method deleted and added again.

src/test/java/org/apache/commons/io/build/InputStreamOriginTest.java
```diff
@@ -58,13 +59,6 @@ void testGetFile() {
         assertThrows(UnsupportedOperationException.class, super::testGetFile);
     }
 
-    @Override
```
garydgregory, line 61: This method doesn't need to be moved; it makes the PR larger for no reason.

src/test/java/org/apache/commons/io/build/OutputStreamOriginTest.java
```diff
@@ -89,13 +90,6 @@ void testGetFile() {
         assertThrows(UnsupportedOperationException.class, super::testGetFile);
     }
 
-    @Override
```
garydgregory, line 92: This method doesn't need to be moved; it makes the PR larger for no reason.

src/test/java/org/apache/commons/io/build/ReaderOriginTest.java
```diff
@@ -56,13 +57,6 @@ void testGetFile() {
         assertThrows(UnsupportedOperationException.class, super::testGetFile);
     }
 
-    @Override
```
garydgregory, line 59: This method doesn't need to be moved; it makes the PR larger for no reason.

src/test/java/org/apache/commons/io/build/WriterStreamOriginTest.java
```diff
@@ -89,13 +90,6 @@ void testGetFile() {
         assertThrows(UnsupportedOperationException.class, super::testGetFile);
     }
 
-    @Override
```
garydgregory, line 92: This method doesn't need to be moved; it makes the PR larger for no reason.

src/main/java/org/apache/commons/io/build/AbstractOrigin.java
ppkarwasz: Done in https://github.com/apache/commons-io/pull/784/commits/c75480d9e703cc8d45228b3f4e43c5dd16fc6202

src/test/java/org/apache/commons/io/channels/ByteArraySeekableByteChannelCompressTest.java
ppkarwasz: You are right: all but one test were not supposed to modify any data.

Fixed in https://github.com/apache/commons-io/pull/784/commits/f24ab1d326d748c4eb84e3e59249354477ddabd9

src/test/java/org/apache/commons/io/build/CharSequenceOriginTest.java
ppkarwasz: This helped me check whether each `✖` in the `AbstractOrigin` support matrix has a corresponding test, but you are right, I shouldn't have changed the order of the existing methods.

Note that keeping alphabetical order is also not ideal as it separates related methods: I usually put the `static` method that generates parameters for a parameterized test near the test method. Reordering methods with Eclipse breaks them apart.

src/test/java/org/apache/commons/io/build/CharSequenceOriginTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/784/commits/46235ff46e47dcb5fbf79b9c51ce64f193956e39

src/test/java/org/apache/commons/io/build/InputStreamOriginTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/784/commits/46235ff46e47dcb5fbf79b9c51ce64f193956e39

src/test/java/org/apache/commons/io/build/OutputStreamOriginTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/784/commits/46235ff46e47dcb5fbf79b9c51ce64f193956e39

src/test/java/org/apache/commons/io/build/ReaderOriginTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/784/commits/46235ff46e47dcb5fbf79b9c51ce64f193956e39

src/test/java/org/apache/commons/io/build/WriterStreamOriginTest.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/784/commits/46235ff46e47dcb5fbf79b9c51ce64f193956e39

ppkarwasz:
@garydgregory, done!

FYI: the `feat:` in my titles is a reference to [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/). I use these prefixes often in my commits, the message of the first commit is usually used to fill the PR and after squashing, it becomes the title of a commit on `master`.

PR #786, opened 2025-09-22
feat: Add `Channels.closeShield` for close-shield proxies of NIO channels

ppkarwasz:
Adds `Channels.closeShield(Channel)` to return a JDK proxy that preserves the delegate’s `Channel` sub-interfaces and shields the underlying channel from `close()`.

**Behavior**

* `close()` flips shield state; delegate is not closed.
* `isOpen()` reflects shield state and delegates if shield is not closed.
* After shield-close, I/O/mutating methods throw `ClosedChannelException`; safe queries (e.g., `NetworkChannel.supportedOptions()`) still work.
* Fluent methods that return `this` (e.g., `SeekableByteChannel.position/truncate`, `NetworkChannel.bind/setOption`) return the **proxy**.
* Stable `equals`/`hashCode`/`toString`
* The method is idempotent (no double-wrapping).

**Tests**

* All methods of `SeekableByteChannel` / `NetworkChannel` and super-interfaces are tested.

### DIFF
--- a/src/main/java/org/apache/commons/io/channels/Channels.java
+++ b/src/main/java/org/apache/commons/io/channels/Channels.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/commons/io/channels/Channels.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.LinkedHashSet;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * Channel utilities.
+ *
+ * @since 2.21.0
+ */
+public final class Channels {
```
garydgregory: Calling this class `Channels` is going to make using `java.nio.channels.Channels` and `jdk.nio.Channels` awkward. How about callling it `ClosedShieldChannel` since that's all it does and can be documented as such at the class level.

ppkarwasz: Good call! :100:

I renamed the shielding method to `CloseShieldChannel.wrap` in https://github.com/apache/commons-io/pull/786/commits/44cd3a20eea718045e9ab9fb668af5e44c5f7302.

garydgregory:
@ppkarwasz TY. PR merged!

PR #789, opened 2025-09-28
Add `Closeable` to `AbstractOrigin`

ppkarwasz:
This change makes `AbstractOrigin` implement `Closeable` and updates the Javadoc of all “getter” methods that return `Closeable` resources. The Javadoc now explicitly warns that:

* Closing the returned resource will also close the underlying `AbstractOrigin`.
* Ownership of the underlying resource is transferred to the caller.

### Rationale

* Callers must always close the returned resource. Some implementations create a new resource on each call, which must be released explicitly.
* Other implementations return the underlying origin directly, making it impossible to guarantee a close-shielded wrapper.
* The primary usage of `AbstractOrigin` is within `AbstractStreamBuilder`, whose implementations typically wrap closeable resources.

- [x] I used AI to generate the Javadoc.

### DIFF
--- a/src/test/java/org/apache/commons/io/build/AbstractOriginTest.java
+++ b/src/test/java/org/apache/commons/io/build/AbstractOriginTest.java
@@ -23,7 +23,6 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -46,8 +45,6 @@ import java.util.Objects;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.io.build.AbstractOrigin.RandomAccessFileOrigin;
-import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 import org.junit.jupiter.params.ParameterizedTest;
@@ -66,17 +63,16 @@ public abstract class AbstractOriginTest<T, B extends AbstractOrigin<T, B>> {
     protected static final String FILE_NAME_RW = AbstractOriginTest.class.getSimpleName() + ".txt";
     private static final int RO_LENGTH = 20;
 
-    protected AbstractOrigin<T, B> originRo;
-    protected AbstractOrigin<T, B> originRw;
-
     @TempDir
     protected Path tempPath;
 
-    @BeforeEach
-    void beforeEach() throws IOException {
-        setOriginRo(newOriginRo());
-        resetOriginRw();
-        setOriginRw(newOriginRw());
+    /**
+     * Asserts that the origin is open.
+     *
+     * @param origin The origin to test.
+     */
+    protected void assertOpen(final AbstractOrigin<T, B> origin) {
+        // No-op
     }
 
     private void checkRead(final ReadableByteChannel channel) throws IOException {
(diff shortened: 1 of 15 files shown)


garydgregory (changes requested):
This PR seems wrong to me in many ways; I'm not sure where to begin 😞 The good thing is that it's forcing me to think through it and explain the use cases, then maybe the need for changes will surface.

The reason the root origin class is called `AbstractOrigin` and not `AbstractCloseable` is to make it generic. This PR flips the design inside-out or upside down; I'm not sure how to even describe it.

The fact that so many tests had to change is also a red flag to me.

Let's remember that the end user for this code is a Builder. There are two use cases related to resource management for a Builder:
- A client allocates a `Closeable` (or `AutoCloseable`) resource, creates a Builder, and gives the Builder that resource by calling a setter method. No matter what happens next, the client is responsible for releasing the resource (`Closeable.close()`).  In this case, the origin wraps but doesn't own the closeable resource. There is no "transfer of ownership". 
- A client creates a Builder and gives it a non-`Closeable` object, like a `File` or a `Path`. The client then calls the Builder's factory method (`get()`), and that call returns a `Closeable` or a resource that requires releasing in some other way. No matter what happens next, the client is responsible for releasing that resource. In this case, the origin doesn't wrap a closeable resource.

In both cases, the client causes the allocation and is responsible for releasing the resource. `Closeable` doesn't belong on `AbstractOrigin` any more than it belongs on a Builder. 

Notice that our builders never implement `Closeable`. There is no "transfer of ownership".

----
The implementation of close() is also an OO red flag to me:
```java
        if (origin instanceof Closeable) {
            ((Closeable) origin).close();
        }
```
The above tells me the hierarchy is wrong and that I'd expect to see a `CloseableOrigin` and subclassses. Maybe we could have a `CloseableOrigin` or an `AbstractCloseableOrigin`, I'm not sure. What I did do is keep the `AbstractOrigin` hierarchy as flat as possible to avoid additional abstractions that were not strictly necessary.

garydgregory:
> That’s useful, but since users generally don’t code against concrete subclasses, the Javadoc on AbstractOrigin#getInputStream() and similar methods should also be improved to define rules that apply consistently across all implementations.

I think the table in the Javadoc for the class `AbstractOrigin` now documents all the conversion rules quite nicely.

Do you think each row should be copied to its respective `AbstractOrigin` subclass?

I added a resource management paragraph at the class level. I think the table does a really nice job of documenting the whole thing.

ppkarwasz:
Hi @garydgregory,

> I added a resource management paragraph at the class level. I think the table does a really nice job of documenting the whole thing.

I am not sure which “resource management paragraph” are you referring to. Is it the first paragraph?

https://github.com/apache/commons-io/blob/2207f9a8043ea35d58186aacb79b2f2520372238/src/main/java/org/apache/commons/io/build/AbstractOrigin.java#L62-L64

I am not entirely sure that this conveys the entire semantics of `AbstractOrigin`, which is:

- If the origin is a resource, getters will return a wrapped resource.
- If the origin is not a resource, getters will return a new resource, that the caller must close.

garydgregory:
> I am not entirely sure that this conveys the entire semantics of AbstractOrigin

I forgot to push, it's there now. Sorry about that.

src/test/java/org/apache/commons/io/build/AbstractOriginTest.java
```diff
@@ -133,50 +117,57 @@ private boolean isValid(final RandomAccessFile raf) throws IOException {
...
+        close(originRo);
     }
 
     @Test
     void testGetByteArrayAt_0_0() throws IOException {
-        assertArrayEquals(new byte[] {}, getOriginRo().getByteArray(0, 0));
+        final AbstractOrigin<T, B> originRo = newOriginRo();
+        assertArrayEquals(new byte[]{}, originRo.getByteArray(0, 0));
+        assertOpen(originRo);
+        close(originRo);
```
garydgregory, line 133: Hello @ppkarwasz 

The benefit of the previous approach is that the `@AfterEach cleanup()` method was _always_ called, no matter what, whether a test passed or failed. In this PR, the release of an origin is no longer guaranteed since the `@AfterEach` is gone and the new calls to `close(...)` are _not_ in a `finally` block. The origin is never released when an assertion fails. 

Am I missing something?

If all the tests now contain a `try-finally` block, it says to me that an `@AfterEach` method is warranted. to avoid copy-pasta all over and increasing the clutter.

WDYT? Is there a way to have better tests, but with release guarantees that don't clutter up the tests?

garydgregory (changes requested):
Hi @ppkarwasz 

Please see my one comment which applies to multiple tests.

src/test/java/org/apache/commons/io/build/AbstractOriginTest.java
ppkarwasz: My main motivation was to prevent `toByteArray()` implementations of this kind:

```java
IOUtils.toByteArray(getInputStream());
```
This would be clearly wrong, if `getInputStream()` creates a **new** input stream (instead of a wrapper).

However, I see now that I went into the wrong direction and the proposed changes would never detect that.

What would work is to move `originRo` to a temporary path, which would cause a test failure on Windows if the test leaves any unclosed `RandomAccessFile` or `FileInputStream`, because this prevents the removal of the temporary directory. An unclosed stream opened with `Files.getInputStream()` (which we currently use) doesn't prevent the removal of the parent directory even on Windows.

I am closing this PR. Thanks for sparring with me.

PR #790, opened 2025-09-28
Backport `Objects.checkFromIndexSize` to `IOUtils`

ppkarwasz:
Java 9 introduced `Objects.checkFromIndexSize` for validating the offset/length ranges used in `read` and `write` operations.

This PR adds an equivalent method, `IOUtils.checkFromIndexSize`, to provide the same validation in Commons IO. It simplifies range checks in stream implementations and makes the utility available to projects that still target Java 8.

### DIFF
--- a/src/main/java/org/apache/commons/io/IOUtils.java
+++ b/src/main/java/org/apache/commons/io/IOUtils.java
@@ -407,6 +407,36 @@ public class IOUtils {
     }
 
     /**
+     * Checks if the sub-range described by an offset and length is valid for an array of the given length.
+     *
+     * <p>This method is functionally equivalent to
+     * {@code java.util.Objects#checkFromIndexSize(int, int, int)} introduced in Java 9,
+     * but is provided here for use on Java 8.</p>
+     *
+     * <p>The range is valid if all of the following hold:</p>
+     * <ul>
+     *   <li>{@code off >= 0}</li>
+     *   <li>{@code len >= 0}</li>
+     *   <li>{@code arrayLength >= 0}</li>
+     *   <li>{@code off + len <= arrayLength}</li>
+     * </ul>
+     *
+     * <p>If the range is invalid, this method throws an
+     * {@link IndexOutOfBoundsException} with a descriptive message.</p>
+     *
+     * @param off         the starting offset into the array
+     * @param len         the number of elements in the range
+     * @param arrayLength the length of the array to be checked against
+     * @throws IndexOutOfBoundsException if the range {@code [off, off + len)} is out of bounds
+     * @since 2.21.0
+     */
+    public static void checkFromIndexSize(final int off, final int len, final int arrayLength) {
+        if ((off | len | arrayLength) < 0 || arrayLength - len < off) {
+            throw new IndexOutOfBoundsException(String.format("Range [%s, %<s + %s) out of bounds for length %s", off, len, arrayLength));
+        }
+    }
+
+    /**
      * Clears any state.
      * <ul>
      * <li>Removes the current thread's value for thread-local variables.</li>
--- a/src/test/java/org/apache/commons/io/IOUtilsTest.java
+++ b/src/test/java/org/apache/commons/io/IOUtilsTest.java
@@ -50,6 +50,7 @@ import java.io.Reader;
 import java.io.SequenceInputStream;
 import java.io.StringReader;
 import java.io.Writer;
+import java.lang.reflect.InvocationTargetException;
 import java.net.ServerSocket;
 import java.net.Socket;
 import java.net.URI;
@@ -65,6 +66,7 @@ import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
@@ -85,7 +87,9 @@ import org.apache.commons.io.output.StringBuilderWriter;
 import org.apache.commons.io.output.UnsynchronizedByteArrayOutputStream;
 import org.apache.commons.io.test.TestUtils;
 import org.apache.commons.io.test.ThrowOnCloseReader;
+import org.apache.commons.lang3.JavaVersion;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.SystemUtils;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
@@ -346,6 +350,54 @@ class IOUtilsTest {
     @Test
     void testByteArrayWithNegativeSize() {
         assertThrows(NegativeArraySizeException.class, () -> IOUtils.byteArray(-1));
+    }
+
+    static Stream<Arguments> testCheckFromIndexSizeValidCases() {
+        return Stream.of(
+                // Valid cases
+                Arguments.of(0, 0, 42),
+                Arguments.of(0, 1, 42),
+                Arguments.of(0, 42, 42),
+                Arguments.of(41, 1, 42),
+                Arguments.of(42, 0, 42)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource
+    void testCheckFromIndexSizeValidCases(int off, int len, int arrayLength) {
+        assertDoesNotThrow(() -> IOUtils.checkFromIndexSize(off, len, arrayLength));
+    }
+
+    static Stream<Arguments> testCheckFromIndexSizeInvalidCases() {
+        return Stream.of(
+                Arguments.of(-1, 0, 42),
+                Arguments.of(0, -1, 42),
+                Arguments.of(0, 0, -1),
+                // off + len > arrayLength
+                Arguments.of(1, 42, 42),
+                Arguments.of(Integer.MAX_VALUE, 1, Integer.MAX_VALUE)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource
+    void testCheckFromIndexSizeInvalidCases(int off, int len, int arrayLength) {
+        final IndexOutOfBoundsException ex = assertThrows(IndexOutOfBoundsException.class, () -> IOUtils.checkFromIndexSize(off, len, arrayLength));
+        assertTrue(ex.getMessage().contains(String.valueOf(off)));
+        assertTrue(ex.getMessage().contains(String.valueOf(len)));
+        assertTrue(ex.getMessage().contains(String.valueOf(arrayLength)));
+        // Optional requirement: compare the exception message for Java 8 and Java 9+
+        if (SystemUtils.isJavaVersionAtLeast(JavaVersion.JAVA_9)) {
+            final IndexOutOfBoundsException jreEx = assertThrows(IndexOutOfBoundsException.class, () -> {
+                try {
+                    Objects.class.getDeclaredMethod("checkFromIndexSize", int.class, int.class, int.class).invoke(null, off, len, arrayLength);
+                } catch (InvocationTargetException ite) {
+                    throw ite.getTargetException();
+                }
+            });
+            assertEquals(jreEx.getMessage(), ex.getMessage());
+        }
     }
 
     @Test


garydgregory:
Hi @ppkarwasz 

Shouldn't his be called from our own code base to show it's what we really need in IO? Otherwise it belongs in Lang since it hangs on Object in Java 9.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -17,6 +17,8 @@
 
 package org.apache.commons.io;
 
+import static java.util.Objects.requireNonNull;
```
garydgregory: In general, we only use static imports in unit tests for JUnit. These changes creates a lot of clutter in the PR which makes it larger than it needs to be.

ppkarwasz:
Hi @garydgregory,

> There are a lot of builds failures here.

Yes, I noticed: they all come from subtle changes in how those classes validate their arguments. I’ll take a closer look tomorrow.

For `ClosedInputStream` and `ClosedReader`, the tests need to be updated: the Javadoc specifies that when the buffer length is `0`, the method should return `0`. Only when the buffer is non-empty should it attempt to read bytes or characters.

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/cbfb55cc15c3f4344bbe7079026b0214024b11e9

src/main/java/org/apache/commons/io/input/NullInputStream.java
```diff
@@ -292,7 +294,8 @@ public int read(final byte[] bytes) throws IOException {
      */
     @Override
     public int read(final byte[] bytes, final int offset, final int length) throws IOException {
-        if (bytes.length == 0 || length == 0) {
+        IOUtils.checkFromIndexSize(bytes, offset, length);
```
garydgregory, line 300: Hi @ppkarwasz 

When I look at:
- java.io.BufferedReader.read(char[], int, int)
- java.io.BufferedInputStream.read(byte[], int, int)
- java.io.CharArrayReader.read(char[], int, int)
- java.io.PipedReader.read(char[], int, int)
- java.io.PushbackInputStream.read(byte[], int, int)
- java.io.PushbackReader.read(char[], int, int)
- ...

The check for a functionally equivalent `checkOpen()` like `ensureOpen()` happens _before_ the check for indices.

Let's follow the same pattern.

garydgregory (changes requested):
Hi @ppkarwasz 

Thank you for the PR. See my one comment, which applies to multiple call sites.

src/main/java/org/apache/commons/io/input/NullInputStream.java
ppkarwasz: I totally agree for the `Reader` implementations: all the implementations I have seen call `ensureOpen()` **before** arguments validation.

However in `InputStream` I see the inverse convention:

1. The Javadoc states that:
    > If `len` is zero, then no bytes are read and `0` is returned; otherwise, there is an attempt to read at least one byte.
   
   and
   
   > `IOException` - If the **first** byte cannot be read for any reason other than end of file, or if the input stream has been closed, or if some other I/O error occurs.

2. Implementations seem to agree that `ensureOpen()` or equivalent is checked **after** arguments validation:
    - `FileInputStream` delegates to the native method `readBytes` ([source](https://github.com/openjdk/jdk/blob/2746c1a555891564963299182b3b0293eaefc901/src/java.base/share/native/libjava/io_util.c#L74)).
    - `InputStream.nullInputStream()` ([source](https://github.com/openjdk/jdk/blob/2746c1a555891564963299182b3b0293eaefc901/src/java.base/share/classes/java/io/InputStream.java#L87)) validates **before** checking if it is open.
    - The default `InputStream#read(byte[], int, int)` implementation ([source](https://github.com/openjdk/jdk/blob/2746c1a555891564963299182b3b0293eaefc901/src/java.base/share/classes/java/io/InputStream.java#L283)) also validates before calling `read()`.
    - The same pattern can be seen in `sun.nio.ch.ChannelInputStream` and `sun.nio.ch.SocketInputStream`, which are wrappers returned by `Channels.newInputStream(ReadableByteChannel)`.
    - You are right about `BufferedInputStream` and `PushbackInputStream`: they call an equivalent of `ensureOpen()` **before** arguments validation.

src/main/java/org/apache/commons/io/input/ClosedReader.java
```diff
@@ -71,6 +71,10 @@ public void close() throws IOException {
      */
     @Override
     public int read(final char[] cbuf, final int off, final int len) {
+        IOUtils.checkFromIndexSize(cbuf, off, len);
```
garydgregory, line 81: Hello @ppkarwasz 

This is now in contradiction to the Javadoc for the class, the method, all method parameters, and the return tag.

For each call to `checkFromIndexSize`, the Javadoc might need to be adjusted or the behavior should remain the same, each use case must be reviewed 😞  Let's start with this one. What to do?

src/main/java/org/apache/commons/io/input/BrokenReader.java
```diff
@@ -114,6 +115,10 @@ public void mark(final int readAheadLimit) throws IOException {
      */
     @Override
     public int read(final char[] cbuf, final int off, final int len) throws IOException {
+        IOUtils.checkFromIndexSize(cbuf, off, len);
```
garydgregory: The class and method Javadoc, and the implementation are now divergent. The point of this class is to always throw an `IOException` (see class Javadoc), so... I'm not sure changing this implementation in such a subtle way is a good thing.

src/main/java/org/apache/commons/io/output/BrokenWriter.java
```diff
@@ -121,6 +122,7 @@ private RuntimeException rethrow() {
      */
     @Override
     public void write(final char[] cbuf, final int off, final int len) throws IOException {
+        IOUtils.checkFromIndexSize(cbuf, off, len);
```
garydgregory: Probably not? See comments for other "Broken" classes.

src/main/java/org/apache/commons/io/output/ClosedOutputStream.java
```diff
@@ -73,6 +75,7 @@ public void flush() throws IOException {
      */
     @Override
     public void write(final byte b[], final int off, final int len) throws IOException {
+        IOUtils.checkFromIndexSize(b, off, len);
```
garydgregory, line 80: Probably not? Not in agreement with Javadoc. See comments for the "Broken" classes.

src/main/java/org/apache/commons/io/output/ClosedWriter.java
```diff
@@ -77,6 +79,7 @@ public void flush() throws IOException {
      */
     @Override
     public void write(final char[] cbuf, final int off, final int len) throws IOException {
+        IOUtils.checkFromIndexSize(cbuf, off, len);
```
garydgregory: Probably not? Not in agreement with Javadoc. See comments for the "Broken" classes.

src/main/java/org/apache/commons/io/output/FilterCollectionWriter.java
```diff
@@ -128,6 +129,7 @@ public void write(final char[] cbuf) throws IOException {
     @SuppressWarnings("resource") // no allocation
     @Override
     public void write(final char[] cbuf, final int off, final int len) throws IOException {
+        IOUtils.checkFromIndexSize(cbuf, off, len);
```
garydgregory: This is not in agreement with the class Javadoc and may break some apps perhaps.

src/main/java/org/apache/commons/io/output/FilterCollectionWriter.java
```diff
@@ -159,6 +161,7 @@ public void write(final String str) throws IOException {
     @SuppressWarnings("resource") // no allocation
     @Override
     public void write(final String str, final int off, final int len) throws IOException {
+        IOUtils.checkFromIndexSize(str, off, len);
         forAllWriters(w -> w.write(str, off, len));
     }
```
garydgregory: This is not in agreement with the class Javadoc and may break some apps perhaps.

src/main/java/org/apache/commons/io/output/NullAppendable.java
```diff
@@ -51,6 +53,7 @@ public Appendable append(final CharSequence csq) throws IOException {
 
     @Override
     public Appendable append(final CharSequence csq, final int start, final int end) throws IOException {
+        IOUtils.checkFromToIndex(csq, start, end);
```
garydgregory, line 71: This contradicts the class Javadoc. It seems to me that our "Null", "Broken", and "Closed" classes are different beasts from other stream/reader/writer implementations.

src/main/java/org/apache/commons/io/output/NullOutputStream.java
```diff
@@ -75,7 +77,7 @@ public void write(final byte[] b) throws IOException {
      */
     @Override
     public void write(final byte[] b, final int off, final int len) {
-        // noop
+        IOUtils.checkFromIndexSize(b, off, len);
```
garydgregory, line 85: This contradicts the class Javadoc. It seems to me that our "Null", "Broken", and "Closed" classes are different beasts from other stream/reader/writer implementations.

src/main/java/org/apache/commons/io/output/NullWriter.java
```diff
@@ -110,6 +112,7 @@ public void flush() {
      */
     @Override
     public void write(final char[] chr) {
+        write(chr, 0, chr.length);
```
garydgregory, line 124: This defeats the purpose of writing to nowhere ("/dev/null").

This contradicts the class Javadoc. It seems to me that our "Null", "Broken", and "Closed" classes are different beasts from other stream/reader/writer implementations.

src/main/java/org/apache/commons/io/output/NullWriter.java
```diff
@@ -87,7 +89,7 @@ public Writer append(final CharSequence csq) {
      */
     @Override
     public Writer append(final CharSequence csq, final int start, final int end) {
-        //to /dev/null
+        IOUtils.checkFromToIndex(csq, start, end);
```
garydgregory, line 99: This contradicts the class Javadoc. It seems to me that our "Null", "Broken", and "Closed" classes are different beasts from other stream/reader/writer implementations.

garydgregory (changes requested):
Hi @ppkarwasz 

It seems to me that our "Null", "Broken", and "Closed" classes are different beasts from other stream/reader/writer implementations.

Please see my comments scattered throughout. Some comments apply to more than one call site.

TY

src/main/java/org/apache/commons/io/input/ClosedReader.java
ppkarwasz: The method's Javadoc needs certainly to be updated. The class Javadoc itself says:

> Typically uses of this class include testing for corner cases in methods that accept readers and acting as a sentinel value instead of a `null` reader.

If it is used as a sentinel in production code (like, e.g. in `CloseShieldReader`), callers should still be able to rely on the standard `Reader` semantics.

For test scenarios, if callers don't want to pass a char array as argument, they can use Mockito instead.

I fixed the Javadoc and tests in https://github.com/apache/commons-io/pull/790/commits/68c1ba463d3d0d48ddf2e0ea0a06c5153d5052e1.

src/main/java/org/apache/commons/io/input/BrokenReader.java
ppkarwasz: I agree: users should be able to use the class to emulate a “Stream closed” condition, which is notably checked **before** parameter validation.

In https://github.com/apache/commons-io/pull/790/commits/48fe1c34398470fb139ed306c48885c91da9f857 I reverted the changes, improved the test suite to include cases with invalid parameters and added an override of `read(char[])` and `read(CharBuffer)` to ensure that the right exception is thrown, even if the parameter is `null`.

ppkarwasz: Thinking more about it, the `read(char[])` and `read(CharBuffer)` overrides are excessive and I removed them in https://github.com/apache/commons-io/pull/790/commits/bbde62705159668acddaaab29163346e06d05119. Since almost no class overrides those, users should expect an NPE if the argument is `null`.

src/main/java/org/apache/commons/io/output/BrokenWriter.java
ppkarwasz: Same as above: fixed in https://github.com/apache/commons-io/pull/790/commits/de2d1e376421622ea3f700e6c59ff90a8e2c143d with additional tests to record this conversation.

src/main/java/org/apache/commons/io/output/ClosedOutputStream.java
ppkarwasz: This is similar to https://github.com/apache/commons-io/pull/790#discussion_r2390105476, with the difference that I couldn't find `java.io` classes that check for closure, before validating arguments.

`FileOutputStream`, `OutputStream.nullOutputStream()` and even `BufferedOutputStream` and `PipedOutputStream` check the arguments **first**.

src/main/java/org/apache/commons/io/output/ClosedWriter.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/06cb03b158c105427f79c1c39b7a50a5bb57ba8d with some additional tests.

src/main/java/org/apache/commons/io/output/FilterCollectionWriter.java
ppkarwasz: I agree and I reverted the changes in https://github.com/apache/commons-io/pull/790/commits/3678d1e22600febb067c606171ab77ff26c77241.

This class should just follow the behavior of its delegate writers.

src/main/java/org/apache/commons/io/output/FilterCollectionWriter.java
ppkarwasz: As above, fixed in https://github.com/apache/commons-io/pull/790/commits/3678d1e22600febb067c606171ab77ff26c77241

src/main/java/org/apache/commons/io/output/NullAppendable.java
ppkarwasz: I fixed the Javadoc in https://github.com/apache/commons-io/pull/790/commits/d00d5effe0da276f7c0d0caa1b1cef77f6c2103d and added some tests to assert that arguments are validated.

The current behavior aligns with `Writer.nullWriter()`, except `close()` is a no-op in this implementation.

src/main/java/org/apache/commons/io/output/NullOutputStream.java
ppkarwasz: I fixed the Javadoc in https://github.com/apache/commons-io/pull/790/commits/1e558f5e05cb2c67fc3a4f7404a73ddae02c6a81 and added some tests to assert that arguments are validated.

The proposed behavior aligns with `OutputStream.nullOutputStream()`, except `close()` is a no-op in this implementation.

src/main/java/org/apache/commons/io/output/NullWriter.java
ppkarwasz: I fixed the Javadoc of `NullWriter` and expanded the test suite in https://github.com/apache/commons-io/pull/790/commits/6585383d23a7cc02acfb977125b16e00965d5458.

Now the class behaves like `Writer.nullWriter()`, except `close()` is a no-op.

src/main/java/org/apache/commons/io/output/NullWriter.java
ppkarwasz: I fixed the Javadoc of `NullWriter` and expanded the test suite in https://github.com/apache/commons-io/pull/790/commits/6585383d23a7cc02acfb977125b16e00965d5458.

Now the class behaves like `Writer.nullWriter()`, except `close()` is a no-op.

src/main/java/org/apache/commons/io/input/NullInputStream.java
```diff
@@ -292,7 +294,8 @@ public int read(final byte[] bytes) throws IOException {
...
+        if (length == 0) {
```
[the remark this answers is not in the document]
ppkarwasz: If `bytes.length` is zero, then either `length` is zero too or `IndexOutOfBoundsException` is thrown for any other value of `length`.

src/main/java/org/apache/commons/io/input/buffer/CircularBufferInputStream.java
```diff
@@ -125,12 +125,9 @@ public int read() throws IOException {
 
     @Override
     public int read(final byte[] targetBuffer, final int offset, final int length) throws IOException {
-        Objects.requireNonNull(targetBuffer, "targetBuffer");
-        if (offset < 0) {
-            throw new IllegalArgumentException("Offset must not be negative");
-        }
-        if (length < 0) {
-            throw new IllegalArgumentException("Length must not be negative");
+        IOUtils.checkFromIndexSize(targetBuffer, offset, length);
```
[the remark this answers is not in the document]
ppkarwasz: The new error message contains all three values (`off`, len` and `b.length`), so it is equivalent.

src/main/java/org/apache/commons/io/output/StringBuilderWriter.java
```diff
@@ -149,9 +151,8 @@ public String toString() {
      */
     @Override
     public void write(final char[] value, final int offset, final int length) {
-        if (value != null) {
-            builder.append(value, offset, length);
-        }
+        IOUtils.checkFromIndexSize(value, offset, length);
+        builder.append(value, offset, length);
```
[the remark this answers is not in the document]
ppkarwasz: You are right, the behavior changes, but users are probably not exploiting this behavior, since they need to write code that works with **any** `Writer` implementation.

@garydgregory, what do you think?

src/main/java/org/apache/commons/io/output/XmlStreamWriter.java
```diff
@@ -306,6 +306,7 @@ public String getEncoding() {
      */
     @Override
     public void write(final char[] cbuf, final int off, final int len) throws IOException {
+        IOUtils.checkFromIndexSize(cbuf, off, len);
```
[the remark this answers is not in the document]
ppkarwasz: The check appears before the call to `prologWriter`.

src/main/java/org/apache/commons/io/input/ClosedInputStream.java
```diff
@@ -86,6 +86,10 @@ public int read() {
      */
     @Override
     public int read(final byte[] b, final int off, final int len) throws IOException {
+        IOUtils.checkFromIndexSize(b, off, len);
```
garydgregory, line 91: The class and method Javadoc are out of sync with the code now.
Javadoc `@throws` IndexOutOfBoundsException is missing.

src/main/java/org/apache/commons/io/output/NullWriter.java
```diff
@@ -77,17 +79,24 @@ public Writer append(final CharSequence csq) {
...
+     *              subsequence.
+     * @return {@code this} instance.
+     * @throws IndexOutOfBoundsException If {@code start} or {@code end} are negative, {@code end} is
+     *                                   greater than {@code csq.length()}, or {@code start} is greater
+     *                                   than {@code end}.
      * @since 2.0
      */
     @Override
     public Writer append(final CharSequence csq, final int start, final int end) {
-        //to /dev/null
```
garydgregory, line 90: What about moving this comment down below (instead of deleting it) like in the other methods with this kind of comment?

src/main/java/org/apache/commons/io/output/StringBuilderWriter.java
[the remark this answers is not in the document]
garydgregory: I'd think maintaining the behavior is simpler than creating a possible surprise for downstream users. We can always review later. The point of this PR is only to use the new API.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -406,6 +406,183 @@ private static char[] charArray(final int size) {
...
+     * @throws IndexOutOfBoundsException if the range {@code [fromIndex, toIndex)} is out of bounds for {@code seq}
+     * @see Appendable#append(CharSequence, int, int)
+     * @since 2.21.0
+     */
+    public static void checkFromToIndex(final CharSequence seq, final int fromIndex, final int toIndex) {
+        checkFromToIndex(fromIndex, toIndex, seq != null ? seq.length() : 4);
+    }
+
+    static void checkFromToIndex(final int fromIndex, final int toIndex, final int length) {
+        if (fromIndex < 0 | toIndex < fromIndex | length < toIndex) {
```
garydgregory: Using `|` instead of `||` evaluates everything unneccessarily, see the JRE implementation (Java 16-25's `jdk.internal.util.Preconditions.checkFromToIndex(long, long, long, BiFunction<String, List<Number>, X>)`)

src/main/java/org/apache/commons/io/input/BoundedReader.java
```diff
@@ -121,6 +123,7 @@ public int read() throws IOException {
      */
     @Override
     public int read(final char[] cbuf, final int off, final int len) throws IOException {
+        IOUtils.checkFromIndexSize(cbuf, off, len);
```
garydgregory, line 128: Javadoc `@throws` IndexOutOfBoundsException is missing.

src/main/java/org/apache/commons/io/input/CharSequenceReader.java
```diff
@@ -209,14 +210,13 @@ public int read() {
      */
     @Override
     public int read(final char[] array, final int offset, final int length) {
+        IOUtils.checkFromIndexSize(array, offset, length);
```
garydgregory, line 214: Javadoc `@throws` IndexOutOfBoundsException is missing.

src/main/java/org/apache/commons/io/input/NullReader.java
```diff
@@ -276,6 +278,10 @@ public int read(final char[] chars) throws IOException {
      */
     @Override
     public int read(final char[] chars, final int offset, final int length) throws IOException {
+        IOUtils.checkFromIndexSize(chars, offset, length);
```
garydgregory, line 283: Javadoc `@throws` IndexOutOfBoundsException is missing.

src/main/java/org/apache/commons/io/output/ChunkedWriter.java
```diff
@@ -74,6 +74,7 @@ public ChunkedWriter(final Writer writer, final int chunkSize) {
      */
     @Override
     public void write(final char[] data, final int srcOffset, final int length) throws IOException {
+        IOUtils.checkFromIndexSize(data, srcOffset, length);
```
garydgregory, line 80: Javadoc `@throws IndexOutOfBoundsException ...` is missing.

src/main/java/org/apache/commons/io/input/BOMInputStream.java
```diff
@@ -444,6 +444,10 @@ public int read(final byte[] buf) throws IOException {
      */
     @Override
     public int read(final byte[] buf, int off, int len) throws IOException {
+        IOUtils.checkFromIndexSize(buf, off, len);
```
garydgregory, line 453: Javadoc `@throws` IndexOutOfBoundsException is missing.

src/main/java/org/apache/commons/io/input/NullInputStream.java
garydgregory: k, I am resolving these types of issues (GitHub PR resolving that is).
But:
Javadoc `@throws` IndexOutOfBoundsException is missing.

src/main/java/org/apache/commons/io/input/ReaderInputStream.java
```diff
@@ -456,14 +455,11 @@ public int read(final byte[] b) throws IOException {
      */
     @Override
     public int read(final byte[] array, int off, int len) throws IOException {
-        Objects.requireNonNull(array, "array");
-        if (len < 0 || off < 0 || off + len > array.length) {
-            throw new IndexOutOfBoundsException("Array size=" + array.length + ", offset=" + off + ", length=" + len);
-        }
-        int read = 0;
+        IOUtils.checkFromIndexSize(array, off, len);
```
garydgregory, line 461: Javadoc `@throws` IndexOutOfBoundsException is missing.

src/main/java/org/apache/commons/io/input/SequenceReader.java
```diff
@@ -114,9 +115,9 @@ public int read() throws IOException {
      */
     @Override
     public int read(final char[] cbuf, int off, int len) throws IOException {
-        Objects.requireNonNull(cbuf, "cbuf");
-        if (len < 0 || off < 0 || off + len > cbuf.length) {
-            throw new IndexOutOfBoundsException("Array Size=" + cbuf.length + ", offset=" + off + ", length=" + len);
+        IOUtils.checkFromIndexSize(cbuf, off, len);
```
garydgregory, line 113: Either we should inherit the whole Javadoc or add a real Javadoc to document `@throws` `IndexOutOfBoundsException`.

src/main/java/org/apache/commons/io/output/AppendableWriter.java
```diff
@@ -124,10 +126,7 @@ public T getAppendable() {
      */
     @Override
     public void write(final char[] cbuf, final int off, final int len) throws IOException {
-        Objects.requireNonNull(cbuf, "cbuf");
-        if (len < 0 || off + len > cbuf.length) {
-            throw new IndexOutOfBoundsException("Array Size=" + cbuf.length + ", offset=" + off + ", length=" + len);
-        }
+        IOUtils.checkFromIndexSize(cbuf, off, len);
```
garydgregory, line 133: Javadoc `@throws IndexOutOfBoundsException ...` is missing.

src/main/java/org/apache/commons/io/output/ChunkedOutputStream.java
```diff
@@ -172,6 +172,7 @@ int getChunkSize() {
      */
     @Override
     public void write(final byte[] data, final int srcOffset, final int length) throws IOException {
+        IOUtils.checkFromIndexSize(data, srcOffset, length);
```
garydgregory, line 178: Javadoc `@throws IndexOutOfBoundsException ...` is missing.

src/main/java/org/apache/commons/io/output/ClosedOutputStream.java
garydgregory: OK.

Javadoc `@throws IndexOutOfBoundsException ...` is missing.

src/main/java/org/apache/commons/io/output/StringBuilderWriter.java
```diff
@@ -149,9 +151,8 @@ public String toString() {
      */
     @Override
     public void write(final char[] value, final int offset, final int length) {
-        if (value != null) {
-            builder.append(value, offset, length);
-        }
+        IOUtils.checkFromIndexSize(value, offset, length);
```
garydgregory: Javadoc `@throws IndexOutOfBoundsException ...` is missing.

src/main/java/org/apache/commons/io/output/ThresholdingOutputStream.java
```diff
@@ -249,6 +250,7 @@ public void write(final byte[] b) throws IOException {
     @SuppressWarnings("resource") // the underlying stream is managed by a subclass.
     @Override
     public void write(final byte[] b, final int off, final int len) throws IOException {
+        IOUtils.checkFromIndexSize(b, off, len);
```
garydgregory, line 256: Javadoc `@throws IndexOutOfBoundsException ...` is missing.

src/main/java/org/apache/commons/io/output/WriterOutputStream.java
```diff
@@ -448,6 +448,7 @@ public void write(final byte[] b) throws IOException {
      */
     @Override
     public void write(final byte[] b, int off, int len) throws IOException {
+        IOUtils.checkFromIndexSize(b, off, len);
```
garydgregory, line 454: Javadoc `@throws IndexOutOfBoundsException ...` is missing.

src/main/java/org/apache/commons/io/output/XmlStreamWriter.java
garydgregory, line 312: Javadoc `@throws IndexOutOfBoundsException ...` is missing.

garydgregory (changes requested):
Hi @ppkarwasz 

I added comments.

A lot of them are: Javadoc `@throws IndexOutOfBoundsException ...` is missing.

src/main/java/org/apache/commons/io/input/NullInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/14a250629ba13693c802a5a47592b1c634e1fb2e

src/main/java/org/apache/commons/io/output/StringBuilderWriter.java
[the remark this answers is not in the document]
ppkarwasz: I restored the behavior in https://github.com/apache/commons-io/pull/790/commits/94760281b40b4a442c7d784713525e94bb0ed694

src/main/java/org/apache/commons/io/output/NullWriter.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/b80971821c227b09de4e2aec5a2adcee7b4b4a3b

src/main/java/org/apache/commons/io/output/ClosedOutputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/1632fb2edcd4dd3f425b0e84e1111f0cfa251fe9

src/main/java/org/apache/commons/io/IOUtils.java
ppkarwasz: Nice catch! :100:

The bitwise or was obviously a typo. Fixed in https://github.com/apache/commons-io/pull/790/commits/a70be15bbb8c493a95144542f5a8fd8788e5b8cd

src/main/java/org/apache/commons/io/output/StringBuilderWriter.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/94760281b40b4a442c7d784713525e94bb0ed694

src/main/java/org/apache/commons/io/input/ClosedInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/c4e4f3b5b09d59d0078a7d8b3f3772dea728e06f

src/main/java/org/apache/commons/io/output/ThresholdingOutputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/c4e4f3b5b09d59d0078a7d8b3f3772dea728e06f

src/main/java/org/apache/commons/io/input/BoundedReader.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/8d520f72c3698440e2275b860a991a41757ad2fb

src/main/java/org/apache/commons/io/input/CharSequenceReader.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/8d520f72c3698440e2275b860a991a41757ad2fb

src/main/java/org/apache/commons/io/input/NullReader.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/8d520f72c3698440e2275b860a991a41757ad2fb

src/main/java/org/apache/commons/io/output/ChunkedWriter.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/8d520f72c3698440e2275b860a991a41757ad2fb

src/main/java/org/apache/commons/io/input/BOMInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/8d520f72c3698440e2275b860a991a41757ad2fb

src/main/java/org/apache/commons/io/input/ReaderInputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/8d520f72c3698440e2275b860a991a41757ad2fb

src/main/java/org/apache/commons/io/input/SequenceReader.java
ppkarwasz: I removed the Javadoc in https://github.com/apache/commons-io/pull/790/commits/8d520f72c3698440e2275b860a991a41757ad2fb

src/main/java/org/apache/commons/io/output/AppendableWriter.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/8d520f72c3698440e2275b860a991a41757ad2fb

src/main/java/org/apache/commons/io/output/ChunkedOutputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/8d520f72c3698440e2275b860a991a41757ad2fb

src/main/java/org/apache/commons/io/output/WriterOutputStream.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/8d520f72c3698440e2275b860a991a41757ad2fb

src/main/java/org/apache/commons/io/output/XmlStreamWriter.java
ppkarwasz: Fixed in https://github.com/apache/commons-io/pull/790/commits/8d520f72c3698440e2275b860a991a41757ad2fb

PR #791, opened 2025-10-02
Improve `FileUtils.forceDelete()` tests on Windows

ppkarwasz:
On Windows, the `DeleteFile` Win32 API has a little quirk: it refuses to delete files with the legacy **DOS read-only attribute** set. (Because apparently 99% of Windows users don’t realize that “deleting a file” is actually an operation on the *directory*, not the file itself 😉).

So the usual drill is: clear the read-only flag first, then delete.

* Until JDK 25, `File.delete()` did this for you behind the scenes: it quietly stripped the flag before calling into `DeleteFile`. That meant your file might be left behind (with the flag missing) if the *real* ACLs didn’t allow deletion.
* From JDK 25 onward, `File.delete()` doesn’t touch the flag anymore. If the bit is set, `DeleteFile` fails, end of story.
* `FileUtils.forceDelete()` already knows how to juggle the flag itself, so its behavior didn’t change.

This PR:

* Updates two tests that were (unfairly) comparing `File.delete` with `FileUtils.forceDelete`. With JDK 25, their expectations diverged.
* <s>Adds a new test to confirm that `FileUtils.forceDelete` restores the read-only flag if the actual deletion fails.</s>


garydgregory:
Merged, thank you @ppkarwasz.

ppkarwasz:
Since the issues around `FileUtils.delete` and `PathUtils.setReadOnly` turned out to be broader than expected, this PR only contains the minimal changes needed to get the existing tests passing.

I’ve opened three follow-up JIRA tickets to track the remaining problems:

* [IO-878(https://issues.apache.org/jira/browse/IO-878): semantic and implementation issues in `PathUtils.setReadOnly` that still need clarification.
* [IO-879](https://issues.apache.org/jira/browse/IO-879): DOS read-only flag not being restored when a deletion fails.
* [IO-880](https://issues.apache.org/jira/browse/IO-880): recursive deletion overriding permissions, where `rm -r` semantics would normally fail.

Partial fixes for these are available in the [`fix/read-only`](https://github.com/apache/commons-io/tree/fix/read-only) branch, but additional work is required.

PR #795, opened 2025-10-03
Add missing `read` argument validation in `IOUtils`

ppkarwasz:
In #790 I introduced `IOUtils#checkIndexFromLength` calls to validate arguments across the codebase. Ironically, the `IOUtils` class itself was left out.

This PR addresses that omission by adding argument validation to `IOUtils#read` and `IOUtils#readFully`.

Key points:

* Ensures consistency with the rest of Commons IO by validating `offset` and `length`.
* Fixes inconsistent exception behavior:

  * Previously, `length < 0` resulted in an `IllegalArgumentException`.
  * `offset < 0` did not trigger validation and failed later with an `IndexOutOfBoundsException`.
* With this change, both invalid cases are handled consistently and upfront.


garydgregory:
Merged, thank you @ppkarwasz.

PR #796, opened 2025-10-03
Fix inconsistent exception in `IOUtils.toByteArray`

ppkarwasz:
The implementation of `IOUtils.toByteArray(InputStream, int, int)` added in #776 throws different exceptions depending on the requested size:

* For request sizes larger than the internal chunk size, it correctly throws an `EOFException`.
* For smaller requests, it incorrectly throws a generic `IOException`.

This PR makes the behavior consistent by always throwing an `EOFException` when the stream ends prematurely.

> [!NOTE]
> This also affects `RandomAccessFiles.read`. Its previous truncation behavior was undocumented and inconsistent with `RandomAccessFile.read` (which reads as much as possible). The new behavior is not explicitly documented here either, since it is unclear whether throwing on truncation is actually desirable.


garydgregory:
Merged, thank you @ppkarwasz.

PR #805, opened 2025-10-19
Add PathFence

garydgregory:
A Path fence guards against using paths outside of a "fence" of made of root paths.

This was extracted from [Apache Commons Text](https://commons.apache.org/proper/commons-text/)'s private [PathFence](https://github.com/apache/commons-text/blob/master/src/main/java/org/apache/commons/text/lookup/PathFence.java).

- [x] I used AI to create any part of, or all of, this pull request: _The initial unit test was created and then heavily and manually modified using Co-Pilot_.


ppkarwasz:
This API needs a clear security model: is it intended for reading existing files, creating new files or both?

The path resolution strategy should change depending on the use case:

1. **Reading trusted configuration files**
  When reading from trusted system locations (e.g. configuration directories managed by sysadmins), we typically must allow symlinks. Example: On Debian, `CATALINA_BASE` is `/var/lib/tomcat`, but `/var/lib/tomcat/conf` is a symlink to `/etc/tomcat`, which should be allowed.

2. **Reading files from untrusted user input**
  For user-supplied paths, following symlinks can enable path traversal attacks. In this case, symlinks should generally **not be followed**.

At the moment, the class performs only **syntactic** path validation using `path.toAbsolutePath().normalize()`. This is sufficient for trusted environments (e.g. system-managed configuration directories), but it does **not** call `Path.toRealPath()`, which resolves symlinks. As a result, when used with a root directory that may contain **untrusted content**, it can be bypassed by symlink-based path traversal attacks.

src/main/java/org/apache/commons/io/file/PathFence.java
```diff
@@ -0,0 +1,161 @@
...
+    // Cannot implement both UnaryOperator<Path> and Function<String, Path>
+    public Path apply(final String fileName) {
+        return getPath(fileName, Paths.get(fileName));
+    }
+
+    private Path getPath(final String fileName, final Path path) {
+        if (roots.isEmpty()) {
+            return path;
+        }
+        final Path pathAbs = path.normalize().toAbsolutePath();
```
ppkarwasz:
```suggestion
        final Path pathAbs = path.toAbsolutePath().normalize();
```

If you call it in the reverse order, `normalize()` will not remove leading `..` segments. For example on UNIX:

- if `path` equals `../../etc/passwd`,
- then `path.normalize()` equals `../../etc/passwd`,
- and `path.normalize().toAbsolutePath()` equals `/home/piotr/../../etc/passwd`.

src/main/java/org/apache/commons/io/file/PathFence.java
```diff
@@ -0,0 +1,171 @@
...
+    /**
+     * Gets a Path for the given file name, checking that it is within our fence.
+     *
+     * @param fileName the file name to test.
+     * @return The given path.
+     * @throws IllegalArgumentException Thrown if the file name is not within our fence.
+     */
+    // Cannot implement both UnaryOperator<Path> and Function<String, Path>
+    public Path apply(final String fileName) {
+        return getPath(fileName, Paths.get(fileName));
```
Marcono1234, line 157: Usage of `Paths` ties this to the default file system. Is that intended (if so would be good to document it), or would it make more sense to ensure that all `roots` have the same file system and then use `root.getFileSystem().getPath(...)`? (would only work though if `roots` is non-empty)

src/main/java/org/apache/commons/io/file/PathFence.java
```diff
@@ -0,0 +1,171 @@
...
+
+    private Path getPath(final String fileName, final Path path) {
+        if (roots.isEmpty()) {
+            return path;
+        }
+        final Path pathAbs = absoluteNormalize(path);
+        final Optional<Path> first = roots.stream().filter(pathAbs::startsWith).findFirst();
+        if (first.isPresent()) {
+            return path;
+        }
```
Marcono1234, line 168: Use `Stream#anyMatch` instead of `stream.findFirst().isPresent()`?

src/main/java/org/apache/commons/io/file/PathFence.java
```diff
@@ -0,0 +1,171 @@
...
+     */
+    // Cannot implement both UnaryOperator<Path> and Function<String, Path>
+    public Path apply(final String fileName) {
+        return getPath(fileName, Paths.get(fileName));
+    }
+
+    private Path getPath(final String fileName, final Path path) {
+        if (roots.isEmpty()) {
+            return path;
+        }
```
Marcono1234, line 163: So if `roots` is empty, the path fence is effectively disabled?

Maybe would be better to add an explicit `Builder` method / static `PathFence` factory method for that instead? Otherwise when accidentally not calling `setRoots` or calling it with an empty array, it might give a false sense of security when `PathFence` is actually ineffective.

src/test/java/org/apache/commons/io/file/PathFenceTest.java
```diff
@@ -0,0 +1,144 @@
...
+    void testEmpty(final String test) {
+        // An empty fence accepts anything
+        final PathFence fence = PathFence.builder().get();
+        final Path path = Paths.get(test);
+        assertEquals(path, fence.apply(test));
+        assertSame(path, fence.apply(path));
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = { "/a/b", "/a/b/c", "/a/b/c/d", "a", "a/b", "a/b/c", "a/b/c/d" })
```
Marcono1234, line 83: Should this also cover `""` or `"/"` to ensure https://github.com/apache/commons-io/pull/805#discussion_r2443377598 is fixed?

Marcono1234 (commented):
Hopefully these comments are useful, and not disruptive. Though I understand that this PR is still a draft and the design and implementation decisions on this PR are not yet final.

src/main/java/org/apache/commons/io/file/PathFence.java
ecki: Agreed, this should fail securely

src/main/java/org/apache/commons/io/file/PathFence.java
```diff
@@ -0,0 +1,171 @@
...
+ * <p>
+ * To use the current directory as the path fence:
+ * </p>
+ * <pre>
+ * PathFence fence = PathFence.builder().setRoots(PathUtils.current()).get();
+ * </pre>
+ *
+ * @since 2.21.0
+ */
+// Cannot implement both UnaryOperator<Path> and Function<String, Path>, so don't pick one over the other
```
ecki, line 63: What about a predicate instead?

src/main/java/org/apache/commons/io/file/PathFence.java
```diff
@@ -0,0 +1,171 @@
...
+
+    /**
+     * Checks that that a Path is within our fence.
+     *
+     * @param path The path to test.
+     * @return The given path.
+     * @throws IllegalArgumentException Thrown if the path is not within our fence.
+     */
+    // Cannot implement both UnaryOperator<Path> and Function<String, Path>
+    public Path apply(final Path path) {
```
ecki, line 144: I find apply() a bit non descriptive, and why does it return a Path instead of a boolean? Not sure if IAE should be thrown as a result of a policy check.

src/main/java/org/apache/commons/io/file/PathFence.java
```diff
@@ -0,0 +1,171 @@
...
+     * @param fileName the file name to test.
+     * @return The given path.
+     * @throws IllegalArgumentException Thrown if the file name is not within our fence.
+     */
+    // Cannot implement both UnaryOperator<Path> and Function<String, Path>
+    public Path apply(final String fileName) {
+        return getPath(fileName, Paths.get(fileName));
+    }
+
+    private Path getPath(final String fileName, final Path path) {
```
ecki, line 160: GetPath is a strange name for this, how about “isInsideFence” or something like that? Also why does it have two argumente, one of it unused?

src/main/java/org/apache/commons/io/file/PathFence.java
```diff
@@ -0,0 +1,171 @@
...
+        if (roots.isEmpty()) {
+            return path;
+        }
+        final Path pathAbs = absoluteNormalize(path);
+        final Optional<Path> first = roots.stream().filter(pathAbs::startsWith).findFirst();
+        if (first.isPresent()) {
+            return path;
+        }
+        throw new IllegalArgumentException(String.format("[%s] -> [%s] not in the fence %s", fileName, pathAbs, roots));
```
ecki, line 169: It’s typically frowned upon to disclose system infos (like resolved absolute path) in exceptions, at least in the toString/getMessage

src/test/java/org/apache/commons/io/file/PathFenceTest.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Test
+    void testResolveRelativeRoot() throws Exception {
+        final Path relPathTop = getRelPathToTop();
+        final PathFence fence = PathFence.builder().setRoots(relPathTop.resolve("foo/bar")).get();
+        final Path relPath = relPathTop.resolve("foo/bar");
+        assertSame(relPath, fence.apply(relPath));
+    }
+
```
ecki, line 143: Should probably also define what to do with \\ separator and test for windows ans linux machines.

Should it also deny \0 bytes before using Path API?

ecki, line 143: And it should probably also describe if or if not how syml8nks are handled (toRealPath?)

ecki (changes requested):

src/main/java/org/apache/commons/io/file/PathFence.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /**
+     * Gets a Path for the given file name, checking that it is within our fence.
+     *
+     * @param fileName the file name to test.
+     * @return The given path.
+     * @throws IllegalArgumentException Thrown if the file name is not within our fence.
+     */
+    // Cannot implement both UnaryOperator<Path> and Function<String, Path>
```
ecki, line 155: This comment is pretty confusing

PR #827, opened 2026-02-05
IO-845 PathUtils.copyDirectory copies or follows symbolic links

peterdemaeyer:
Defined the behavior of `PathUtils.copyDirectory` for symbolic links. It depends on the option `LinkOption.NOFOLLOW_LINKS`. Non-symbolic (hard) links are out of scope, given that the copy behavior for them was already well-defined: they're treated as regular files.
For the sake of conciseness, the following paragraphs use the term "link" for "symbolic link".

Without  `NOFOLLOW_LINKS` option:
- Links are followed, so that there are no links in the target directory.
- Cyclic links result in `FileSystemLoopException`.
- Broken links are ignored, they are not copied at all.

With `NOFOLLOW_LINKS` option:
- Links are copied as links, so that there are links in the target directory for every link in the source directory.
- Links that link to a file or directory inside the source directory are copied to relative links inside the target directory, linking to the copied file or directory.
- Links that link to a file or directory outside the source directory are copied to links inside the target directory, linking to the same file or directory as the link in the source directory.
- Cyclic links are preserved, mirrored to cyclic links in the target directory.
- Broken links are ignored, they are not copied at all.

All of this is also explained in the updated JavaDoc.

- [x] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute?
- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied. This may not always be possible, but it is a best practice.

### DIFF
--- a/src/main/java/org/apache/commons/io/file/CopyDirectoryVisitor.java
+++ b/src/main/java/org/apache/commons/io/file/CopyDirectoryVisitor.java
@@ -28,6 +28,7 @@ import java.util.Arrays;
 import java.util.Objects;
 
 import org.apache.commons.io.file.Counters.PathCounters;
+import org.apache.commons.io.filefilter.TrueFileFilter;
 
 /**
  * Copies a source directory to a target directory.
@@ -53,7 +54,7 @@ public class CopyDirectoryVisitor extends CountingPathVisitor {
      * @param copyOptions Specifies how the copying should be done.
      */
     public CopyDirectoryVisitor(final PathCounters pathCounter, final Path sourceDirectory, final Path targetDirectory, final CopyOption... copyOptions) {
-        super(pathCounter);
+        super(pathCounter, TrueFileFilter.INSTANCE, TrueFileFilter.INSTANCE);
         this.sourceDirectory = sourceDirectory;
         this.targetDirectory = targetDirectory;
         this.copyOptions = toCopyOption(copyOptions);
--- a/src/main/java/org/apache/commons/io/file/CopyDirectoryVisitor.java
+++ b/src/main/java/org/apache/commons/io/file/CopyDirectoryVisitor.java
@@ -28,7 +28,6 @@ import java.util.Arrays;
 import java.util.Objects;
 
 import org.apache.commons.io.file.Counters.PathCounters;
-import org.apache.commons.io.filefilter.TrueFileFilter;
 
 /**
  * Copies a source directory to a target directory.
@@ -54,7 +53,7 @@ public class CopyDirectoryVisitor extends CountingPathVisitor {
      * @param copyOptions Specifies how the copying should be done.
      */
     public CopyDirectoryVisitor(final PathCounters pathCounter, final Path sourceDirectory, final Path targetDirectory, final CopyOption... copyOptions) {
-        super(pathCounter, TrueFileFilter.INSTANCE, TrueFileFilter.INSTANCE);
+        super(pathCounter);
         this.sourceDirectory = sourceDirectory;
         this.targetDirectory = targetDirectory;
         this.copyOptions = toCopyOption(copyOptions);
--- a/src/main/java/org/apache/commons/io/file/CountingPathVisitor.java
+++ b/src/main/java/org/apache/commons/io/file/CountingPathVisitor.java
@@ -304,18 +304,20 @@ public class CountingPathVisitor extends SimplePathVisitor {
     }
 
     /**
-     * Updates the counters for visiting the given file.
+     * Updates the counters for visiting the given file, ignoring symbolic links.
+     * <p/>
+     * According to the JavaDoc, {@link BasicFileAttributes#size} is only well-defined for regular files.
+     * For symbolic links on Linux for example, it counts the # (charset dependent?) bytes in the inode name,
+     * which is <i>not</i>> what we want to count here.
+     * Intuitively, the appropriate check would be {@link Files#isRegularFile}.
+     * However, for symbolic links, {@code isRegularFile} returns {@code true} under a "follow links" regime.
+     * That would still not give us what we want, so instead we settle for a {@code !Files.isSymbolicLink} check.
      *
      * @param file       the visited file.
      * @param attributes the visited file attributes.
      */
     protected void updateFileCounters(final Path file, final BasicFileAttributes attributes) {
         pathCounters.getFileCounter().increment();
-        // According to the JavaDoc, BasicFileAttributes.size() is only well-defined for regular files.
-        // For symbolic links on Linux for example, it counts the # (charset dependent?) bytes in the inode name, which is NOT what we want to count here.
-        // Intuitively, the appropriate check would be Files.isRegularFile.
-        // However, for symbolic links, isRegularFile returns true under a "follow links" regime.
-        // That would still not give us what we want, so instead we settle for a !Files.isSymbolicLink check.
         if (!Files.isSymbolicLink(file)) {
             pathCounters.getByteCounter().add(attributes.size());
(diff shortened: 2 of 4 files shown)


src/main/java/org/apache/commons/io/file/PathUtils.java
```diff
@@ -377,7 +400,22 @@ public static long copy(final IOSupplier<InputStream> in, final Path target, fin
...
-        return visitFileTree(new CopyDirectoryVisitor(Counters.longPathCounters(), absoluteSource, targetDirectory, copyOptions), absoluteSource)
+        // CopyOption.NOFOLLOW_LINKS is the explicit option, "follow symlinks" the implicit default.
+        // For FileVisitOption it's the other way around: FileVisitOption.FOLLOW_LINKS is the explicit option, and "do not follow symlinks" is the implicit
+        // default.
+        // If they're not in sync, the behavior is inconsistent, so we have to make sure they're in sync.
+        // CopyOption is given by the caller, FileVisitOption is under our control here, so we sync the latter to the former.
+        Set<FileVisitOption> fileVisitOptions = FOLLOW_LINKS_FILE_VISIT_OPTIONS;
+        if (copyOptions != null) {
+            for (CopyOption copyOption : copyOptions) {
+                if (LinkOption.NOFOLLOW_LINKS.equals(copyOption)) {
```
garydgregory: I think we want to use `==` when comparing enum values.

src/main/java/org/apache/commons/io/file/CopyDirectoryVisitor.java
```diff
@@ -53,7 +54,7 @@ private static CopyOption[] toCopyOption(final CopyOption... copyOptions) {
      * @param copyOptions Specifies how the copying should be done.
      */
     public CopyDirectoryVisitor(final PathCounters pathCounter, final Path sourceDirectory, final Path targetDirectory, final CopyOption... copyOptions) {
-        super(pathCounter);
+        super(pathCounter, TrueFileFilter.INSTANCE, TrueFileFilter.INSTANCE);
```
garydgregory: I think we have interactions to define for users:

- before the PR, we were calling `super(pathCounter)`, which means the default file filer is `new SymbolicLinkFileFilter(FileVisitResult.TERMINATE, FileVisitResult.CONTINUE)` and the default dir filter is `TrueFileFilter.INSTANCE`
- In this PR, we set both to `TrueFileFilter.INSTANCE`, so no change for dirs.

It seems both could be "wrong" in the sense that they compete with `CopyOption...` which can contain `java.nio.file.LinkOption.NOFOLLOW_LINKS`.

I'm sorry this is all confusing! We have:
- `java.nio.file.LinkOption.NOFOLLOW_LINKS` (where `LinkOption implements OpenOption, CopyOption`)
- `java.nio.file.FileVisitOption.FOLLOW_LINKS`
- An IO file filter
- An IO dir filter

I understand we have a bug, but we should take this opportunity to document what this change means.

I should add that if coming up with a good solution means adding APIs and/or deprecating existing ones, I'm OK with that.

I also would like to create a release candidate to pick up existing fixes (for other Commons components and projects) but I don't want to rush this issue. If you see a release candidate tag or emails on the dev mailing list, it's for that reason, not because I don't think this is important :)

garydgregory (changes requested):
Hello @peterdemaeyer !
I have a comment in the copy visitor, and a small implementation issue about comparing enums.

src/main/java/org/apache/commons/io/file/CountingPathVisitor.java
```diff
@@ -311,7 +311,14 @@ protected void updateDirCounter(final Path dir, final IOException exc) {
      */
     protected void updateFileCounters(final Path file, final BasicFileAttributes attributes) {
         pathCounters.getFileCounter().increment();
-        pathCounters.getByteCounter().add(attributes.size());
```
garydgregory, line 314: Great documentation. This should be in the Javadoc of this method IMO.

src/main/java/org/apache/commons/io/file/CopyDirectoryVisitor.java
peterdemaeyer: If you're worried about this API change, I propose to revert it and call the other constructor in the only the place where it's called, which is `PathUtils.copyDirectory`, line 417.

Beyond the scope of the issue at hand, I'd like to also share with you my impressions while working on the fix.

The reason we're even discussing this is because `CopyDirectoryVisitor` is on the API, which maybe should not have been the case in the first place. As a user and fixer of this bug, my focus is to get `copyDirectory` working. I am familiar with the JDK and its `FileVisitor` API, it is the API I would like to use. Instead, I was burdened with having to understand, implement and extend `CopyDirectoryVisitor` < `CountingPathVisitor` < `SimplePathVisitor` < `PathVisitor`/`SimpleFileVisitor`, which does a bunch of stuff I don't need, such as counting bytes and filtering.

Given that `CopyDirectoryVisitor` is only used for the implementation of `PathUtils.copyDirectory`, it looks to me like an implementation detail that leaked onto the API.

Don' get me wrong, I can see a use case for counting bytes and filtering, but I would have offered them as independent `ByteCountingFileVisitor` and `FilteringFileVisitor` decorators, for users to only use them if and when they need them. This is an example of where applying the design principle "prefer composition over inheritance" would have resulted in better separation of concerns and a more flexible API.

garydgregory: I agree with your comments. Thank you for digging in and sharing your thoughts. We can break binary compatibility in a new major version only. Until then I'm happy to entertain new code and deprecated old code.

src/main/java/org/apache/commons/io/file/CountingPathVisitor.java
peterdemaeyer: I followed your advice and promoted it to JavaDoc.

src/main/java/org/apache/commons/io/file/CopyDirectoryVisitor.java
peterdemaeyer: API change reverted and solved as I proposed.

src/main/java/org/apache/commons/io/file/CountingPathVisitor.java
```diff
@@ -304,14 +304,23 @@ protected void updateDirCounter(final Path dir, final IOException exc) {
     }
 
     /**
-     * Updates the counters for visiting the given file.
+     * Updates the counters for visiting the given file, ignoring symbolic links.
+     * <p/>
```
garydgregory: You should use open and close tags for a paragraph, not add an empty one.

src/main/java/org/apache/commons/io/file/CountingPathVisitor.java
```diff
@@ -304,14 +304,23 @@ protected void updateDirCounter(final Path dir, final IOException exc) {
...
+     * According to the JavaDoc, {@link BasicFileAttributes#size} is only well-defined for regular files.
+     * For symbolic links on Linux for example, it counts the # (charset dependent?) bytes in the inode name,
+     * which is <i>not</i>> what we want to count here.
```
garydgregory: For emphasis, use `<em>`, and let the Javadoc stylesheet decide how to render it (same for bold, use `<strong>` instead).

garydgregory (changes requested):
Hello @peterdemaeyer 

Thank you for your update. You'll want to run `mvn` by itself before you push to run all build checks; this runs the default Maven goal as defined in the POM.

PR #857, opened 2026-06-12
[IO-438] Clarify IOUtils.closeQuietly() Javadocs

sarankumarbaskar:
IOUtils.closeQuietly() ignores exceptions thrown during close, which is useful for cleanup paths but can be unsafe when closing output streams or writers after successful writes.

This PR clarifies the Javadocs to warn that output close failures may indicate buffered data was not fully written. It also removes examples that used closeQuietly() after IOUtils.copy(), since those examples could encourage ignoring output close errors.

No runtime behavior changed.

Resolves [IO-438](https://issues.apache.org/jira/browse/IO-438).

Checklist:
- [x] Read the contribution guidelines for this project.
- [x] Read the ASF Generative Tooling Guidance.
- [x] Ran `mvn -Drat.skip=true -DskipTests javadoc:javadoc`

### DIFF
--- a/src/main/java/org/apache/commons/io/IOUtils.java
+++ b/src/main/java/org/apache/commons/io/IOUtils.java
@@ -800,6 +800,20 @@ public class IOUtils {
      * }
      * </pre>
      * <p>
+     * Closing all streams:
+     * </p>
+     *
+     * <pre>
+     * try {
+     *     IOUtils.copy(inputStream, outputStream);
+     *     outputStream.close(); // close errors are handled
+     *     outputStream = null;
+     * } finally {
+     *     IOUtils.closeQuietly(inputStream);
+     *     IOUtils.closeQuietly(outputStream); // only if normal close was skipped
+     * }
+     * </pre>
+     * <p>
      * Also consider using a try-with-resources statement where appropriate.
      * </p>
      *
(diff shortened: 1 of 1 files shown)


garydgregory (changes requested):
@Sarankumar18 
Don't delete examples, propose new ones.

sarankumarbaskar:
@garydgregory  
Thanks, updated. I kept the examples and changed them to close the output stream normally first, then fall back to closeQuietly() only if the normal close was skipped. This keeps the example while avoiding silently ignored output close failures in the successful path.

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -801,10 +805,12 @@ private static void closeQ(final Closeable closeable) {
...
      * <pre>
      * try {
-     *     return IOUtils.copy(inputStream, outputStream);
+     *     IOUtils.copy(inputStream, outputStream);
+     *     outputStream.close(); // close errors are handled
+     *     outputStream = null;
      * } finally {
      *     IOUtils.closeQuietly(inputStream);
-     *     IOUtils.closeQuietly(outputStream);
+     *     IOUtils.closeQuietly(outputStream); // only if normal close was skipped
```
garydgregory, line 813: @Sarankumar18 
This example makes no sense to me:
- First, you `close()` the OS on line 809 with the contradictory comment "close errors are handled"; the close() call _does not_ handle errors.
- Second, in the finally block, the comment "only if normal close was skipped" is not true because the finally block is always executed.
Please come up with an example and comments that make sense and are consistent with what actually happens.
TY

src/main/java/org/apache/commons/io/IOUtils.java
```diff
@@ -853,9 +859,11 @@ public static void closeQuietly(final Closeable closeable) {
      *
      * <pre>
      * try {
-     *     return IOUtils.copy(inputStream, outputStream);
+     *     IOUtils.copy(inputStream, outputStream);
+     *     outputStream.close(); // close errors are handled
```
garydgregory, line 863: Same comment as above.

garydgregory (changes requested):
@Sarankumar18 
Please see my comments.

src/main/java/org/apache/commons/io/IOUtils.java
garydgregory: Ping.

garydgregory:
Closing: no activity in 3 weeks.

PR #860, opened 2026-06-17
[IO-891] Fix BoundedReader.skip(long) accounting and bounds handling

sarankumarbaskar:
BoundedReader.skip(long) was updating charsRead using the requested skip amount instead of the actual number of characters skipped by the underlying reader.

This could make the reader lose track of its bounded position when the delegate skipped fewer characters than requested. It also allowed skip requests to go past maxCharsFromTargetReader, while read() correctly stops at the configured bound.

This PR updates skip(long) to:
- cap skip requests to the remaining bounded range
- respect mark/readAheadLimit limits
- update charsRead using the actual skipped count
- avoid implicit long-to-int narrowing from charsRead += n

Added regression tests for large skip values, bounded skip behavior, actual skipped count, and mark/readAheadLimit behavior.

Fixes: [IO-891](https://issues.apache.org/jira/browse/IO-891)

- [x] Read the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html).
- [x] Run a successful build using the default Maven goal with `mvn`.
- [x] Wrote unit tests that match the behavioral changes and fail without the runtime fix.
- [x] Wrote a pull request description explaining what changed and why.
- [x] Each commit has a meaningful subject line and body.

### DIFF
--- a/src/test/java/org/apache/commons/io/input/BoundedReaderTest.java
+++ b/src/test/java/org/apache/commons/io/input/BoundedReaderTest.java
@@ -24,6 +24,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.FilterReader;
 import java.io.IOException;
 import java.io.LineNumberReader;
 import java.io.Reader;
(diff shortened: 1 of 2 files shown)


src/test/java/org/apache/commons/io/input/BoundedReaderTest.java
```diff
@@ -242,4 +243,42 @@ void testSkipTest() throws IOException {
...
+        }, 5)) {
+            assertEquals(2, mr.skip(5));
+            assertEquals('2', mr.read());
+        }
+    }
+
+    @Test
+    void testSkipRespectsReadAheadLimit() throws IOException {
+        try (BoundedReader mr = new BoundedReader(new StringReader("01234567890"), 10)) {
+            mr.mark(3);
```
garydgregory: Hello @sarankumarbaskar 
I don't think you understand what mark is supposed to do. Please read the docs and see if there's a real issue once you fix these new tests.

garydgregory (changes requested):
@sarankumarbaskar 
Please see my comments.

src/test/java/org/apache/commons/io/input/BoundedReaderTest.java
sarankumarbaskar: @garydgregory  Thanks, I see your point. Before changing this further, I want to clarify the intended behavior.

Should BoundedReader.skip(long) respect only maxCharsFromTargetReader, or should it also honor the active mark/readAheadLimit the same way read() currently does?

I can keep this PR limited to the clear issue — capping skip() to maxCharsFromTargetReader and updating charsRead with the actual skipped count — unless you think skip() should also mirror the readAheadLimit check.

garydgregory: Please review the contract of the mark and reset methods and notice the word "attempt". You're trying to turn override one feature (reading) with another (mark/reset).

Unless you can provide a strong argument that marking should act as a hard limit, I don't see a use case here.

sarankumarbaskar: @garydgregory  Thanks, that clarifies it. I removed the mark/readAheadLimit-specific logic and test. The PR is now limited to skip(long) respecting maxCharsFromTargetReader and updating charsRead with the actual skipped count returned by the delegate.

garydgregory:
@sarankumarbaskar 
The build is broken.
Always run 'mvn' by itself before you push and fix errors.

garydgregory:
Jira ticket is [IO-891](https://issues.apache.org/jira/browse/IO-891).

garydgregory:
I fixed the build, PR merged 🚀

PR #863, opened 2026-06-29
Avoid allocating the extension substring in FilenameUtils.isExtension

nishantmehta:
## Summary

`FilenameUtils.isExtension(String, String)` returned `getExtension(fileName).equals(extension)`, and the `String...` overload did `Stream.of(extensions).anyMatch(getExtension(fileName)::equals)`. Both allocate the extension **substring** (and the varargs form an extra `Stream` pipeline) only to compare it — on a file-filtering path that returns a `boolean`.

This compares the extension region of `fileName` in place via `String.regionMatches`, through a small private helper that mirrors `getExtension(fileName).equals(extension)` exactly: `NOT_FOUND` is treated as the empty extension and a `null` extension never matches, so all documented cases — no-extension files, empty and null extensions, and empty entries in the varargs form — are unchanged. The `String...` overload also drops the `Stream` for a plain loop.

## Measurement

ThreadMXBean allocation driver, 200k warmed ops, `"/home/user/docs/report.final.pdf"`:

| call | before | after |
|------|--------|-------|
| `isExtension(s, "pdf")` | 48 B/op | 0 B/op |
| `isExtension(s, "doc", "pdf")` | 175 B/op | 0 B/op |

## Testing

`FilenameUtilsTest` passes unchanged (41 tests), including the existing null / empty-extension / varargs edge cases (`isExtension("file", "")`, `isExtension("file", "rtf", "")`, the null-char injection test, etc.).

### DIFF
--- a/src/main/java/org/apache/commons/io/FilenameUtils.java
+++ b/src/main/java/org/apache/commons/io/FilenameUtils.java
@@ -26,7 +26,6 @@ import java.util.Deque;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.stream.Stream;
 
 /**
  * General file name and file path manipulation utilities. The methods in this class operate on strings that represent relative or absolute paths. Nothing in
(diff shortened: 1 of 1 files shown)


garydgregory:
Closing: no one's asking for the extra complications and bloat.

PR #869, opened 2026-08-02
Buffers.clear(...) only zeros the buffer's own window of the backing array

kdelay:
## What

`Buffers.clear(XBuffer)` promises to clear *this* buffer. For array-backed buffers it does more than that: every overload calls

```java
Arrays.fill(buffer.array(), 0);
```

`array()` returns the whole backing array, so `arrayOffset()` and `capacity()` are ignored. A buffer returned by `slice()` shares its parent's array at a non-zero offset, so clearing a slice also zeros bytes that belong to the parent and are outside the slice.

The direct branch of the same method is correct, because it writes zeros through the buffer and therefore respects the window. So the two branches of one method disagree.

## Reproduction

JDK 17, `master` at eb60bd5, 10-element heap array filled with `7`, slice over `[4, 7)`:

```java
byte[] backing = new byte[10];
Arrays.fill(backing, (byte) 7);
ByteBuffer parent = ByteBuffer.wrap(backing);
parent.position(4);
parent.limit(7);
ByteBuffer slice = parent.slice();   // capacity 3, arrayOffset 4

Buffers.clear(slice);
```

| path | result |
| --- | --- |
| array-backed slice | `[0, 0, 0, 0, 0, 0, 0, 0, 0, 0]` |
| same test with `allocateDirect` | `[7, 7, 7, 7, 0, 0, 0, 7, 7, 7]` |

The direct result is what the Javadoc describes.

## Change

`clearBuffer()` has already set the position to zero and the limit to the capacity, so the region that belongs to this buffer is `[arrayOffset(), arrayOffset() + capacity())`. This switches the seven array-backed branches to the ranged `Arrays.fill()` overload. Buffers with `arrayOffset() == 0` are unaffected, which covers everything `allocate()` and `wrap()` produce, and the direct branches are untouched.

`Buffers` is new in 2.23.0 and is not released yet, and the only in-tree callers go through `clearDirect(...)`, so this does not change behavior anyone can be depending on today.

## Tests

Four regression tests in `BuffersTest` (byte, char, int, short, covering each cast variant of the fill). Verified they are load-bearing:

- with the change: `Tests run: 61, Failures: 0, Errors: 0, Skipped: 0`
- with `Buffers.java` reverted: `Tests run: 61, Failures: 4`

Full default goal (`mvn`, so rat, checkstyle, japicmp, spotbugs, pmd and javadoc included): `BUILD SUCCESS`, `Tests run: 6414, Failures: 0, Errors: 0, Skipped: 36`.

---

- [x] I used AI to create any part of, or all of, this pull request. Which AI tool was used to create this pull request, and to what extent did it contribute?
  - Tool: Claude Code (Anthropic), running as an autonomous contribution workflow.
  - Extent: all of it. The defect was found by diffing the `@since 2.23.0` Javadoc contracts against the implementations, and the probe, the patch, the tests and this description were written by the tool.
  - Verification: every number quoted above comes from a run on this machine, not from the model. The probe was executed before and after the patch, and the reverted-source run is how the four tests were confirmed to fail without the change.
  - Human oversight: none before submission. Please review it as unreviewed work. I will answer any question about the change on this thread.
- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
- [x] Each commit in the pull request should have a meaningful subject line and body.

### DIFF
--- a/src/main/java/org/apache/commons/io/Buffers.java
+++ b/src/main/java/org/apache/commons/io/Buffers.java
@@ -86,7 +86,7 @@ public final class Buffers {
             return null;
         }
         if (clearBuffer(buffer).hasArray()) {
-            Arrays.fill(buffer.array(), (byte) 0);
+            Arrays.fill(buffer.array(), buffer.arrayOffset(), buffer.arrayOffset() + buffer.capacity(), (byte) 0);
         } else {
             final byte[] zeros = new byte[IOUtils.DEFAULT_BUFFER_SIZE];
             while (buffer.hasRemaining()) {
@@ -108,7 +108,7 @@ public final class Buffers {
             return null;
         }
         if (clearBuffer(buffer).hasArray()) {
-            Arrays.fill(buffer.array(), (char) 0);
+            Arrays.fill(buffer.array(), buffer.arrayOffset(), buffer.arrayOffset() + buffer.capacity(), (char) 0);
         } else {
             final char[] zeros = new char[IOUtils.DEFAULT_BUFFER_SIZE];
             while (buffer.hasRemaining()) {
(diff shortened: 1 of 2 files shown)


garydgregory:
I converted this PR to draft.

> Human oversight: none before submission. Please review it as unreviewed work. I will answer any question about the change on this thread.

This is not acceptable. Do your own homework before dumping it on someone else.

PR #870, opened 2026-08-05
IO-892: Add maximum traversal depth to FileAlterationObserver

zx8410-ops:
Thanks for your contribution to [Apache Commons](https://commons.apache.org/)!

- [x] Read the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html).
- [x] I used AI to create part of this pull request. OpenAI Codex assisted with analysis, implementation, tests, local verification, and preparation of this description.
- [ ] Run a successful build using the default Maven goal with `mvn`.
- [x] Write unit tests that match behavioral changes.
- [x] Write a detailed pull request description.
- [x] Each commit has a meaningful subject line and body.

## Summary

This PR adds an optional maximum traversal depth to `FileAlterationObserver` through `Builder.setMaxDepth(int)`.

The observed root has depth 0. Setting `maxDepth` to 0 prevents listing entries below the root. At greater depths, entries at `maxDepth` are monitored, but their children are not listed and do not generate events.

The default remains `Integer.MAX_VALUE`, so existing builder usage and deprecated constructors preserve full recursive monitoring. The depth limit applies during both initialization and every `checkAndNotify()` cycle.

## Why this is not a FileFilter

A regular `FileFilter` cannot provide the same traversal behavior. `FileAlterationObserver` calls `directory.listFiles(fileFilter)`, so a depth-aware filter would still enumerate every child in a boundary directory. It also cannot express "monitor this directory but do not descend into it" with one boolean result.

This implementation checks the depth boundary before invoking `listFiles()`.

## Tests

Tests cover default compatibility, invalid values, boundary directory events, ignored events below the boundary, repeated check cycles, verification that `listFiles()` is not called below the boundary, custom root entries, and serialization round trips.

Local Windows/NTFS results for 100,000 files below a depth-1 boundary:

| Mode | Initialize | Steady check average |
| --- | ---: | ---: |
| Unlimited | 9069.894 ms | 8772.802 ms |
| `maxDepth=1` | 126.719 ms | 16.517 ms |

Verification completed successfully:

- `mvn -q "-Dtest=FileAlterationObserverTest,FileEntryTest,FileAlterationMonitorTest" test`
- `mvn -q -DskipTests checkstyle:check javadoc:javadoc package japicmp:cmp`
- `git diff --check origin/master...HEAD`

JIRA: https://issues.apache.org/jira/browse/IO-892

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -62,6 +62,7 @@ The <action> type attribute can be add,update,fix,remove.
       <action type="fix" dev="ggregory" due-to="Gary Gregory">DeferredFileOutputStream better NPE message and new file permission tests (#862).</action>
       <action type="fix" dev="ggregory" due-to="Gary Gregory">Add messages when throwing NullPointerException.</action>
       <!-- ADD -->
+      <action type="add" due-to="bart" issue="IO-892">Add FileAlterationObserver.Builder.setMaxDepth(int) to limit directory traversal depth.</action>
       <action type="add" dev="ggregory" due-to="Gary Gregory">Add IOConsumer.accept(IOConsumer, T) (#846).</action>
       <action type="add" dev="ggregory" due-to="Gary Gregory">Add UnsynchronizedBufferedReader.unwrap() (#850).</action>
       <action type="add" dev="ggregory" due-to="Gary Gregory">Add UnsynchronizedBufferedReader.getPosition() (#851).</action>
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -62,7 +62,7 @@ The <action> type attribute can be add,update,fix,remove.
       <action type="fix" dev="ggregory" due-to="Gary Gregory">DeferredFileOutputStream better NPE message and new file permission tests (#862).</action>
       <action type="fix" dev="ggregory" due-to="Gary Gregory">Add messages when throwing NullPointerException.</action>
       <!-- ADD -->
-      <action type="add" due-to="bart" issue="IO-892">Add FileAlterationObserver.Builder.setMaxDepth(int) to limit directory traversal depth.</action>
+      <action type="add" dev="ggregory" due-to="bart" issue="IO-892">Add FileAlterationObserver.Builder.setMaxDepth(int) to limit directory traversal depth.</action>
       <action type="add" dev="ggregory" due-to="Gary Gregory">Add IOConsumer.accept(IOConsumer, T) (#846).</action>
       <action type="add" dev="ggregory" due-to="Gary Gregory">Add UnsynchronizedBufferedReader.unwrap() (#850).</action>
       <action type="add" dev="ggregory" due-to="Gary Gregory">Add UnsynchronizedBufferedReader.getPosition() (#851).</action>
--- a/src/main/java/org/apache/commons/io/monitor/FileAlterationObserver.java
+++ b/src/main/java/org/apache/commons/io/monitor/FileAlterationObserver.java
@@ -116,6 +116,12 @@ import org.apache.commons.io.filefilter.TrueFileFilter;
  * {@link FileEntry} can be used to capture additional properties that the basic implementation does not support. The {@link FileEntry#refresh(File)} method is
  * used to determine if a file or directory has changed since the last check and stores the current state of the {@link File}'s properties.
  * </p>
+ * <h2>Maximum Depth</h2>
+ * <p>
+ * Use {@link Builder#setMaxDepth(int)} to limit the depth of entries monitored below the root directory. The root directory has depth 0. Setting the maximum
+ * depth to 0 prevents listing any entries below the root. At greater depths, entries at the maximum depth are monitored, but their children are not listed and
+ * no events are generated for those children. The default is {@link Integer#MAX_VALUE}, which preserves full recursive monitoring.
+ * </p>
  * <h2>Deprecating Serialization</h2>
  * <p>
  * <em>Serialization is deprecated and will be removed in 3.0. Consider using a serialization proxy.</em>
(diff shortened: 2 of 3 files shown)


src/main/java/org/apache/commons/io/monitor/FileAlterationObserver.java
```diff
@@ -236,8 +264,14 @@ private static Comparator<File> toComparator(final IOCase ioCase) {
      */
     private final Comparator<File> comparator;
 
+    /**
+     * The maximum depth of entries to monitor.
+     */
+    private final Integer maxDepth;
```
garydgregory: Make this an `int`.

src/main/java/org/apache/commons/io/monitor/FileAlterationObserver.java
```diff
@@ -508,6 +555,17 @@ public Iterable<FileAlterationListener> getListeners() {
...
 
+    /**
+     * Gets the maximum depth of entries to monitor. The root directory has depth 0.
+     *
+     * @return The maximum depth.
+     * @since 2.23.0
+     */
+    public int getMaxDepth() {
+        // A null value preserves unlimited recursion when deserializing streams written before this field was added.
+        return maxDepth != null ? maxDepth : Integer.MAX_VALUE;
```
garydgregory: This value should always be set to an `int` on construction. See my other comment.

garydgregory (changes requested):
Hello @zx8410-ops 
Thank you for your PR. Please comments.

src/changes/changes.xml
```diff
@@ -62,6 +62,7 @@ The <action> type attribute can be add,update,fix,remove.
       <action type="fix" dev="ggregory" due-to="Gary Gregory">DeferredFileOutputStream better NPE message and new file permission tests (#862).</action>
       <action type="fix" dev="ggregory" due-to="Gary Gregory">Add messages when throwing NullPointerException.</action>
       <!-- ADD -->
+      <action type="add" due-to="bart" issue="IO-892">Add FileAlterationObserver.Builder.setMaxDepth(int) to limit directory traversal depth.</action>
```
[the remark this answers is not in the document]
zx8410-ops: Addressed in 5cde511f8: added `dev="ggregory"` to the IO-892 change entry.

src/main/java/org/apache/commons/io/monitor/FileAlterationObserver.java
zx8410-ops: Addressed in 5cde511f8: `maxDepth` is now a primitive `int` and is assigned by the constructor.

src/main/java/org/apache/commons/io/monitor/FileAlterationObserver.java
zx8410-ops: Addressed in 5cde511f8: all construction paths assign the primitive `int` field, and `getMaxDepth()` now returns it directly.

src/main/java/org/apache/commons/io/monitor/FileAlterationObserver.java
```diff
@@ -527,7 +585,19 @@ public void initialize() throws Exception {
...
+        return Stream.of(listFiles(file, entry)).map(f -> createFileEntry(entry, f)).toArray(FileEntry[]::new);
+    }
+
+    /**
+     * Lists the contents of a directory when the parent entry is below the maximum depth.
+     *
+     * @param directory The directory to list.
+     * @param entry     The directory entry.
+     * @return The directory contents or a zero length array if the maximum depth is reached.
+     */
```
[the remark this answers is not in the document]
zx8410-ops: Addressed in 5cde511f8: the Javadoc now states that the directory is listed when its own entry is below the maximum depth.

zx8410-ops:
@garydgregory Thank you for the review. I addressed all four inline comments in 5cde511f8:

- changed `maxDepth` to a primitive `int` assigned on every construction path;
- simplified `getMaxDepth()` to return the field directly;
- clarified the traversal-boundary Javadoc;
- added the `dev` attribute to the IO-892 change entry.

Local verification passed for the focused monitor tests, Checkstyle, Javadoc, package, and japicmp checks. Java CI, CodeQL, and Dependency Review are also all passing.

Could you please take another look?

garydgregory:
Hello @zx8410-ops 
I propose https://github.com/apache/commons-io/pull/871 as an alternative to this PR to reuse Java's [Files.walkFileTree()](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/nio/file/Files.html#walkFileTree(java.nio.file.Path,java.util.Set,int,java.nio.file.FileVisitor)) and its `maxDepth` feature instead of managing depth ourselves.

zx8410-ops:
Thank you. I agree that reusing `Files.walkFileTree()` is the preferable implementation direction and that #871 can supersede this PR.

One detail I think we should verify before closing #870: `FileAlterationObserver` still builds its entry tree recursively (`createFileEntry()` -> `listFileEntries()` -> `listFiles()`), while #871 starts a new walk with the full `maxDepth` for every directory. Because `maxDepth` is relative to each walk's start path, this appears to reset the depth at each recursive call. For example, with `maxDepth=1`, the root walk can return the boundary directory, and creating that entry then starts another `maxDepth=1` walk from the boundary, potentially exposing its children. Returning `SKIP_SUBTREE` for non-start directories also makes each individual walk effectively list one level.

The deep-event test in #871 currently uses `maxDepth=0`, so it may not catch the `maxDepth=1` case. Could we keep a test that sets `maxDepth=1` and verifies that events below the boundary are ignored, and either pass the remaining depth (`maxDepth - relativeDepth`) or perform one root-relative walk?

With that boundary case covered, I am happy for #871 to replace #870 and for this PR to be closed.

garydgregory:
I went with this PR as the other one didn't do the file walking quite right (too deep).

